Issue #65, opened 2015-02-24
is URL rewriting with content negotiation supported?   


nandana:
I wanted to ask whether w3id supports/permits to do URL rewriting with conditions for content negotiation, for instance using [RewriteCond](http://httpd.apache.org/docs/2.0/mod/mod_rewrite.html#rewritecond).

My usecase is to host my ontology in Github pages and use w3id URL as a permanent identifier.  Further, I want to have content negotiation on that URL but unfortunately Github pages doesn't support content negotiation. However, if I can do this at the URL rewrite stage I can still properly do what I wanted to do. 

For example, something like the following.

> RewriteCond %{HTTP:Accept} text/turtle 
> RewriteRule ^foo$ http://proj.github.io/vocab/foo.ttl [R=303,L]
> 
> RewriteCond %{HTTP:Accept} application/rdf+xml
> RewriteRule ^foo$ http://proj.github.io/vocab/foo.xml [R=303,L]
> 
> RewriteRule ^foo$ http://proj.github.io/vocab/foo.html [R=303,L]

P.S. I am not an expert on HTTPD config, so there may be a better way to do this.

prototypo:
Hi Nandana,

Yes, I don’t see any reason this wouldn’t work unless the Apache wasn’t compiled with RewriteCond support. Try it :)

Regards,
Dave
--
http://about.me/david_wood

davidlehn:
Your best bet is to set up a local server to test out configs.  Another project has some content negotiation support that might be useful though I've never tried it.  If you do come up with a good solution let us know and maybe we can start some docs.

https://github.com/perma-id/w3id.org/blob/master/isa/.htaccess

perma-id:
Hi David,

> On Feb 24, 2015, at 13:00, David I. Lehn <notifications@github.com> wrote:
> 
> Your best bet is to set up a local server to test out configs. Another project has some content negotiation support that might be useful though I've never tried it. If you do come up with a good solution let us know and maybe we can start some docs.
> 
> 


That wouldn’t answer the question of whether RewriteCond support is compiled into the w3id.org <http://w3id.org/> Apache.

Regards,
Dave
--
http://about.me/david_wood

davidlehn:
`mod_rewrite` is enabled, it's used by almost every `.htaccess` file on the site.  If there is something missing that's needed, we can enable it.

nandana:
Thanks a lot Dave and David! 
I will do a pull request to try it out.

[...]

PR #66, opened 2015-02-24
Rewrite rule for http://w3id.org/people/nandana

nandana:
Added a set of URL rewrite rules with content negotiation as discussed in https://github.com/perma-id/w3id.org/issues/65. I tested the change in a local Apache server and it worked as expected.

### DIFF
--- a/people/.htaccess
+++ b/people/.htaccess
@@ -22,4 +22,12 @@ RewriteRule ^aidan$ http://aidan.droppages.com/ [R=302,L]
 RewriteRule ^taltmann$ http://52pies.blogspot.com/ [R=302,L]
 RewriteRule ^nxg$ http://nxg.me.uk/norman/ [R=303,L]
 RewriteRule ^daniel$ http://bosatsu.net/foaf/daniel.rdf [R=303,L]
+#A set of rewrite rules with content negotiation for w3id.org/people/nandana
+RewriteCond %{HTTP:Accept} application/rdf\+xml 
+RewriteRule ^nandana$ http://nandana.github.io/foaf/card.rdf [R=303,L]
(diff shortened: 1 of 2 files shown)


davidlehn:
There is a `HTTP_ACCEPT` var available vs the `HTTP:Accept` you used.  I'm not sure if it matters which you use.  Maybe some minor performance thing?  The strict testing for a specific `Accept` header might have issues for more complex `Accept` values.  Some people do more flexible searching for such strings.  I'm guessing a proper solution requires something that better understands content negotiation processing.  See for instance: http://www.snee.com/bobdc.blog/2011/05/quick-and-dirty-linked-data-co.html

Considering you have a handful of rules, does it make sense to make a dir with an .htaccess file with all your rules?  Similar to how some other `people/*` are setup?

nandana:
Yes. I agree using the HTTP_ACCEPT var is better and also made the search a bit more flexible using the reference. Still it is hard to convert all the content negotiation logic into RewriteCond rules. For instance, an accept header like the following will fail under this ruleset.

> Accept: application/rdf+xml;q=0.9,text/turtle
 
I also moved the rules to a separate dir as suggested.

nandana:
@davidlehn Apparently theres is no solution for that corner case. 
* http://stackoverflow.com/questions/28745224/content-negotiation-using-mod-rewrite-rewritecond-rules
* http://webmasters.stackexchange.com/questions/76884/mod-rewrite-content-negotiation-for-mixed-accept-header
* http://serverfault.com/questions/655834/redirect-depending-on-preferred-language-accept-language-with-htaccess

But I think I can live with it for the moment. Will it be possible to test the current configuration?

PR #70, opened 2015-03-23
Temporary URLs for credentials vocabulary, context, openbadges context

ottonomy:
A draft of the [credentials vocabulary](http://specification.openbadges.org/credentials/) is available on the Open Badges Specification site, and developers wish to start using it very soon.

SSL certificate will soon be created, but for now these redirects are to insecure pages.

### DIFF
--- a/openbadges/.htaccess
+++ b/openbadges/.htaccess
@@ -1,4 +1,5 @@
 Options +FollowSymLinks
 RewriteEngine on
 RewriteRule ^v1$ https://dev.truecred.com/contexts/openbadges-v1.jsonld [R=302,L]
+RewriteRule ^v1.1$ http://specification.openbadges.org/1.1/context.json [R=302,L]
 
--- a/openbadges/.htaccess
+++ b/openbadges/.htaccess
@@ -1,5 +1,6 @@
 Options +FollowSymLinks
 RewriteEngine on
+RewriteRule ^$ http://specification.openbadges.org/ [R=302,L]
 RewriteRule ^v1$ https://dev.truecred.com/contexts/openbadges-v1.jsonld [R=302,L]
 RewriteRule ^v1.1$ http://specification.openbadges.org/1.1/context.json [R=302,L]
 
--- a/credentials/.htaccess
+++ b/credentials/.htaccess
@@ -1,4 +1,4 @@
 Options +FollowSymLinks
 RewriteEngine on
-RewriteRule ^$ http://specification.openbadges.org/vocabulary.html [R=302,L]
-RewriteRule ^v1$ http://specification.openbadges.org/v1/context.json [R=302,L]
+RewriteRule ^$ http://specification.openbadges.org/credentials/ [R=302,L]
+RewriteRule ^v1$ http://specification.openbadges.org/credentials/context.json [R=302,L]
--- a/credentials/.htaccess
+++ b/credentials/.htaccess
@@ -1,0 +1,4 @@
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^$ http://specification.openbadges.org/vocabulary.html [R=302,L]
+RewriteRule ^v1$ http://specification.openbadges.org/v1/context.json [R=302,L]


openbadges/.htaccess
```diff
@@ -1,4 +1,5 @@
 Options +FollowSymLinks
 RewriteEngine on
 RewriteRule ^v1$ https://dev.truecred.com/contexts/openbadges-v1.jsonld [R=302,L]
+RewriteRule ^v1.1$ http://specification.openbadges.org/1.1/context.json [R=302,L]
```
msporny: We need to talk about this line on the call tomorrow @davidlehn @dlongley @ottonomy. Doing a v1.1 context, while the path of least resistance, doesn't make us have the important discussion wrt. how we're going to merge the Open Credentials and Open Badges vocabularies. I don't have an opinion on how the merge should happen... just that it /should/ happen and we're hitting the point where it needs to happen.

gkellogg: I'm pretty wary about using version identifiers in namespaces, certainly once a spec is stable, the namespace URI should never change. That's why we're still stuck with `http://xmlns.com/foaf/0.1/` after all these years, it could never be changed.

I would say that once the w3id.org URI is determined, that is what distinguishes it from a previous version such as truecred or some other interim identifier.

ottonomy:
Thanks, @msporny and @gkellogg:

I'm trying to learn as quickly as I can and keep a bunch of different long-term side effects in my head at once while weighing different options. All comments and feedback welcome -- definitely pushing for some quick action here, because we've been delayed in releasing OBI v1.1, but it is important to get a solution that doesn't create big problems for the future.

The namespace for credentials is proposed to be https://w3id.org/credentials/ and for open badges to be https://w3id.org/openbadges/ -- The context files point various classes and properties to IRIs within those namespaces, so if the vocabulary eventually is hosted on w3.org, for example, the namespace for the properties doesn't need to change. We can go through multiple versions of the context within the same namespace.

gkellogg:
The best case if for the vocabulary and the context to be in the same place, that way both the namespace and vocabulary can be loaded from https://w3.id.org/credentials. I'd prefer using a fragment identifier for terms rather than a trailing "/" so that terms dereference to the same resource, but that's a matter of opinion.

For an example of including the vocabulary definition along with the context, see https://github.com/w3c/csvw/blob/gh-pages/ns/csvw.jsonld. This allows the same resource to serve as both context and RDFS vocabulary, and be at the same location as HTML and Turtle variations so that context negotiation will serve the same purpose. Credentials is a smaller vocabulary, so the file should be fairly small. (note another @context embedded under `@graph` which allows us to define terms convenient for the vocabulary definition, without polluting the main namespace, but this is entirely optional at the expense of slightly more verbose JSON-LD).

This allows the context to live at `https://w3id.org/credientials` be something like the following, if requested as JSON-LD:

```json
{
  "@context": {
    "credentials": "https://w3id.org/credentials#",
    ...
  }
}
```

As for versioning, the best practice (IMO) is to not include a version in the URI _at all_, but just use a separate URI. A version number implies that it may change, which would not be the case for a standard; this might mean either overloading an existing context in a forward-comparable way, or using a different URI entirely, if the namespace is effectively already taken.

ottonomy:
/me muses: How advanced does a vocabulary need to be before its community is ready to declare that it has reached the One True listing of terms (excepting the ability to add terms and note previous terms as deprecated) and is ready to be served at the same resource URI as the vocabulary? Does a group at the W3C Community Group level really have the clout to declare that its vocabulary and context is mature enough for that status?

The Badge Alliance really needs to get a release into the world soon, and it does seem like a versioned context file (that applies to open badges only, not to other types of credentials) is the easiest route to doing that without creating big side effects on the future credentials standardization work. We just need a few class names and properties (and their IRIs) finalized in the Credentials Vocabulary so they can be referenced from Open Badges contexts. I included a credentials context separately in this 

@gkellogg: I hear you on the usefulness of having a context and vocabulary available from the same resource URI. I haven't dug into the semantic web world deep enough to know all I need to learn about namespaces and vocabulary development, and I appreciate your time commenting on this to help out.

ottonomy:
Had a very productive conversation about the issues this PR raised. I'm withdrawing it for now and will submit a revised version next week or so.

PR #72, opened 2015-03-27
Initial library ontologies

cKlee:
Tried to enable Content-Negotiation. I hope this will work.


cKlee:
Is there a chance that this will be accepted?

dlongley:
@cKlee, yes! I don't know why it hasn't been yet, I'll look into it later today to determine why it's been held up or if it's simply flown under the radar (most likely the latter).

davidlehn:
An issue that applies to this request and others is how to determine if the namespace being claimed is appropriate.  "library" is a fairly generic term but perhaps that's fine.  It's hard to say without research and discussion so we've just been doing first come, first served.

prototypo:
Since these ontologies are applicable to many libraries, I have no objection.

PR #80, opened 2015-05-05
Bootstrap 3 (fixing layout issues on mobile devices such as iPhone 5)

AlexanderWillner:


msporny:
Thanks for the patch @AlexanderWillner!

davidlehn:
@AlexanderWillner That other "federation sub ontology" commit was supposed to happen too right?

AlexanderWillner:
Well, was supposed to happen in general, so no issue that it has been pushed - but it doesn't really belong to the Bootstrap update of course.

davidlehn:
It got lumped in the pull, but the commit isn't really connected.  Just making sure it didn't slip in by mistake.

PR #83, opened 2015-05-11
Add new permalink

rivets:


msporny:
Hey @rivets - in the future, try to use a better commit comment description. Most every addition to this site is "adding new permalink". For example, something like "Add Greycite website tool redirect." would've been better.

PR #86, opened 2015-05-11
Capture string after base and copy to path

dr-shorthair:
Also, change to 307


dr-shorthair:
Thanks David - I think we've bottomed this out for now. Getting the results I was expecting :-)

PR #88, opened 2015-05-20
add people/jennybc

jennybc:


jennybc:
These URLs won't resolve when visited, e.g., as in a browser. They are API endpoints. Is that going to be a problem? I could commit browser-ready URLs here and then convert to what I need on my end.

jennybc:
I went ahead and made the switch discussed above, just to cut to the chase.

davidlehn:
You are welcome to use whatever URLs are appropriate for your use case.

PR #89, opened 2015-05-21
correct wording in README

jennybc:


davidlehn:
Do you need the `<` `>` around the URLs in the README?  The link checker is doing a very basic regex to pull out URLs to test and includes the trailing `>` which causes the link check to fail.

jennybc:
No they are not necessary in GitHub-flavored markdown, so I've removed them. They are compatible with GitHub-flavored markdown and are necessary for Pandoc-flavored markdown, so you might see them crop up in other people's READMEs FWIW.

Thanks.

davidlehn:
Ok.  If it becomes a bigger problem we'll fix the link scanner.  Easier to just avoid the issue for now.

PR #100, opened 2015-07-01
w3ids for isil and zdb

cKlee:


msporny:
Could you please squash these commits into a single commit or two? We don't want to spam the commit log. After you do that, we'll merge your changes.

cKlee:
squashed into one commit

msporny:
Merged, thank you. :)

PR #101, opened 2015-07-03
US-* needs to be lowercase in URI

cKlee:
Sorry this needs to be corrected.

### DIFF
--- a/isil/.htaccess
+++ b/isil/.htaccess
@@ -1,16 +1,44 @@
-# Turn off MultiViews
-Options -MultiViews
-
-# Directive to ensure *.rdf *.owl *.jsonld *.ttl files served as appropriate content type,
-# if not present in main apache config
-AddType application/rdf+xml .rdf .owl
-AddType text/turtle .ttl
-AddType application/ld+json .jsonld
-
-# Rewrite engine setup
 RewriteEngine On
 
-# Default Rule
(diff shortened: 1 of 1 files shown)


davidlehn:
Some pages talking about using this technique have some warnings about performance.  If that turns out to be true we might need to find another way to do this transformation.

cKlee:
I don't know about the Apache configuration of w3id. But enabling int:tolower in httpd.conf should make a difference

    # Add RewriteMap for redirecting to lowercase URIs
    <IfModule mod_rewrite.c>
    RewriteMap lc int:tolower
    </IfModule>

Then in the .htaccess this is possible

    RewriteCond %{REQUEST_URI} [A-Z]
    RewriteRule ^(.*)$ ${lc:$1} [R=301,L]

davidlehn:
We'll leave this as is for now and optimize if needed.

Upon further consideration, I'm unsure the patch should have been accepted in the first place.  We originally only thought the service would just be doing simple rewrites and redirects.  This is venturing further into parameter transformations.  It seems like it will be fragile.  What if the US-* service starts publishing URLs in UPPERCASE?  Then the always-lowercase transformation will be broken.  Why are you not just using the proper target lowercase US-foo string in the first place?

cKlee:
> Why are you not just using the proper target lowercase US-foo string in the first place?

This is because the ISO 15511 allows ISIL with uppercase letters and Library of Congress coined lowercase URIs.

Besides W3ID is great service. I hope you'll further allow such URI manipulations.

PR #103, opened 2015-07-08
fix bash-syntax in travis.yml error output

stain:
`> &2` didn't work, but `>&2` does.. Bash is fun.


msporny:
I'm waiting for the build to settle to pull in your changes.

PR #120, opened 2015-09-17
Update .htaccess

diegoref:
udpate htaccess

### DIFF
--- a/mare/.htaccess
+++ b/mare/.htaccess
@@ -3,3 +3,4 @@ Options +FollowSymLinks
 RewriteEngine on
 RewriteRule ^$ http://mare.istc.cnr.it:7001/mare/demo/ [R=302,L]
 RewriteRule ^(.+)$ http://mare.istc.cnr.it:7001/mare/demo/$1 [R=302,L]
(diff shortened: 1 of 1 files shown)


diegoref:
I would like to add a line at the .htaccess file. What should I do????

dlongley:
> I would like to add a line at the .htaccess file. What should I do????

You submit a PR, as you have, and wait for it to be merged by a human. However, you've submitted five different PRs, leaving multiple ones open that affect the same line, which makes it unclear to a human as to your intent.

Note that we can't merge more than one PR that affects the same line, and if you don't tell us which one you want us to use, we'll just wait for you to close the ones you don't want so we don't mess you up. Once you've got just one PR and it's ready to go, we'll merge.

The usual flow is to get everything sorted out and then submit your PR. If you happen to make a mistake and catch it before we merge, leave a comment indicating so and then fix your existing PR and force push to change your history. Don't open multiple PRs that affect the same thing, because we won't know what you want us to do.

Would you like me to merge this PR now? I see the others have been closed.

diegoref:
I figured the five pull requests and I was able to remove. I submitted now once that should be the right one. Yes,  could you please merge it? thx

dlongley:
> I figured the five pull requests and I was able to remove.

Sure, it's not a problem. The issue was that you left several different conflicting ones open so we didn't know which one you wanted us to merge.

PR #122, opened 2015-09-17
Update .htaccess

diegoref:
Please one more time! :)


davidlehn:
You probably want to also change the http://w3id.org to https://w3id.org.

PR #136, opened 2015-09-29
Revert "added NE and P flags to rewrite"

zednis:
Reverts perma-id/w3id.org#135

pull request did not have the desired effects.  Please revert.  Thank you.

### DIFF
--- a/national-ocean-council/api/.htaccess
+++ b/national-ocean-council/api/.htaccess
@@ -1,3 +1,3 @@
 Options +FollowSymLinks
 RewriteEngine on
-RewriteRule ^(.+)$ https://nocv.tw.rpi.edu/elda/api/$1 [L,NE,P]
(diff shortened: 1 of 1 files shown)


zednis:
Thanks.

davidlehn:
What effect are you trying to do?

zednis:
I would like to change the redirect to a proxy so the w3id.org URL is preserved when the user loads the webpage in a browser and the nocv.tw.rpi.edu source is not necessarily visible.

https://w3id.org/national-ocean-council/api/ -> https://nocv.tw.rpi.edu/elda/api/

davidlehn:
w3id.org doesn't work as a full proxy at this time, sorry.  That would require more infrastructure and resources than we have in place now.  It is annoying that the URL will change in browsers so if that's important you may need to look into other services.  For use cases where the URL isn't exposed in the UI the redirect works well.

PR #139, opened 2015-10-17
Adding /unit from Lepetit Florian

stain:
From Lepetit Florian's [email to perma-id](https://lists.w3.org/Archives/Public/public-perma-id/2015Oct/0000.html) on 2015-10-16

> We are a group of students from INSA Rouen (a French engineering school)
> working on linked data applications based on UNIT's ( http://www.unit.eu/ )
> meta-data of learning resources.

### DIFF
--- a/unit/README.md
+++ b/unit/README.md
@@ -5,11 +5,11 @@ Homepage:
 * http://www.unit.eu/
 
 Vocabularies:
-* https://www.w3id.org/unit/lro/schema
+* https://w3id.org/unit/lro/schema
 
 SPARQL endpoints:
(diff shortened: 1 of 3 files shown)


stain:
As to claiming `/unit` - I guess it's first come, first served - and http://unit.eu/ is already owned by the same institution.

I didn't put a README example for https:/www.w3id.org/unit/lro/schema/* as I don't know any terms that would work instead of `*` - the ontology at http://unit-1.crihan.fr/lro/schema only defines `#`-based resources like http://unit-1.crihan.fr/lro/schema#hasDocType

PR #163, opened 2015-11-16
Add 302 redirect to ESC turtle on GitHub.

acka47:
See https://github.com/hbz/vocabs-edu/issues/5.


class/.htaccess
```diff
@@ -0,0 +1,3 @@
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^class/esc/.* https://raw.githubusercontent.com/hbz/vocabs-edu/master/esc.ttl [R=302,L]
```
davidlehn: This probably isn't what you want.  It will redirect URLs like https://w3id.org/class/class/esc/.

acka47:
@davidlehn Thanks for the hint. I am still learning htaccess. It should be ok now.

PR #175, opened 2015-11-24
permanent urls for games

dgarijo:
Added 3 permanent urls for three game applications.

### DIFF
--- a/games/.htaccess
+++ b/games/.htaccess
@@ -1,0 +1,10 @@
+Options +FollowSymLinks
+RewriteEngine on
+#Rewrite rules for Asteroids (no content negotiation)
+RewriteRule ^asteroids$ http://52.31.80.76/HTML5-Asteroids/ [R=302,L]
+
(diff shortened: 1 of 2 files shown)


dgarijo:
I don't understand why the build failed. All my links seem to be ok (they are simple 302 redirections), and the error it shows seems to be on the "rdw" repository . Can someone help me here?

davidlehn:
The failure is due to the link checker checking links for every project.  An unrelated redriect is returning 503 which will cause every commit check to fail now.  Don't worry about it.

Why are you using IPs in the URLs vs host names?  That seems like a fragile setup compared to using the DNS system.

dgarijo:
Thanks for your response!
Unfortunately I did not set those IPs and I have to live with them, but I will contact the responsible asap. As soon as I get new URLs, I will change them.

davidlehn:
Ok.  If possible please keep resources on something that is somewhat stable to avoid the need for frequent w3id updates.  Thanks.

PR #179, opened 2015-12-07
Redirection to an extension of QUDT Quantity Vocabulary 1.1

KMax:


KMax:
The build failed because of other redirections which have got 404 error, e.g.:
```
URL        `http://localhost/people/jennybc/googlesheets_gap_url'
Real URL   http://localhost/people/jennybc/googlesheets_gap_url
Check time 0.009 seconds
Size       307B
Result     Error: 404 Not Found
```

Did it actually check my .htaccess file? I.e. my changes.

KMax:
I just checked some of the failed links from the build log and they actually work...

PR #191, opened 2015-12-11
.htaccess added

pipauwel:


davidlehn:
While pondering the wisdom of letting common words like "list" be claimed, I noticed in the RDF you link to that you use http://w3id.org/list vs https://w3id.org/list.  I would suggest you use the https version since it's more secure and all http requests are slower since they will have an extra 301 redirect to https anyway.

pipauwel:
Thank you for pointing this out! Let me properly switch to https...

PR #193, opened 2015-12-14
name of the ontology has changed

bozicb:
therefore new link was needed.


davidlehn:
The new target URL is 404.

bozicb:
fixed now.

PR #200, opened 2015-12-30
New space for OCC

essepuntato:
This space should be used for guaranteeing the persistence of the entities of the Open Citations Corpus -- see:
- http://www.nature.com/news/publishing-open-citations-1.13937
- http://www.emeraldinsight.com/doi/abs/10.1108/JD-12-2013-0166

### DIFF
--- a/oc/.htaccess
+++ b/oc/.htaccess
@@ -1,0 +1,4 @@
+Header set Access-Control-Allow-Origin *
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^/?$ http://www.opencitations.net [R=302,L]


davidlehn:
`www.opencitations.net` doesn't exist and `opencitations.net` doesn't respond.  Are they supposed to be working?

essepuntato:
Hi David,

> www.opencitations.net doesn't exist and opencitations.net doesn't respond. Are they supposed to be working?

We are working in these days in order to run the server up again, and then we are planning to increase the dataset with new data. 

It’s true that the website is not active now, but we would like to reserve the w3id.org <http://w3id.org/> space in advance, in order to have everything ready when the server will be up again. I hope this could be possible.

Have a nice day :-)

S.

PR #209, opened 2016-01-03
Redirects for Kane Mar.

kane:


davidlehn:
Can you pull latest upstream and update this patch? The file was alphabetized so please update your redirect in the proper place. Thanks.

PR #210, opened 2016-01-04
Revert "pm added for paul mckellar"

msporny:
Reverts perma-id/w3id.org#207.

We ask people to put their personal website stuff under /people/

### DIFF
--- a/pm/.htaccess
+++ b/pm/.htaccess
@@ -1,4 +1,0 @@
-Options +FollowSymLinks
-RewriteEngine on
-RewriteRule ^$ http://paulmckellar.com [R=302,L]
-


bsletten:
Are these "rules" encoded somewhere?

msporny:
They are now :p

https://github.com/perma-id/w3id.org#guidelines-on-identifiers

bsletten:
So, we are enforcing structure to URL’s now based on what they are? That doesn’t seem like an overly scalable strategy. Do the tests currently catch the case where we have a .htaccess rule for bsletten, but then someone tries to create a /people/bsletten domain? If not, I’ll try to get something in there to help automate this. Things are going to become trickier to deal with, particularly with you making up new rules all the time. ;)

Are there domains other than people that we are going to want to dictate structure to? We should get policies in place beforehand if so.

msporny:
> So, we are enforcing structure to URL’s now based on what they are? 

Only for people.

Everything else is not a person (project, organization, etc.)

> That doesn’t seem like an overly scalable strategy. 

I'll point out that this whole website isn't a scalable strategy. It just seemed like a good idea at the time (5+ years ago). :P 

> Do the tests currently catch the case where we have a .htaccess rule for bsletten, but then someone tries to create a /people/bsletten domain? If not, I’ll try to get something in there to help automate this. 

Nope, and thanks. :)

> Things are going to become trickier to deal with, particularly with you making up new rules all the time. ;)

We have a very clear policy on not harassing the site founders. :)

> Are there domains other than people that we are going to want to dictate structure to? We should get policies in place beforehand if so.

Not that I can think of, no. I started to say something about trademarks, but I think in that case - the trademark owner wins via cease-and-desist (and I hope this doesn't become a contentious issue).

davidlehn:
What is the status of this?
@paulmars Would you mind if /pm was moved to /people/pm?  Or should we just leave it as is and close this?
@msporny Can we delete the related revert branch for pr 207?

paulmars:
It sort of bums me out honestly. As a internet geek I like having "pm" on various places, and the concept of a perma-id changing is surprising. Also, I've never been kicked off my name before. Instagram + twitter have sort of defined 'pm' for me.

Could there be an alternative.

davidlehn:
Closing.  Reopen if needed.

PR #211, opened 2016-01-04
Revert "Redirects for Pierre Reimertz"

msporny:
Reverts perma-id/w3id.org#203.

This wasn't the correct way to do a single person redirect. They want to add an entry to people/.htaccess instead.

### DIFF
--- a/people/r/.htaccess
+++ b/people/r/.htaccess
@@ -1,3 +1,0 @@
-Options +FollowSymLinks
-RewriteEngine on
-RewriteRule ^$ http://reimertz.co/ [R=302,L]


msporny:
Hmm, on second thought, I have no idea what he's trying to do. A single person redirect or a single person directory w/ multiple redirects to different projects?

reimertz:
This is a single person redirect. Have I missunderstood something?

msporny:
Why didn't you do something like this: https://github.com/perma-id/w3id.org/blob/master/people/.htaccess ?

Also, I suggest you use more than a single letter for your redirect. Using 'r' makes it so that we can't have something like /people/r/reimertz2 in the future.

So, this sort of link might be better:

https://w3id.org/people/reimertz

and if you're just redirecting to a single URL, then change this file:

/people/.htaccess

msporny:
To be clear, I'm not trying to be difficult.

Creating a directory with an .htaccess file in it with only one entry is a bit overkill. Adding a single line to the /people/.htaccess file is simpler and easier on the server.

reimertz:
Of course! No worries at all. You are awesome. :)

I'm just a noob doing to little .htaccess ha kin'
I'll fix this tomorrow.

PR #213, opened 2016-01-04
Add people/nick

nickandrew:
Add the standard `.htaccess` file to enable people/nick.


msporny:
@nickandrew - see this thread before we pull - https://github.com/perma-id/w3id.org/pull/211

nickandrew:
Ok, redoing it per #211.

msporny:
Thanks, also as a side question - how did you hear about w3id.org? We've had a flood of requests tonight and we're trying to figure out if it was because of some story or blog post.

nickandrew:
It was mentioned on the twitters.

PR #230, opened 2016-01-08
redirect for the DIO ontology added

nimonika:
This pull request is to add a redirect to the DIO ontology


nimonika:
Please can this pull request be merged.Thanks

PR #240, opened 2016-01-13
Redirect for the LODSight's RDF Path vocabulary

jindrichmynarz:


jindrichmynarz:
It seems that the build failed because several redirects are invalid. It appears that the redirect proposed in this commit fails because `Header add` is not supported, since the `sudo a2enmod headers` is not run in the build setup. This is strange because the CORS headers configuration is copied verbatim from the [reference `.htaccess`](https://github.com/perma-id/w3id.org/blob/master/security/.htaccess).

davidlehn:
Perhaps the apache tests never failed before since the "security/" setup doesn't have tests.  I added the headers module for testing so that part should work now.  I our simple test finder may be trying to test the URLs in your README which caused it to find that failure.

jindrichmynarz:
OK, I didn't know the test runner extracts the URLs from README. The README links the W3ID URL set up in this pull request, so this is why the test failed.

PR #251, opened 2016-01-29
Update .htaccess

jhaag75:
added rule for Chinese HTML pages redirects

### DIFF
--- a/xapi/acrossx/.htaccess
+++ b/xapi/acrossx/.htaccess
@@ -9,6 +9,8 @@ RewriteCond %{HTTP_ACCEPT} text/html [OR]
 RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
 RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
 RewriteRule ^([a-z-]+)/?$ http://xapi.vocab.pub/datasets/acrossx/$1 [R=303]
+RewriteRule ^([a-z-]+)/(zh-cn)$ http://xapi.vocab.pub/datasets/acrossx/$1/zh-cn [R=303]
+RewriteRule ^([a-z-]+)/(zh-tw)$ http://xapi.vocab.pub/datasets/acrossx/$1/zh-tw [R=303]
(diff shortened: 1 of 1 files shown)


xapi/acrossx/.htaccess
```diff
@@ -9,6 +9,8 @@ RewriteCond %{HTTP_ACCEPT} text/html [OR]
 RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
 RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
 RewriteRule ^([a-z-]+)/?$ http://xapi.vocab.pub/datasets/acrossx/$1 [R=303]
+RewriteRule ^([a-z-]+)/(zh-cn)$ http://xapi.vocab.pub/datasets/acrossx/$1/zh-cn [R=303]
+RewriteRule ^([a-z-]+)/(zh-tw)$ http://xapi.vocab.pub/datasets/acrossx/$1/zh-tw [R=303]
```
davidlehn: You could use one generic pattern for these and future language urls:
`RewriteRule ^([a-z-]+)/([a-z-]+)$ http://xapi.vocab.pub/datasets/acrossx/$1/$2 [R=303]`

jhaag75:
Thanks for the note. I was thinking of doing that, but noticed there is already a rule for sub paths that conflicts with this.  I tried testing it but it always seems to process the rule below instead with the no escape. Usually, I want to honor that one but in the case of the language IRI I don't. Is there a better way to do this so that both are supported gracefully? Thanks David.

RewriteRule ^([a-z-]+)/([a-z-]+)$ http://xapi.vocab.pub/datasets/acrossx/$1/#$2 [R=303,NE]

davidlehn:
I'm not an expert in the rewrite mod. Check the docs and run tests. I would try to choose a url layout that avoids the conflicts in the first place.

jhaag75:
Ok, thank you. I will look into adding better rules and layout. These are the only examples we have to work with for now. Can you merge this pull request for now? I would like to use these IRIs as a graph for an RDFa import running.

PR #255, opened 2016-02-06
Update .htaccess

bozicb:


davidlehn:
You could use the rewrite module type negotiation to offer different formats.

PR #257, opened 2016-02-06
Create .htaccess

marianofl1971:
I am building an ontology on education.


davidlehn:
You should use https://w3id.org/... vs http in http://informaticaemprendimiento.org/ontologies/education/universities/academic-programs/degree-programs.owl
You also need to update the rules to redirect sub-resources since as written it will just work for the root path.

marianofl1971:
I have just modified the file. Could you tell me if it is right, please?

Thank you very much for your help.

Mariano

2016-02-06 23:50 GMT+01:00 David I. Lehn <notifications@github.com>:

davidlehn:
By https vs http I mean the w3id URLs in the .owl files on your site.  When dereferenced, they will always redirect to https.  It's recommended to just use https.  Also in the .htaccess file you don't need the https://w3id.org part, you can just use something like ^(.*)$.

marianofl1971:
I have just carried out the changes.

Thank you very much for your patience.

Regards,

Mariano

2016-02-14 3:47 GMT+01:00 David I. Lehn <notifications@github.com>:

marianofl1971:
Great!

Thank you very much!

Mariano

2016-02-18 17:39 GMT+01:00 David I. Lehn <notifications@github.com>:

PR #268, opened 2016-02-16
Update .htaccess

luigi-asprino:


cis/.htaccess
```diff
@@ -1,5 +1,6 @@
 Header set Access-Control-Allow-Origin *
 Options +FollowSymLinks
 RewriteEngine on
-RewriteRule ^$ http://mare.istc.cnr.it:7001/mare/demo/ [R=302,L]
-RewriteRule ^(.+)$ http://mare.istc.cnr.it:7001/mare/demo/$1 [R=302,L]
+RewriteRule ^resource/(.+)$ http://lodview.it/lodview/?sparql=http://wit.istc.cnr.it/sparql&prefix=http://w3id.org/cis/resource/&IRI=http://w3id.org/cis/resource/$1 [R=302,L]
```
davidlehn: I notice you are using http://w3id.org here vs https://w3id.org.  Any reason for that?  w3id.org will always redirect http requests to https.

luigi-asprino:
I made a mistake!

PR #293, opened 2016-03-07
redirect for linked data vocabularies

alanrobson:
Please consider merging this into w3id.org and giving my linked data vocabularies a permanent place to call home on the web. 

There is not much to see now, but I will be to adding vocabularies for network management. The destination domain is hosted by godaddy and should respond to a get request at <http://asdfghjkl1234567.com/linked/index.html> (I know, not a catchy domain but it was available and I can remember it)

An actual test vocabulary (not my vocabulary, but a placeholder) can be downloaded from <http://asdfghjkl1234567.com/linked/vocab/aggregate/0/aggregate.ttl> and if I get my wish also from http(s)://w3id/vocab/aggregate/0/aggregate.ttl soon as well though I'll probably refactor the URLS later to make them less clunky.

Many thanks

Alan

### DIFF
--- a/vsat/.htaccess
+++ b/vsat/.htaccess
@@ -1,3 +1,3 @@
 Options +FollowSymLinks
 RewriteEngine on
-RewriteRule (.*) http://asdfghjkl1234567/linked/$1 [R=302,L]
+RewriteRule (.*) http://asdfghjkl1234567.com/linked/$1 [R=302,L]
--- a/vsat/.htaccess
+++ b/vsat/.htaccess
@@ -1,0 +1,3 @@
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule (.*) http://asdfghjkl1234567/linked/$1 [R=302,L]
--- a/vsat/README.md
+++ b/vsat/README.md
@@ -2,3 +2,14 @@ Linked data url reditrection
 ============================
 
 A permanent home for anchoring linked data URIs for vocabularies pertaining to VSAT networks
+
+contact 
+-------
+
+alan.robson@viasat.com
+
+Intent
+------
+
+To pubish a collection of linked data vocabularies for network management purposes with a specific focus
+on specifying threshold events for timeseries data.
--- a/vsat/README.md
+++ b/vsat/README.md
@@ -1,7 +1,7 @@
 Linked data url reditrection
 ============================
 
-A permanent home for anchoring linked data URIs for vocabularies pertaining to VSAT networks
+A permanent home for anchoring linked data URIs for vocabularies pertaining to VSAT networks at https://w3id.org/vsat/...
 
 contact 
 -------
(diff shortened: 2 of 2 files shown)


vsat/.htaccess
```diff
@@ -0,0 +1,3 @@
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule (.*) http://asdfghjkl1234567/linked/$1 [R=302,L]
```
davidlehn: Missing .com.

davidlehn:
> http(s)://w3id/vocab/aggregate/0/aggregate.ttl

w3id.org http always redirs to https. Location would be:
https://w3id.org/vsat/vocab/aggregate/0/aggregate.ttl

You are right that your domain name is not catchy.  Why not link to a real project site?  Is this something ViaSat, Inc is working on?  What is the project here?  We'd prefer to add links to active projects that require very long term identifiers rather than to just placeholders.  If it's a work in progress that needs the links working now, that's fine, but in that case please at least add some details, links, and/or contact info to the README.  Thanks.

alanrobson:
Hi

I am currently employed by ViaSat Inc. but I don’t speak for them. I am trying to establish the case for using linked data for network management but I don’t have a project in the sense that, for example “Callimachus” is a project. I have no suitable infrastructure with which to demonstrate and establish this idea hence the temporary hosting arrangement to start with. I’m sure everyone would prefer that a corporation of any reasonable size host their own PURL style solution but that’s really not practicable at this stage and in any case I hope that the vocabularies that I build will be of general interest to anyone in the field.

I had hoped to establish a long term identifier because it becomes part of each vocabulary’s identity from the start, even before I have a chance to establish the idea and find a permanent hosting location for it. One of the main attractions of using linked data is the stability of the identifiers, so I am reluctant to publish a system and then have to rework it each time identifiers change.

So I would categorize it as a “work in progress” and I will add contact and other details to the README

Thanks for your consideration

Alan

davidlehn:
Ok, that's fine.  How best to handle long term identifiers is up for debate.  purl.org and w3id.org are one way to do it, there are others.  Just fix the rule address and add contact info and we'll merge it.  Note that if you're just redirecting to simple files you could even host them on github.

alanrobson:
I'm hoping the address info is fine now (I used the hydra entry as a template and that seems to work). I will tell my users to browse to https://w3id/vsat/...

I also added contact info.

Many thanks

Alan

davidlehn:
This won't work:
RewriteRule (.*) http://asdfghjkl1234567/linked/$1 [R=302,L]

Change to this before we can merge this PR:
RewriteRule (.*) http://asdfghjkl1234567.com/linked/$1 [R=302,L]

Telling your users this won't work:
https://w3id/vsat/...

Use this:
https://w3id.org/vsat/...

alanrobson:
That was stupid of me, I have got it now,

My apologies and thanks

Alan

davidlehn:
I removed the trailing "..." after the w3id.org url in the readme since it breaks the link checker.

alanrobson:
Thank you

PR #297, opened 2016-03-12
Add /id-rules/

stain:
permanent identifiers for the individual rules of the 10 Simple Rules for Identifiers paper.

### DIFF
--- a/id-rules/.htaccess
+++ b/id-rules/.htaccess
@@ -1,0 +1,29 @@
+RewriteEngine on
+
+# main page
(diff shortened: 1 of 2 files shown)


stain:
BTW - The Travis check fails because the unrelated https://w3id.org/BCI-ontology which sometimes gives a 500 Server Error at its final destination http://bci.pet.cs.nctu.edu.tw/ontology - however testing this with `curl -L` gives correctly an OWL file, and testing this in a browser also gives a HTML rendering.

PR #322, opened 2016-04-05
Set up for RAND project use.

johnpane:
Establish top-level directory for RAND Corporation, and
project-specific directory for the RAND OBD project. RAND is a
non-profit, non-partisan research firm (www.rand.org).

### DIFF
--- a/RAND/.htaccess
+++ b/RAND/.htaccess
@@ -1,0 +1,4 @@
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^RAND$ RAND/ [NC]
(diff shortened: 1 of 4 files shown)


davidlehn:
Oops, this patch is kind of broken, I merged without thinking.
You want to take the current dir prefixs out since they are implied by the .htacces being in a dir.
Also can simplify the rules to make the trailing / optional.
Ie, both dirs could just have a rule like:
RewriteRule ^/?$ https://www.rand.org/ [R=301,L]

johnpane:
Thanks for your help! I was trying to figure out what I did wrong! Will fix now.

PR #324, opened 2016-04-06
Master

dgarijo:
Added Materials for paper web page


dgarijo:
I don't understand why this has conflicts with the main branch. I have just added some files and a simple redirection. I would be grateful if some admin would tell me what am I doing wrong.

davidlehn:
I'm guessing you just wanted to add this commit:
https://github.com/perma-id/w3id.org/pull/324/commits/f38a43c6486315dfdd820eb62c202f93252cd727

But the merge seems to have changed some other things:
https://github.com/perma-id/w3id.org/pull/324/files

No idea why or what happened.  You should probably try to update your branch so it has latest master changes then rebase you commit on top of master and force push the result to github.  Not sure why it got so tricky.  End result should be that the PR just has your commit changes.

dgarijo:
I will do as you suggest and create another pull request

PR #355, opened 2016-04-25
Add lobid folder and rpb2 redirect.

acka47:


lobid/.htaccess
```diff
@@ -0,0 +1,4 @@
+
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^lobid/rpb2 https://raw.githubusercontent.com/lobid/vocabs/master/rpb/rpb2.ttl [R=302,L]
```
davidlehn: I think you mean to have ^rpb2$ here since this is already in the lobid dir.  Otherwise it will match /lobid/lobid/rpb2.

acka47:
@davidlehn Thanks! I hope now it's correct.

PR #360, opened 2016-04-27
Adding the Qanary Vocabulary

D063520:
Hi from France!

We have developed a new vocabulary called Qanary in the Framework of the WDAqua project: 

http://wdaqua.informatik.uni-bonn.de

Could you create for us a permanent uri?

Merci
Dennis


D063520:
Thank you!

PR #362, opened 2016-04-28
added the Audio Feature Ontology redirects

darkjazz:
Redirects to be created:
http://w3id.org/afo/onto/1.1# -> http://sovarr.c4dm.eecs.qmul.ac.uk/af/ontology/1.1#
http://w3id.org/afo/vocab/1.1# -> http://sovarr.c4dm.eecs.qmul.ac.uk/af/vocabulary/1.1#

### DIFF
--- a/afo/.htaccess
+++ b/afo/.htaccess
@@ -1,4 +1,4 @@
 RewriteEngine on
 
-RewriteRule ^onto/1.1(.*)$ http://sovarr.c4dm.eecs.qmul.ac.uk/af/ontology/1.1$1 [R=302,L]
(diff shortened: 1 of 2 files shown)


davidlehn:
Did you want the `.../1.1(.*)` and `.../1.1$1` wildcard parts?  The `#xxx` part is client side only so won't get passed through that way.  It would help with `.../1.1/foo` though.  If you expect to use the same target server for other purposes or new versions it may make sense to take off 1.1 and just use a wildcard for everything.

Also your comment mentions "http://w3id.org/...".  Use https since http usage is always redirected to https first.

darkjazz:
sorry, yes you are right, i don't want to include the version numbers, just the wildcards. i have committed the updates.

PR #373, opened 2016-04-30
URI for LOD Quality Monitor Service

jerdeb:


jerdeb:
Hi,

Can you please tell me why the build fail? Do I need to do something here?

Cheers,
Jeremy

PR #383, opened 2016-05-07
Update of .htaccess

anuzzolese:
This update of .htaccess enables the redirect to all subfolders of
scholarlydata.

### DIFF
--- a/scholarlydata/.htaccess
+++ b/scholarlydata/.htaccess
@@ -2,3 +2,4 @@ Header set Access-Control-Allow-Origin *
 Options +FollowSymLinks
 RewriteEngine on
 RewriteRule ^/?$ http://www.scholarlydata.org [R=302,L]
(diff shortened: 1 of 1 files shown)


davidlehn:
Use this instead:
RewriteRule ^(.+)$ http://www.scholarlydata.org/$1 [R=302,L]

anuzzolese:
@davidlehn you mean that rule as the only rewrite rule of the htaccess?

davidlehn:
No, I meant just the additional rule.  You could combine both into just:
RewriteRule ^(.*)$ http://www.scholarlydata.org/$1 [R=302,L]

AnLiGentile:
I changed it to the single rule
RewriteRule ^(.*)$ http://www.scholarlydata.org/$1 [R=302,L]

davidlehn:
Done in #386

PR #388, opened 2016-05-16
Handling all URLs

essepuntato:
Added a rule for redirecting everything (even composite URL) to the main http://www.opencitations.net website.


davidlehn:
You could simplify both rules into one:
RewriteRule ^(.*)$ http://www.opencitations.net/$1 [R=302,L]

essepuntato:
@davidlehn 

I've created a new pull request with the modification you suggested, but still I received an error relating to some automatic check. Any idea about how I can avoid it and, finally, have my modification available on w3id.org? 

Thanks in advance for your time. 
Have a nice day :-)

S.

davidlehn:
Fixed in #390.

PR #396, opened 2016-05-18
using LODE visualisation for html, fixed owl/rdf

muddymudskipper:


dgarijo:
@muddymudskipper Are you aware that if you use this type of redirection in your ontology, if you try to derreference a term like "effect" will lead to a 404?
I have tested the target urls and they work. I'll just wait for you to acknowledge this limitation.

muddymudskipper:
thanks for pointing that out, so don't merge at this point. I will fix this in the near future.

dgarijo:
Ok, you can have a look at some other published vocabularies like https://github.com/perma-id/w3id.org/blob/master/linkedEarth/.htaccess if you want a reference :)

PR #399, opened 2016-05-19
Create w3id.org for personal homepage

holgerst:


dgarijo:
@holgerst why do you add the (.*) plus $1 in the redirection? Are you planning on redirecting additional contents on your personal page? If you are just using a 302 redirect to your personal page, then you can change the redirection to something like: RewriteRule ^$ http://www.uni-saarland.de/~holger/ [R=302,L]

holgerst:
@dgarijo: Yes, you are right! I want to directly link via w3id.org to additional content directly and not just to the home (directory).

PR #404, opened 2016-05-27
Create ctdl

CTITechPlanning:
We would like to create a project for the credential transparency initiative. 
We are new to this process, so please let us know if there are issues.
Thanks
Michael Parsons

### DIFF
--- a/ctdl
+++ b/ctdl
@@ -1,0 +1,6 @@
+Options +FollowSymLinks
+RewriteEngine on
+# main site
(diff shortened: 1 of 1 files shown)


dgarijo:
@CTITechPlanning I merged the pull request because the urls are fine, but I have reverted it because there are no .htaccess files. You should 
a) Create a folder called ctdl.
b) add a .htaccess file in that folder with the contents of your current ctdl file.
I will try to create the proper folder redirection.

PR #411, opened 2016-05-30
 Publish an ontology - Permanent identifier

fortiss-itsm:
Permanent identifier for ontology publication. 
Would you please create a persistent URL for this ontology?

Best Regards,
Matthias

### DIFF
--- a/bmontology/.htaccess
+++ b/bmontology/.htaccess
@@ -1,0 +1,7 @@
+
+Header set Access-Control-Allow-Origin *
+Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^$ http://download.fortiss.org/public/bm/ontology [R=302,L]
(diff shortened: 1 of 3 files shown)


fortiss-itsm:
README.md updated

dgarijo:
@fortiss-itsm http://213.136.65.6:8080 doesn't look like a stable URL. Can you please place your resources somewhere else? Github is a possibility...

davidlehn:
Also please squash your PR into one commit.

fortiss-itsm:
@dgarijo thank you very much for your feedback. I changed the URL and also squash my pr into one commit @davidlehn

dgarijo:
@fortiss-itsm you seem to want to publish the ontology according to the W3C best practices (www.w3.org/TR/swbp-vocab-pub/). However, you are not doing content negotiation correctly (you only add a 302 redirection, instead of a 303).  
If you want a reference of how this can be done, you can look at this vocabulary: https://github.com/perma-id/w3id.org/blob/master/games/spec/.htaccess.

If you don't want to enable content negotiation, please confirm and I will merge the pull request as it is.

fortiss-itsm:
@dgarijo I apologize for the inconvenience. At the moment i am fine without no content negotiation, so you can merge the pull reguest as it is. thank you very much for your x check

PR #414, opened 2016-06-01
Correct "Create w3id.org for personal homepage"

holgerst:
Please revert my previous pull request because it creates a file "holger" in the directory "people" instead of creating a directory of that name (with a ".htaccess" that contains the contents of the current file "holger").


holgerst:
@dgarijo Thanks a lot for merging! ...and sorry for the noise! :-S

dgarijo:
@holgerst no worries ;)

PR #422, opened 2016-06-06
Move resolve location to archive.org

phillord:
Archive provides a better long term solution for archiving, than my own website!


phillord:
Please ignore!

PR #432, opened 2016-06-13
added default response, for some reason opening http://w3id.org/aufx/…

muddymudskipper:
…ontology/1.0 in Protege (5) gives error

any ideas why this is not working with Protege? this .htaccess works with localhost, but not with the w3id url.

### DIFF
--- a/aufx/.htaccess
+++ b/aufx/.htaccess
@@ -26,4 +26,4 @@ RewriteCond %{HTTP_ACCEPT} text/turtle
 RewriteRule ^ontology/(.*) http://eecs.qmul.ac.uk/~thomasw/aufx/ontology/$1/aufx.ttl [R=303,L]
 
 # Choose the default response
-#RewriteRule ^ontology$ ontology.owl [R=303]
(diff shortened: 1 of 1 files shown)


dgarijo:
Yeah. This happens because Protege doesn't like the http->https redirection. You have to change the URI of you ontology to https://w3id.org/aufx/ontology/1.0. Then it works (at least with version 5.0.0). Previous Protege versions won't load the ontology because an issue when dealing with https.
It is not a redirection issue; it's an issue with Protege.

dgarijo:
@muddymudskipper I could load your ontology in version 5.0.0. Do you still want to merge the pull request? I am afraid the result will be the same.

muddymudskipper:
so it worked? I could change the URI to https if not.

muddymudskipper:
anyway, thanks for the help, i just change it and make another pull request

muddymudskipper:
alright, this works, please merge.

muddymudskipper:
I added a default response, that's all in this.

dgarijo:
@muddymudskipper I loaded **https**://w3id.org/aufx/ontology/1.0 in Protege 5 successfully, yes (see below). Since the redirection works, can I just close this pull request?


protege

muddymudskipper:
yes, thanks

PR #439, opened 2016-06-21
Pull request for redirected ontologies

nimonika:
please could you accept these two pull requests


dgarijo:
Both URLs lead to a 404. Could you please fix that?

nimonika:
Apologies, my bad. All working now.

dgarijo:
Merged. Just for your information, you can do the content negotiation directly in this htaccess instead of doing a 302 and then a 303.

PR #441, opened 2016-06-22
Update of .htaccess

anuzzolese:
Request to enable permanent URLs for stlab.

### DIFF
--- a/scholarlydata/.htaccess
+++ b/scholarlydata/.htaccess
@@ -2,3 +2,4 @@ Header set Access-Control-Allow-Origin *
 Options +FollowSymLinks
 RewriteEngine on
 RewriteRule ^/?$ http://www.scholarlydata.org [R=302,L]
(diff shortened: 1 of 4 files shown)


dgarijo:
This branch has conflicts with the main branch. Please solve the conflicts, otherwise I cannot merge

dgarijo:
The url http://150.146.207.10 does not seem stable. Can you please move it to some stable URL?

anuzzolese:
Hi @dgarijo,

what's the problem with such an URL?

dgarijo:
The URL resolves, but it's a fragile setup. We recommend having the URLs using DNS in order to avoid constant updates to w3ids.

dgarijo:
@anuzzolese, did you have the chance to look at the DNS thing? I don't want to be a blocker, but it is better for the maintenance of the repo. If you can't find a proper place, I could merge the branch...

anuzzolese:
@dgarijo I changed the target URL.

Issue #394, opened 2016-05-18
rawgit usage


davidlehn:
There are a handful of .htaccess uses of https://rawgit.com/.

I think people use this service to deal with the lack of content-type headers on github servers, but I'm not sure.  It's possible some uses just copy & paste from other projects that needed that feature.

The files with test URLs that hit `rawgit.com` are causing travis to have SSL failures.  However, it works on `cdn.rawgit.com`.  The SSL issue may be related to their cloudflare usage?  It works on local tests so it has something to do with the travis test environment.

Additionally, according to the `rawgit.com` homepage, `rawgit.com` itself is to be used *only* for development purposes and has short caching time.  I would suggest just changing all the uses to `cdn.rawgit.com` but that has permanent caching!  I'm unsure if the handful of users using that host are aware their redirects are cached forever.  It looks like all targets but one are not using versioned resources, so I suspect not.

The .htaccess owners using rawgit should be made aware of this issue.

muddymudskipper:
ok

pietercolpaert:
Rawgit now also has a in production option and has HTTPS. It’s a quick and convenient way to hosting small RDF files I believe.

davidlehn:
**WARNING**: As of Oct 2018, RawGit is closed to new users and will only serve traffic for current users until at least Oct 2019.  https://rawgit.com/

PR #446, opened 2016-06-28
add rules for IEC62264 vocabulary

lavdim:


dgarijo:
Please avoid using raw.github links. See: https://github.com/perma-id/w3id.org/issues/394
Could you please replace them with github pages links?

lavdim:
ok, thanks, we will fix and resubmit the changes again.

PR #448, opened 2016-06-30
Adding redirection for the RAMI and the IEC 62264 vocabularies

igrangel:


dgarijo:
A couple of comments:
1) You were publishing previously under eis/vocabs. Now you are creating a single folder for each of the vocabularies, are you aware of that?
2) You seem to have copied the redirection for the aml vocabulary. As it is, the w3id will resolve this URI: w3id.org/i40/iec62264/**aml** (The same thing with the rami vocabulary). I think you may want to fix this

yoangutierrez:
Hi, I did not know that I was accessing to aml. My intension is to create only two folders: nlp and media, for publishing ontologies. Sorry for the inconveniences

yoangutierrez:
I experienced some errors in the github tool which maybe could provoke these conflicts. 
Regarding the comment number two, do I have to fix them or they were not merged yet?

dgarijo:
Another PR has been created addressing this issue.

PR #449, opened 2016-06-30
Create nlp

yoangutierrez:
ONTOlegolangUAge first version uploaded. Redirection to GPLSI


dgarijo:
@yoangutierrez , 
You are creating a file called nlp instead of .htaccess in a folder called nlp. Please have a look at the many examples we have in the repo to see how this would be done. Can I close this pull req?

PR #455, opened 2016-06-30
Configurations for I40 vocabularies

igrangel:


dgarijo:
@igrangel 
you added redundant htaccess. You don't need a folder for each vocabulary within i40 if you are already doing an htaccess there. Also, for the vocabualries in the folders you are not referencing to the target urls you want to achieve, i.e., you are looking for something like i40/aml/aml instead of i40/aml.

igrangel:
I have removed the general htaccess. So, I created the structure in folders. Is this right?

PR #467, opened 2016-07-11
addition of new redirections for mica project

gloucklegnou:


dgarijo:
@gloucklegnou 
All the new added links seem to lead to a 404. Please fix them.

gloucklegnou:
I've made some corrections to fix the links that lead to 404 error. 
for https://w3id.org/mica/resource/* redirections this redirects to a web service that produce the response on the fly. In the readme.md fil there is an example of such a resource.

PR #473, opened 2016-07-13
Update for vocab redirect

erickorb:
Added vocab redirect pages.


dgarijo:
I think the url you want to redirect to is wrong.  Should it be: https://truecred.com/vocabs/truecred
instead of https://truecred/vocabs/truecred?

erickorb:
Yes. I'll fix.

erickorb:
Update for URL on Vocab

PR #482, opened 2016-07-19
Semantic File System Ontology

Syed-Rahman-Mashwani:
Semantic File System Ontology

### DIFF
--- a/SFS-Ontology
+++ b/SFS-Ontology
@@ -1,0 +1,2724 @@
+<?xml version="1.0"?>
+
+
+<!DOCTYPE rdf:RDF [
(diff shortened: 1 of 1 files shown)


dgarijo:
@Syed-Rahman-Mashwani 
w3id.org is not designed to store your ontologies, you should store them in your personal repositories. Here you should add a folder with an .htaccess and edit the redirections. See: https://github.com/perma-id/w3id.org#adding-a-permanent-identifier-to-w3idorg

Syed-Rahman-Mashwani:
Thanks.

Can I use sourceforge.net to upload my ontology and then link to w3id?

Regards
Syed Rahman

PR #489, opened 2016-07-25
Update .htaccess

Syed-Rahman-Mashwani:


dgarijo:
You  are redirecting to text/turtle, but the ontology in rdf/xml.
You should change the line  ^.*text/turtle.*  to ^.*application/rdf\+xml.*

PR #491, opened 2016-08-02
added earth-science metadata extension for ro ontologies

rapw3k:


davidlehn:
Are you aware of the issues related to using rawgit?
https://github.com/perma-id/w3id.org/issues/394

PR #497, opened 2016-08-19
Pull request for skg ontology

sadeghiafshin:
Hereby I am requesting for a pull, to add skg ontology
that we are making for a scientific knowledge graph at the
University of Bonn.
Contact address: sadeghi@uni-bonn.de
Bests,
Afshin Sadeghi

PR #498, opened 2016-08-25
Create w3id for blockcerts

kimdhamilton:
Create w3id of blockcerts for blockcerts.org with .htaccess rules.

### DIFF
--- a/bc/.htaccess
+++ b/bc/.htaccess
@@ -1,0 +1,5 @@
+RewriteEngine on
+
+# JSON-LD context
+RewriteRule ^$ http://www.blockcerts.org/bc/1.2.0/context.json [R=302,L]
+
--- a/bc/.htaccess
+++ b/bc/.htaccess
@@ -1,5 +1,7 @@
+Header set Access-Control-Allow-Origin *
+Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
+Options +FollowSymLinks
 RewriteEngine on
-
-# JSON-LD context
-RewriteRule ^$ http://www.blockcerts.org/bc/1.2.0/context.json [R=302,L]
+RewriteRule ^$ https://www.blockcerts.org/ [R=302,L]
+RewriteRule ^context$ http://www.blockcerts.org/bc/1.2.0/context.json [R=302,L]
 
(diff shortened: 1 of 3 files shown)


kimdhamilton:
The build failures in this PR (and recent PRs #497 and #499) do not appear to be the cause of the failure.

There are errors like the following, which suggest problems with existing namespaces. However, I'm not seeing failures there either. For example, https://w3id.org/rdw/vehicles resolves correctly with no errors that I see

URL        `http://localhost/rdw/vehicles#'
Real URL   http://localhost/rdw/vehicles
Check time 0.063 seconds
Info       Redirected to
           `https://rawgit.com/dvh/KentekenAPI/master/vocabulary.json'.
Result     Error: SSLError: [Errno 1] _ssl.c:504: error:14077438:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert internal error

davidlehn:
Ignore the errors for other rules.  Mostly all related to problems using the rawgit service.

Note the main redirect to https://www.blockcerts.org/ (vs http) is using a github cert I think.  Fails the name check.

I took a peek at the context.  Not quite sure where that's headed.  No time for full review, but some quick comments:
- It's just a copy of the thing dc is pointed at?  You can load other contexts if that's what  you're trying to do.
- "dc" is often used as a prefix for dublic core, could cause confusion.
- "id" is *very* often used in JSON-LD as an alias for "@id", but you are defining it as dc:id.  Very likely to cause confusion.
- The "validation" properties are using "validatesType" and "validationSchema" which are not defined.
- "familyName" and "givenName" should just have "@id":"schema:familyName" and "schema:givenName", not "schema:Person#xxxName".
- "schema:text" appears to be a property, not a datatype. See https://schema.org/text.  You can just leave that "@type": "schema:text" part out all over since properties will default to be strings, which is probably what you want, and I would hope is compatible with the https://schema.org/Text datatype.

kimdhamilton:
Thanks for the review. Actually, I didn't realize this would be a review for the JSON LD context as well -- just the redirects. That is still being updated (e.g. we were originally hoping for a w3id of 'dc' but that was taken already). I'll go ahead and get it in the final form assuming we can have blockcerts

davidlehn:
Review was just because I was curious and looked at the redirects.

kimdhamilton:
this was great feedback -- I hadn't noticed the missing validation properties. 

I updated the [context.json](http://www.blockcerts.org/schema/1.2.0/context.json) with your feedback from above. I'm curious -- are there any tools you use to catch issues like this? Or just visual audits? I've been relying on JSONLD playground so far, but wondering if there's something more targeted to catching issues.

davidlehn:
Mostly a visual inspection.  We at Digital Bazaar helped create JSON-LD and use it a lot so have experience finding issues in our own data. :-)  Seeing the dc and id issues was just from dealing with many files over time.  I saw the non-context validation data so threw the file into the playground and looked at the output triples.  Saw some of the data missing.  Better tools are needed for this sort of thing, but for small files you can often pick up common issues with a quick look.

PR #499, opened 2016-08-25
Need to publish OntoDocRel ontology using w3id

QamarPC103006:
I need to publish an ontology OntoDocRel using w3id service, please allow me.


davidlehn:
Please fix this to not remove the root README.

davidlehn:
That README change is still wrong.  Your README should be in OntoDocRel/README.md.  It currently is updating the main site README.md.  I'm guessing you are trying to fix this in #501?  I'll comment there too.

PR #500, opened 2016-08-25
change redirect to http

kimdhamilton:
@davidlehn thanks for the correction!

I'm working on the json ld context updates now

### DIFF
--- a/blockcerts/.htaccess
+++ b/blockcerts/.htaccess
@@ -2,7 +2,7 @@ Header set Access-Control-Allow-Origin *
 Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
 Options +FollowSymLinks
 RewriteEngine on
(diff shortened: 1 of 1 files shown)


blockcerts/README.md
```diff
@@ -3,12 +3,12 @@
 
 Homepage:
 
-https://w3id.org/blockcerts
+http://w3id.org/blockcerts
```
davidlehn: Leave this as https.  w3id.org redirects all http traffic to https so you should avoid the extra redirect.

blockcerts/README.md
```diff
@@ -3,12 +3,12 @@
...
 
 
 JSON-LD contexts:
 
-https://w3id.org/blockcerts/context
+http://w3id.org/blockcerts/context
```
davidlehn: Leave as https.

kimdhamilton:
I reverted the last commit so it's just the .htaccess change now.

Thanks again for all the help and pointers!

PR #507, opened 2016-09-01
Added PURL for the SemIoT Ontology

KMax:


dgarijo:
Please don't use rawgit (see https://github.com/perma-id/w3id.org/issues/394).
Instead, you could use github pages.

KMax:
> Please don't use rawgit (see #394).

@dgarijo I don't. I use `raw.githubusercontent.com`.

PR #509, opened 2016-09-01
Updated .htaccess file

QamarPC103006:
I have updated .htaccess file for OntoDocRel to change the location of my ontology and its documentation files. I have uploaded these files on github


dgarijo:
@QamarPC103006 
if you serve the content this way, the user will not see the html documentation, just the github html for that file. The same for the ontologies, they won't be imported successfully in programs like Protege.
If you want to serve properly the html you should create a github page of your project, and then copy the links from it. Alternatively, you can generate the documentation, diagrams, etc. with OnToology (http://ontoology.linkeddata.es/), which will push a github page to your repo with everything (this is a bit of self promotion, but it may help you accomplish your objective).

QamarPC103006:
ok thanks for guideline, i am closing this pull request

PR #510, opened 2016-09-01
Added PURLs for Hydra PubSub and Hydra Filter ontologies

KMax:


semiot/ontologies/hydra/.htaccess
```diff
@@ -0,0 +1,4 @@
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^hydra-pubsub$ https://raw.githubusercontent.com/semiotproject/ontologies/master/hydra-pubsub/hydra-pubsub.ttl [R=303,L]
+RewriteRule ^hydra-filter$ https://raw.githubusercontent.com/semiotproject/ontologies/master/hydra-filter/hydra-filter.ttl [R=303,L]
```
davidlehn: These will redirect https://w3id.org/semiot/ontologies/hydra/hydra-xxx, did you mean for the rule to just be `^xxx$` like the README implies?

davidlehn:
Your README uses http://w3id.org/ vs https://w3id.org/ and I see that's also used in the content you redirect to.  Just so you know, that adds an extra redirect step since w3id.org redirs all http requests to https first.

KMax:
@davidlehn Thank you for your comments! I fixed the .htaccess and switched to HTTPS.

PR #516, opened 2016-09-13
yago: added w3id.org/yago/mpi/ as redirect to MPI projects

thomasrebele:


yago/.htaccess
```diff
@@ -6,3 +6,4 @@ RewriteRule ^spotlx(.*)$ https://gate.d5.mpi-inf.mpg.de/webyagospotlx/WebInterfa
 RewriteRule ^svgbrowser(.*)$ https://gate.d5.mpi-inf.mpg.de/webyago3spotlx/SvgBrowser$1 [R=302,L]
 RewriteRule ^(.*)$ http://www.mpi-inf.mpg.de/departments/databases-and-information-systems/research/yago-naga/yago/$1 [R=302,L]
 
+RewriteRule ^mpi(.*)$ http://www.mpi-inf.mpg.de/departments/databases-and-information-systems/research/yago-naga/$1 [R=302,L]
```
davidlehn: The `^(.*)$` rule above will match the `mpi` prefix first so this rule will never run.  If you want to handle specific cases, put them before the default wildcard rule.

thomasrebele:
thanks for noting it! I'll make a new pull request

PR #520, opened 2016-09-16
W3id/people/niloceth

Niloceth:
Adding w3id redirect to personal page since I can no longer update purl.org/colinleath


davidlehn:
@Niloceth Note that cdn.rawgit.com will set headers to permanently cache your content.  See #394, docs at https://rawgit.com/, and look at the headers returned from your URL.

Niloceth:
Thank you!

PR #523, opened 2016-09-17
Linking /niloceth to rawgit.com not cdn.rawgit

ColinLeath:
It's been pointed out that 
Note that cdn.rawgit.com will set headers to permanently cache your content. See #394, docs at https://rawgit.com/, and look at the headers returned from your URL.


ColinLeath:
Deleted this line. . . I was niloceth, but will stick with colinleath

PR #526, opened 2016-09-21
Add NC|nocase flag to ^colinleath

ColinLeath:
Would like ColinLeath, COLINLEATH, etc to match as well.


dgarijo:
Hi @ColinLeath,
could you please avoid using rawgit.com? See https://github.com/perma-id/w3id.org/issues/394. An alternative is to use github pages for publishing your page.

ColinLeath:
Thank you for pointing out pages.github . . I wasn't aware of it.

dgarijo:
No problem!

PR #562, opened 2016-12-06
Create .htaccess

m-markovic:
A redirect to an ontology and its documentation


dgarijo:
Can you please add a readme file with a little description and the maintainer?
Thanks in advance

m-markovic:
I have added a short readme file

dgarijo:
Thanks for adding the readme.
I have checked your request and you are using rawgit. This kind of services have caused SSL problems in the past: https://github.com/perma-id/w3id.org/issues/394. In addition, they are supposed to be used for development purposes, not for accessing stable resources. Could you please use an alternative like Github pages?
Sorry for the trouble.

m-markovic:
Hi Daniel,

If everything is ok now with the PRISM ontology I will also fix the usage of rawgit link in SC-PROV and FS-PROV which have already  been merged some time ago.

PR #570, opened 2016-12-20
content description ontology cdesc

andresgs77:
Redirecting to the content description ontology


dgarijo:
Please, add a readme file indicating the maintainer of the redirection. For example: https://github.com/perma-id/w3id.org/tree/master/bibma

A few comments:
1) The URI of your ontology is : http://w3id.org/contentdesc#. You are creating the redirection of http://w3id.org/cdesc. If you try to resolve http://w3id.org/contentdesc with the current htaccess, you will get a 404. 

2) Your ontology uses http instead of https. This will work because w3id redirects http to https, but I have experienced some problems in the past when loading the ontology in Protege this way. I suggest changing the URI to https (https://w3id.org/cdesc).

andresgs77:
Yes, I was going to revert the pull request when I realised that the ontology name does not correspond to what I define in w3id but you were faster than me.

dgarijo:
Note that you haven't changed the URI of the ontology in the term documentation (e.g. http://w3id.org/contentdesc#Concept)

andresgs77:
yes I have, it should be the cache.

PR #584, opened 2017-01-23
Fixing w3id.org/i40/sto redirection

paulbaptista:


dgarijo:
I recommend changing the redirection to 301 if you do the 303 on the server. Otherwise you are adding a double 303 for your requests.

paulbaptista:
Thanks for your recommendation. I'll do it that way

PR #585, opened 2017-01-23
Create ppdb redirect

luigi-asprino:


dgarijo:
@luigi-asprino,
this redirection will not work because you haven't added an htaccess file. Please look up the guidelines in the readme file: https://github.com/perma-id/w3id.org/blob/master/README.md

PR #593, opened 2017-01-30
Charity Organization Ontology

faiqmiftakhul:
Charity Organization Ontology


dgarijo:
@faiqmiftakhul,
you don't seem to be serving the content correctly. By pointing to the github project page, you are serving HTML in all cases, not rdf, as you intend. This way the ontology will NOT load in Protege, for example. You can address this issue by creating a Github page in your project and storing your rdf files there.

PR #604, opened 2017-02-13
Adding base URL for UN URLs

kohsah:
w3id.org placeholder for UN system redirectors (subsequently ontology)

### DIFF
--- a/UN/.htaccess
+++ b/UN/.htaccess
@@ -1,4 +1,4 @@
 Options +FollowSymLinks
 RewriteEngine on
 RewriteRule ^$ https://www.un.org/ [R=301,L]
-RewriteRule ^FAO/?$ https://www.fao.org [R=301,L]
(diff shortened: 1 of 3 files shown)


UN/.htaccess
```diff
@@ -0,0 +1,4 @@
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^$ https://www.un.org/ [R=301,L]
+RewriteRule ^FAO/?$ https://www.fao.org [R=301,L]
```
davidlehn: The https link is 404.  Is that a temporary problem or should this be http?

kohsah: Oops seems like the www.fao.org is a http site, it should be http then.

kohsah: Should I send another PR with the fixed .htaccess ?

davidlehn: You can update this one.  Either just commit a correction, or add a commit and squash the two into one commit and force push the result.

PR #619, opened 2017-03-07
Add .htaccess cpf-rel

marilenadaquino:


dgarijo:
@marilenadaquino,
please add a readme file describing who will maintain this redirection and a small description.
Also, the htaccess does not seem to be right. You have duplicated the redirection to the owl.

PR #624, opened 2017-03-09
Added .htaccess cpf-rel

marilenadaquino:


dgarijo:
I will accept this one and close the previous two.

PR #644, opened 2017-03-28
update redirection to catch various content types

djodjoni:
relates to https://github.com/valueflows/valueflows/issues/207

ping @elf-pavlik


dgarijo:
Please, don't use rawgit. It may cause problems when checking the availability and it's not a solution designed for in-production links. Instead you could use a gh-page.

elf-pavlik:
Thanks @dgarijo we'll update this PR!

djodjoni:
done

davidlehn:
You should probably use https for github.io links.

djodjoni:
@davidlehn yep probably yes for all. nice catch thx!

PR #645, opened 2017-03-29
Add htaccess file for educational sectors vocab

acka47:


dgarijo:
Hi @acka47,
please add a readme file with your redirection. The readme file should state who is the maintainer of the vocabulary and a small definition.
Also, please do NOT use raw.githubusercontent. That is not a stable link for referring to your vocabulary, and it leads to problems when validating the links. Instead you could use a github page.

acka47:
Thanks for the feedback. I added a readme file. @literarymachine is currently working on a vocabulary service as replacement for raw.githubusercontent that will serve HTML alongside RDF. It will also be used for https://github.com/perma-id/w3id.org/tree/master/class and https://github.com/perma-id/w3id.org/tree/master/lobid where we currently use raw.githubusercontent. It would be great if you merged this for now. We will update the redirects soon.

dgarijo:
@acka47 
no, I won't use it because it causes errors in the validation of the project: https://github.com/perma-id/w3id.org/issues/394. You can easily create a github page to address this issue (https://help.github.com/articles/what-is-github-pages/). It takes very little time and addresses the whole publication of the vocabulary.

acka47:
Fair enough. I will update the redirects when we have another solution than raw.githubusercontent.

davidlehn:
No updates. Closing. Reopen if needed.

PR #655, opened 2017-04-10
Update to new imsglobal.org location, add README

mgylling:
Since January 2017, the Open Badges spec is maintained by IMS Global
Learning Consortium. The spec & context URLs are therefore changed as
we turn the openbadgespec.org site into a redirect to imsglobal.org.

### DIFF
--- a/openbadges/.htaccess
+++ b/openbadges/.htaccess
@@ -1,7 +1,9 @@
 Options +FollowSymLinks
 RewriteEngine on
-RewriteRule ^$ https://openbadgespec.org/ [R=302,L]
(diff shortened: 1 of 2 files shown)


davidlehn:
@mgylling: This change is causing problems for people who need CORS support.  Is it possible to add CORS headers on the new server?  Specifically "Access-Control-Allow-Origin: *".  Thanks.

mgylling:
@davidlehn, yes we are working on it right now. Sorry for the disruption.

davidlehn:
Reverted CORS related changes in https://github.com/perma-id/w3id.org/pull/657.

PR #670, opened 2017-04-24
Content negotiation 

nandadorea:
The initial submission did not address proper content negotiation, as pointed out by the admins when it was accepted. I thank them for their help and patience. This is an attempt to correct the error - but I am very new to ontology writing and hosting, and if further error are detected on the current code, I will appreciate any feedback. 

I also removed the excessive details on Readme, as also pointed out.


nandadorea:
I am sorry for demanding any extra help, but I do not understand why the checks are failing, all the links provided work, I'd appreciate any help if the coding in htaccess is wrong.

dgarijo:
Now the redirection works, but http://datadrivensurveillance.org/ahso/ahso.owl is an html, not the OWL ontology. You should just replace that document with the ontology and the content should be served correctly.

PR #676, opened 2017-05-03
add access to illusnip

dxliuNJU:
illusnip is a online dataset snippet extrator.

### DIFF
--- a/illusnip/.htaccess
+++ b/illusnip/.htaccess
@@ -1,0 +1,4 @@
+Options +FollowSymlinks
+RewriteEngine on
+RewriteRule ^$ http://47.91.88.78:8080/illusnip/ [R=303,L]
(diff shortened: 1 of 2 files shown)


illusnip/.htaccess
```diff
@@ -0,0 +1,4 @@
+Options +FollowSymlinks
+RewriteEngine on
+RewriteRule ^$ http://47.91.88.78:8080/illusnip/ [R=303,L]
+RewriteRule ^(.+)$ http://47.91.88.78:8080/illusnip/ [R=303,NE]
```
davidlehn: Is this IP+port expected to be the same for a relatively long time?

dxliuNJU: Yes,they are expected to last for at least one year.

dgarijo: I will merge it then

PR #677, opened 2017-05-03
Patch 3

nikolaosLagos:


dgarijo:
Please don't use https://raw.githubusercontent.com/. It leads to problems when doing the validation (see https://github.com/perma-id/w3id.org/issues/394) and it is not meant to be a production link. Instead, you could use Github pages (https://pages.github.com/) to host your vocabulary

davidlehn:
raw.githubusercontent.com is probably ok.  It's rawgit.com that's more of an issue.  Still will have issues with not getting an appropriate content type for exotic types.

dgarijo:
Shall I merge the PR then? I still think that useing gh-pages is a cleaner approach...

PR #678, opened 2017-05-03
premon: added .htaccess and README for the PreMOn project

rospocher:


rospocher:
The Travis build says fail but it doesn't seem to be related to the content I've added

PR #683, opened 2017-05-08
added jpcik personal redirect

jpcik:


dgarijo:
@jpcik,
if you are redirecting everything to the same page, why not just add the default redirection (last line) instead of the individual ones?

jpcik:
True, bad copy pasting. will fix it

PR #696, opened 2017-05-12
Add MIDAS ISG Documentation rewrite

maxsibilla:
Add MIDAS ISG Documentation rewrite


davidlehn:
I think w3id.org should reserve "docs" for its own use.  Can you please choose a more project specific name?  Thanks.

PR #714, opened 2017-06-07
Add htaccess file for HOBBIT research project to make semantic web URIs dereferencable

MichaelRoeder:


MichaelRoeder:
Although the check itself is failing, it can be seen from the log that the errors are not caused by the added `https://w3id.org/hobbit` URLs. (Simply search for `hobbit` to find messages related to this pull request)

PR #728, opened 2017-07-05
Update .htaccess

megankatsumi:
-Added remaining iCity sub-ontologies.
-Update to IRI format.
-Update to rewrite rules to account for Ontology IRI and Version IRI.
-TO DO: (in source git update file path to distinguish between future versions)

### DIFF
--- a/icity/.htaccess
+++ b/icity/.htaccess
@@ -10,12 +10,158 @@ AddType application/rdf+xml .owl
 
 RewriteEngine on
 
-#Rewrite rules for iCity-Activity
+#Rewrite rules for iCity-Activity (Ontology IRI and Version IRI)
 #default response: owl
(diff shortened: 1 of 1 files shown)


davidlehn:
You might consider using wildcards to avoid the repetition in all those rules.  I think the entire file would work now with just the below two rules.  First rule would cover any alpha name and an optional owl extension.  Second would cover the current (lack of) versioning, but `$2` could be used in the result URL if appropriate once you figure out a naming scheme.

```
RewriteRule ^iCity-([A-Za-z]*)(\.owl)?$ https://raw.githubusercontent.com/EnterpriseIntegrationLab/icity/master/iCity-$1.owl [R=302,L]
RewriteRule ^iCity-([A-Za-z]*)/([0-9.]*)/$ https://raw.githubusercontent.com/EnterpriseIntegrationLab/icity/master/iCity-$1.owl [R=302,L]
```

megankatsumi:
Thank you, David! That's very helpful.

PR #731, opened 2017-07-06
Update .htaccess

megankatsumi:
-removed requirement for "iCity-" prefix
-added optional "/" suffix


davidlehn:
Did you mean to have the rules match "iCity-Foo"?  If so, you should update the char set to match "-" too: `[A-Za-z-]`.

megankatsumi:
Fixed, thanks! Apologies for all of the revisions!

PR #734, opened 2017-07-13
Redirect broken TrueCred link to openbadgespec.org

mgylling:
Fixes https://github.com/openbadges/openbadges-specification/issues/154


ottonomy:
This looks correct to me and resolves an active problem with a broken Open Badges context.

davidlehn:
That expired cert could certainly cause troubles so I merged this now.  We might need to check around to make sure this change is otherwise ok with everyone.  Since the content is the same it seemed best to merge quickly first.

PR #742, opened 2017-07-21
We removed our deprecated ontology OTLA and update TTLA htaccess

mintoman:


davidlehn:
I'm not sure what the policy is on removing ids.  The idea was these ids would be long lived as in years to decades.  It's possible other people could depend on ids to keep working.  I'll merge and we can resolve issues if they come up.

PR #743, opened 2017-07-21
Update .htaccess

iesnaola:
Redirect to HTML or OWL file depending on the way accessing the URL.


iesnaola:
How can be the content negotiation done correctly?

davidlehn:
The apache configs will allow minimal content negotiation.  There are some other projects in the repo you could look at and adapt from.  Is this tested and ready to merge?  It would also be nice if the commits were squashed.

iesnaola:
Hi. I have adapted the content-negotiation from other projects in the repo. It is ready to merge. Thanks in advance

PR #748, opened 2017-07-26
Update .htaccess

jhaag75:
Update to return JSON to browser for context doc


dgarijo:
@jhaag75 
It seems like you have added the an existing id. Maybe it's a typo?

jhaag75:
It's not a typo. Need this id to return JSON-LD if requested no matter what as it's planning to be used for a persistent JSON-LD @context id.

dgarijo:
The first rule overwrites the second rule, right?
Then why have the second?

jhaag75:
Does it? If so I will remove and resubmit.

dgarijo:
Line 42 states to return profile-context.jsonld no matter what you ask for. Then line 44 states to return that very same file when you ask for json-ld

PR #751, opened 2017-07-29
new project FOODIE with links to two ontologies

rapw3k:


davidlehn:
Are you aware https://rawgit.com/xxx is just for "development"?
See docs on their front page and https://github.com/perma-id/w3id.org/issues/394

rapw3k:
I fixed the targets

PR #774, opened 2017-08-23
DataFabric.cc's root folder. FTS ontology PURL

KMax:
+ Root folder for PURLs related to DataFabric.cc projects,
+ PURL for the FTS ontology.


KMax:
Checked the logs of the build. The PURL added by me works fine.

dgarijo:
I suggest using a github page for publishing the ontology. The raw.githugusercontent are fine, but are more for development.

PR #781, opened 2017-08-25
Update .htaccess

megankatsumi:
test of updated alternate html files


dgarijo:
I will accept the PR, but note that if someone requests xml, you are returning the documentation instead of the RDF/XML of the ontology..

PR #790, opened 2017-09-01
Add negotiation for personal URI redirect

idafensp:


dgarijo:
For adding yourself as a person please do so under the /people path. See:
https://github.com/perma-id/w3id.org/tree/master/people
It's the convention we are following for everyone

PR #806, opened 2017-09-26
redirects to go from w3id to cedar.openmadrigal.org

billrideout:
We are adding the ability to cite scientific data from the CEDAR Madrigal database, and would like to allow those citations to have the option of changing the server hostname.  Thanks - Bill Rideout (brideout@mit.edu)

### DIFF
--- a/cedar/.htaccess
+++ b/cedar/.htaccess
@@ -1,0 +1,3 @@
+RewriteEngine on
+
+RewriteRule ^cedar(.*)$ http://cedar.openmadrigal.org/showExperiment/$1 [R=302,L]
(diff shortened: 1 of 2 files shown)


cedar/.htaccess
```diff
@@ -0,0 +1,3 @@
+RewriteEngine on
+
+RewriteRule ^cedar(.*)$ http://cedar.openmadrigal.org/showExperiment/$1 [R=302,L]
```
davidlehn: This is duplicating the current directory "cedar" and would match /cedar/cedar*.  You probably mean to do "^(.*)$".

billrideout:
Thank you. I have changed ^cedar(.*)$ to ^(.*)$

PR #816, opened 2017-10-22
create SCL

courtneykase:
pull request to set up identifier and redirect


davidlehn:
This needs to go in a `.htaccess` file in directory named "SCL".

courtneykase:
I tried again with Create .htaccess #1
 Open	courtneykase  wants to merge 1 commit into master from courtneykase-scl-patch-1

PR #818, opened 2017-10-24
[NEW] Dataset Descriptor Ontology

psiotwo:


davidlehn:
1) Could you please also add a README.md with at least contact info.
2) I know it's basically first come first served here but there might be a potential future name conflict with another unrelated "DDO" project.  I'm not sure what to do about that or if it's an issue here or not.  Maybe @dlongley or @msporny could comment.

psiotwo:
1. OK, Done. Originally, I took only the "security" dir, where no README.md is. You should consider update the guide to referr to another dir (not "security") that contains README.md.
2. DDO is a brand we use for some time already as well. I don't know what can I do about it.

dgarijo:
Since @psiotwo is not reserving the w3id but using it for an actual ontology, I don't see any reasons why we should oppose. 
@psiotwo note that you are not doing content negotiation on the ontology at the moment.

PR #819, opened 2017-10-24
CREATE .htaccess

ghost:
Would like  w3id.org/SCL to redirect to  https://www.qualtrics.com/
Did I create the correct file and use the correct code?

### DIFF
--- a/SCL/README.md
+++ b/SCL/README.md
@@ -1,2 +1,0 @@
-Piloting surveys and learning programming
-Contact: courtneyannkase@gmail.com


davidlehn:
Are these survey links intended to be long lived?  As in something that needs to work for years or even decades.  I'm wondering if perhaps a traditional URL shortener service would be more appropriate for your use case.

courtneykase:
They are not going to live long.  Very good point you make  Do you have recommendation like google url shortener or other?

davidlehn:
I rarely have the need to shorten urls.  The classic https://tinyurl.com/ lets you pick an alias.  https://goo.gl/, https://bitly.com/, and others have stats and commercial features.  Not hard to try them all.

davidlehn:
New commits removed all changes? Closing. Reopen if needed.

PR #821, opened 2017-10-31
all redirections to single official host

bjdmeest:


davidlehn:
I see your site redirs from http to https, did you want to just use a https redir here?

Also I see on your site you have http://w3id.org/ links, but w3id.org will redir everything to https, so can avoid that redir as well by just using https links.

bjdmeest:
You are right, thanks for checking! I updated to https, both here and on the site :)

PR #822, opened 2017-11-01
Update .htaccess

megankatsumi:
Updated rewrite rules to account for revised file organization and naming.
Temporarily removed html redirect rules.


dgarijo:
Note that raw.githubusercontent urls are not meant for stable production links. I would like to suggest moving to things like github pages.

PR #848, opened 2017-12-01
HumL workshop

HumLworkshop:


dgarijo:
This PR is adding a file with the contents of an htaccess. 
You should have a folder (huml) and then add an .htaccess and readme file with the appropriate contents.
Please see an example : https://github.com/perma-id/w3id.org/tree/master/BCI-ontology

HumLworkshop:
my bad, fixed

PR #856, opened 2017-12-15
huml

AnLiGentile:
Human in the loop workshop at WWW2018


huml/.htaccess
```diff
@@ -0,0 +1,4 @@
+Header set Access-Control-Allow-Origin *
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^(.*)$ http://humlworkshop.github.io/$1 [R=302,L]
```
davidlehn: Should use https here since github will redirect it anyway.

PR #857, opened 2017-12-20
Revision of w3id.org/ibp

GeorgFerdinandSchneider:
We revised the rewrite rules of w3id.org/ibp

Added now rewrite rules dependend on the content type (ttl/html)

Also added rules for each hosted ontology

### DIFF
--- a/ibp/.htaccess
+++ b/ibp/.htaccess
@@ -12,7 +12,7 @@ AddType text/turtle .ttl
 RewriteCond %{HTTP_ACCEPT} text/turtle
 RewriteRule ^StateMachineOntology$ https://github.com/TechnicalBuildingSystems/Ontologies/blob/master/StateMachineOntology/StateMachineOntology.ttl [R=303]
 
-If application looks for .html
+# If application looks for .html
 RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+xml)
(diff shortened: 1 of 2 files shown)


dgarijo:
You are redirecting to wr3id.org/ibp/ibp/NAME, instead of the intended URL (e.g., https://w3id.org/ibp/CTRLont) I suggest you change your rules. For example:
RewriteRule ^ibp/StateMachineOntology
should be RewriteRule ^StateMachineOntology

GeorgFerdinandSchneider:
@dgarijo thanks for indicating the fault. We checked our rules with (http://htaccess.mwl.be/)[http://htaccess.mwl.be/] but did not adapt correctly to the w3id.org scheme. Sorry for that.

I did the corrections via c1a169b93668418384503be27bb479317267189b

Hope the rewrite rules are fine now. Please notify me if not. Thanks!

dgarijo:
@GeorgFerdinandSchneider 
I merged the PR, but your content still won't be served properly in some cases. You provide Github's URL instead of publishing under a github page. For example, if you go to https://github.com/TechnicalBuildingSystems/Ontologies/blob/master/StateMachineOntology/StateMachineOntology.ttl (redirection for TTL) you will be serving GitHub's html page, instead of the TTL that you want to.

I suggest having a look  at GitHub pages to publish this: https://pages.github.com/
(Or use raw.githubusercontent)

PR #861, opened 2018-01-03
Switching to raw.githubusercontent on w3id.org/ibp/...

GeorgFerdinandSchneider:
Dear w3id.org-team,

we switched our htaccess file to forward to raw.githubusercontent domain to enable serving ttl and html files of our ontologies seperatly.

Thanks to Daniel Garijo for suggesting this.

Best

Georg

### DIFF
--- a/ibp/.htaccess
+++ b/ibp/.htaccess
@@ -10,77 +10,77 @@ AddType text/turtle .ttl
 # StateMachineOntology
 # If application looks for .ttl
 RewriteCond %{HTTP_ACCEPT} text/turtle
-RewriteRule ^StateMachineOntology$ https://github.com/TechnicalBuildingSystems/Ontologies/blob/master/StateMachineOntology/StateMachineOntology.ttl [R=303]
+RewriteRule ^StateMachineOntology$ https://raw.githubusercontent.com/TechnicalBuildingSystems/Ontologies/master/StateMachineOntology/StateMachineOntology.ttl [R=303]
 
 # If application looks for .html
 RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+xml)
 RewriteCond %{HTTP_ACCEPT} text/html [OR]
 RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
 RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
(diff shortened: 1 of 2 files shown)


kidehen:
Please note that you need to assign the correct value to the content-type response header too i.e., "text/turtle" rather than "text/plain". At the very least you should designate "text/turtle" as the content type for the resource at: https://raw.githubusercontent.com/DataFabricRus/ontology-fts/master/fts/ontology.ttl

Problem demonstration:

curl -ILH "Accept: text/turtle" https://w3id.org/datafabric.cc/ontologies/fts#
HTTP/1.1 303 See Other
Date: Wed, 03 Jan 2018 13:12:00 GMT
Server: Apache/2.4.7 (Ubuntu)
Access-Control-Allow-Origin: *
Location: https://raw.githubusercontent.com/DataFabricRus/ontology-fts/master/fts/ontology.ttl
**Content-Type: text/html**; charset=iso-8859-1

HTTP/1.1 200 OK
Content-Security-Policy: default-src 'none'; style-src 'unsafe-inline'; sandbox
Strict-Transport-Security: max-age=31536000
X-Content-Type-Options: nosniff
X-Frame-Options: deny
X-XSS-Protection: 1; mode=block
ETag: "7dc1cdfd461c6658cc839e7b6f1aaeca4d9306b1"
Content-Type: text/plain; charset=utf-8
Cache-Control: max-age=300
X-Geo-Block-List:
X-GitHub-Request-Id: 3E6A:21A4:11CCB3:13A029:5A4CD667
Content-Length: 111810
Accept-Ranges: bytes
Date: Wed, 03 Jan 2018 13:12:00 GMT
Via: 1.1 varnish
Connection: keep-alive
X-Served-By: cache-ewr18139-EWR
X-Cache: HIT
X-Cache-Hits: 1
X-Timer: S1514985121.644543,VS0,VE0
Vary: Authorization,Accept-Encoding
Access-Control-Allow-Origin: *
X-Fastly-Request-ID: 869ca65d79e119e16f99174554593e5fe0e31cf5
Expires: Wed, 03 Jan 2018 13:17:00 GMT
Source-Age: 56

davidlehn:
@kidehen It's worth noting that this `ibp/` commit has same problem as the `datafabric.cc/` example you show.  The redirect response is returning HTML content along with the 303, so that should be ok.  I don't think there is a way to control the headers that raw.gihubusercontent.com returns which makes that a much more difficult problem.  Many people want a free hosting service that is in sync with github code for things like this.  What would you suggest as a solution?

PR #862, opened 2018-01-03
f4EEPSA

iesnaola:


dgarijo:
@iesnaola,
line 23 of the htaccess seems to contain a typo. I think you should remove "© 2018 GitHub, Inc."

PR #863, opened 2018-01-03
Handling of SPAR Ontologies in GitHub

essepuntato:
The htaccess has been changed in order to handle properly SPAR Ontologies in the new GitHub repository.


davidlehn:
You might consider using internal redirection vs the 303 codes.  If it's appropriate, it could avoid an extra client round trip.

PR #878, opened 2018-01-13
Add redirection to Audio Commons ontology

miguel76:
This add redirection from https://w3id.org/ac-ontology/aco to
http://www.audiocommons.org/ac-ontology/aco performing content
negotiation.
That’s the ontology we designed in the context of the European project
Audio Commons (audiocommons.org).

### DIFF
--- a/ac-ontology/.htaccess
+++ b/ac-ontology/.htaccess
@@ -1,0 +1,32 @@
+Options +FollowSymLinks
+# Turn off MultiViews
+Options -MultiViews
+
+# Directive to ensure *.rdf files served as appropriate content type,
+# if not present in main apache config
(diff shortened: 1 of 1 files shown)


dgarijo:
@miguel76 please add a readme file with the maintainer of the repo and short description of the w3id. Also, note that in the ontology you use the following URI: https://w3id.org/audiocommons/ontology#.

miguel76:
@dgarijo thanks for the quick reply, I will add the README and make a new PR. Also, the mismatch in the URI was due to a previous idea, but in the end I will go for https://w3id.org/ac-ontology/aco#. I just updated the repo accordingly.

PR #881, opened 2018-01-14
Updated redirection and list of contributos

jm-gimenez-garcia:
- Redirection changed to http://nell-ld.telecom-st-etienne.fr/

- Added Pierre Maret to the List of Contributors


jm-gimenez-garcia:
So, the failed checks are not related to my changes, right? Should I expect them to be merged or should I change anything?

PR #888, opened 2018-01-18
Disable broken linkchecker.

davidlehn:
- Many URLs fail to check, causing confusion for PRs.
- This omits many of those URLs that failed during local testing.
- Additionally, the apache setup on travis is broken.  It gets a SIGTERM.  Just disabling it all for now until fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: bash
 
-before_install:  
+before_install:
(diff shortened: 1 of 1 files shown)


davidlehn:
Note that I disabled linkchecker via commenting because I don't have access to shut travis off on github.

PR #890, opened 2018-01-19
added .htaccess and readme

marilenadaquino:


davidlehn:
I think your default rule is going to be used before those other rules.  And the other rules need to redirect to some other site.  Just starting with /xxx will redirect to another path on w3id.org.

marilenadaquino:
Thanks for your help. Hope this will be fine.

PR #908, opened 2018-02-22
Create .htaccess

Syed-Rahman-Mashwani:
I want a simple redirect to this page https://github.com/Syed-Rahman-Mashwani/360-SFS  
I have written .htaccess for this. If I did any mistake in redirecting please, I request if you could correct it. Thanks.

### DIFF
--- a/360-sfs/.htaccess
+++ b/360-sfs/.htaccess
@@ -1,0 +1,17 @@
+Options +FollowSymLinks
+# Turn off MultiViews
+Options -MultiViews
+
+# Directive to ensure *.rdf files served as appropriate content type,
+# if not present in main apache config
+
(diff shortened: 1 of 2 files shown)


dgarijo:
@Syed-Rahman-Mashwani 
please add a readme file stating the maintainer of the w3id. 
Also, take into account that while the redirect will work, you are not serving the ontology or html representation of the documentation

dgarijo:
You are using a 303 content negotiation that is typically added to semantic web resources.
Why not just do:
RewriteRule ^$ https://github.com/Syed-Rahman-Mashwani/360-SFS [R=302,L]

instead of lines 10-17?
Also please add your readme as part of this PR, I will close the other one.

PR #911, opened 2018-02-23
OCO redirection via htaccess

essepuntato:


dgarijo:
@essepuntato,
in line 11 (and others below) aren't you rewriting to a local file that doesn't exist?  (/oc/ontology)
RewriteRule ^ontology(/[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9])?$ /oc/ontology$1.html [R=303,L]

PR #915, opened 2018-02-26
Permanent URIs for the Italian OntoPiA ontology network for publ…

giorgialodi:
…ic administrations

### DIFF
--- a/italia/controlled-vocabulary/.htaccess
+++ b/italia/controlled-vocabulary/.htaccess
@@ -1,0 +1,10 @@
+Header set Access-Control-Allow-Origin *
+Options +FollowSymLinks
+RewriteEngine on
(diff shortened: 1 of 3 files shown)


dgarijo:
@giorgialodi,
why 302 instead of 303, which is meant for semantic web reosurces?

giorgialodi:
We chose 302 because it is temporary

dgarijo:
Even if temporary, according to W3C recommendations https://www.w3.org/TR/swbp-vocab-pub/ it is better to use 303 when doing semantic resource redirection. I am afraid that validators such as vapour won't validate the URI of your ontology otherwise.

giorgialodi:
we updated the .htaccess files with 303, even if we noticed other merged projects about ontologies with 302 code too.

PR #928, opened 2018-03-21
Adding improved visualiation of HTML Ontology documentation

GeorgFerdinandSchneider:
Revision of htaccess to improve the redirect to the HTML code by adding https://htmlpreview.github.io/?


dgarijo:
You may want to use github pages to expose the html of your doc instead of the current approach: https://pages.github.com/

dgarijo:
You may want to use github pages to expose the html of your doc instead of the current approach: https://pages.github.com/

PR #931, opened 2018-03-24
Update to dedicated domain

blake-regalia:


dgarijo:
@blake-regalia, 
http://gnis-ld.org/lod/ gives a 404... wouldn't you prefer to fix it before doing the redirection?

blake-regalia:
The above URLs resolves fine from my end, but it seems that for certain AS is routing to another machine on the same subnet. I will contact the servers' system administrator.

From my laptop, http://gnis-ld.org/lod/ redirects to the landing page as it should:
screen shot 2018-03-23 at 11 32 31 pm

dgarijo:
To me, it redirects to http://sejp.geog.ucsb.edu/SWJPortal//lod
which returns a 404. 
http://sejp.geog.ucsb.edu/SWJPortal seems to work, but it's not the page you are looking for, I think.

Do you still want me to accept the redirect, or do you prefer to wait for your sys admin?

blake-regalia:
Let's wait for the fix, thanks @dgarijo

dgarijo:
Ok!,
just answer to this thread when the fix is done.

blake-regalia:
Okay @dgarijo it should be all fixed now! Thanks

PR #934, opened 2018-03-27
Added SToPP. Contact: <sven.lieber@ugent.be>

SvenSvenson:
Added SToPP. Contact: <sven.lieber@ugent.be>


dgarijo:
Note that you are not doing content negotiation. Wouldn't you want to handle that in the w3id redirection?

SvenSvenson:
Thanks for noticing. Not all representations are available yet, but I handle content negotiation on a .htaccess on the destination server.

PR #935, opened 2018-03-29
added link to the Computer Science Ontology

angelosalatino:
We are releasing a new resource:

- Computer Science Ontology
- Web portal for crowdsourcing it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,3 @@
 *~
 *.Rproj.user
 *.Rproj
-cso/.DS_Store
-.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 *~
 *.Rproj.user
 *.Rproj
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -3,3 +3,5 @@
 *~
 *.Rproj.user
 *.Rproj
+cso/.DS_Store
+.DS_Store
```
stain: We definitely do not want `.DS_Store`, but only the last of these lines should be needed :)

angelosalatino:
I fixed the gitignore as you suggested

angelosalatino:
If someone reviews this, it would be very much appreciated. We have a deadline and we don't want to risk that people use a not existing link, we already provided, as effect of this pull request. 

Many thanks in advance

dgarijo:
No worries, I usually do this as soon as I see it. I hadn't realized your PR was pending.

PR #943, opened 2018-04-04
adso init

muddymudskipper:


davidlehn:
You seem to be changing the aufx dir by mistake?
Also your rules seem a bit complicated if everything is just going to the same ttl file.

muddymudskipper:
hi, thanks for checking. The aufx folder is for another ontology of mine and I did some minor changes. I'm aware that the adso rules are somewhat complicated now, I planned on changing it in the coming days if thats alright.

PR #956, opened 2018-04-09
Establishing a permanent namespace for ontologies developed in the MOEEBIUS H2020 project

GeorgFerdinandSchneider:
Dear w3id.org-team,

we would like to register a new permanent identifier under the w3id.org service for ontologies developed under the MOEEBIUS H2020 project.

> https://w3id.org/moeebius/

The identifier will be used in the ontologies developed with this project. The identifier needs to be permanent as the ontologies developed are supposed to be available and accessible forever.

Best

Georg

### DIFF
--- a/moeebius/.htaccess
+++ b/moeebius/.htaccess
@@ -1,0 +1,26 @@
+Header set Access-Control-Allow-Origin *
+Options +FollowSymLinks
+RewriteEngine on
+
+# Turn off MultiViews
+Options -MultiViews
+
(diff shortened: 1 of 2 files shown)


dgarijo:
@GeorgFerdinandSchneider,
your documentation is not rendering properly, probably because you are using https://htmlpreview.github.io/ (the javascript calls for loading the rest of the sections of the document don't work). I suggest using Github pages for publishing your ontology.

PR #957, opened 2018-04-09
Create .htaccess

ysedira:


dgarijo:
@ysedira, 
You are using 2 different redirects for the same rule. Which one is it?

ysedira:
Great catch, thanks

PR #981, opened 2018-04-25
Simplifed REC redirects

hammar:


rec/.htaccess
```diff
@@ -5,26 +5,8 @@ RewriteEngine on
 # Redirect any module w/o version in IRI to latest version (at time of writing 2.2)
 RewriteRule ^(\w+?)/(\w*?)$ http://w3id.org/rec/$1/2.2/$2 [R=303,NE,L]
```
davidlehn: While you're changing things, can you switch this to https to avoid an extra auto-redirect?

hammar: Thanks, good catch.

PR #995, opened 2018-05-15
Redirects for the DATS project

agbeltran:


dats/.htaccess
```diff
@@ -0,0 +1,5 @@
+RewriteEngine on
+RewriteRule ^$ https://datatagsuite.github.io/doc [R=302,L]
+RewriteRule ^(.*)$ https://datatagsuite.github.io/context/$1 [R=302,L]
+RewriteRule ^(.*)$ https://datatagsuite.github.io/schema/$1 [R=302,L]
```
davidlehn: The first catch-all rule will always override the second one.

davidlehn: Based on the README maybe you just mean to have one catch all rule like:
```
RewriteRule ^(.*)$ https://datatagsuite.github.io/$1 [R=302,L]
```

agbeltran: yeap, I see what you mean, thanks... will push the change

davidlehn: You might have wanted to keep that first "^$" rule?

PR #1005, opened 2018-05-25
Update .htaccess

MadsHolten:
Redirect to HTML


dgarijo:
https://raw.githubusercontent.com/w3c-lbd-cg/bot/master/bot.html returns a 404. I suggest using the Github page you have instead.
And since you are using the github page, I suggest you replace the raw.githubusercontent.com you have in there with links to the github page as well. Those links are supposed to be for development purposes

MadsHolten:
Sorry, .html was supposed to redirect to https://w3c-lbd-cg.github.io/lbd/bot

You have a good point regarding the reference to the raw-file. I will have to discuss this with the group. Probably we would also need to think of some versioning mechanism at one point, but that's the brilliant thing about w3id.org - we can leave this undecided for now.

The file should be updated now.

PR #1009, opened 2018-05-28
Update .htaccess

MadsHolten:
Added rewritebase

PR #1012, opened 2018-05-29
Update .htaccess

MadsHolten:
Added docs + moved .ttl to gh-pages


MadsHolten:
Now it looks like #1009 except that we moved the .ttl to gh-pages for this one.

PR #1024, opened 2018-06-18
added /mediatype/

nicholascar:
Simple redirection for http://w3id.org/mediatype/ to make Linked Data URIs for IANA's Media Types. The redirects are to a Linked Data API tool.


dgarijo:
@nicholascar,
these rules don't seem to be correct. Have you tried them locally first?
You can look for an example in many other repos. For instance: https://github.com/perma-id/w3id.org/blob/master/cedar/.htaccess

Note that if you add "mediatype" as part of the rule, your w3id will be https://w3id.org/mediatype/mediatype/

nicholascar:
Ooops, right you are Daniel! I silly me regarding the double /mediatype/.

I have tried them locally but only in central Apache conf, not as an .htaccess file. Will test like that and re apply a PR.

PR #1034, opened 2018-06-22
arm: change path for lode docs

zimeon:
Update to `htaccess` for https://w3id.org/arm to change documentation paths


dgarijo:
@zimeon, 
https://ld4p.github.io/arm/core/ontology/0.1/core.html leads to a 404. Could you please fix the link before doing the redirection?

zimeon:
It seems we end up with 404 either way around the changes are made since we can't sync the changes. I'll move the docs as soon as new redirects are in place

zimeon:
@dgarijo - OK, have merged the move so current w3id links are now 404 and new ones will be 200

PR #1035, opened 2018-06-25
arm: move activity ontology

zimeon:
Apologies for a second change in a few days but we want to rationalize the namespace for activity ontology as we wrap up the first phase of this project. Changes in targets _are_ already live, e.g. https://w3id.org/arm/activity/ontology/ will redirect (in browser) to  https://ld4p.github.io/arm/activity/ontology/0.1/activity.html

(re. https://github.com/LD4P/arm/issues/125#issuecomment-400024986)


zimeon:
Thanks!

PR #1041, opened 2018-07-05
Added CWLProv profiles

stain:
Adds version permalinks to https://github.com/common-workflow-language/cwlprov/


stain:
Note that this deliberately does not add `0.1` `0.3` etc. as we want only the full permalinks to be used.

PR #1053, opened 2018-07-21
Added Readme and redirect for Syntaxe du Codex Ontology

PietroLiuzzo:
https://w3id.org/sdc/ontology is the namespace for the ontology based on La Syntaxe du Codex for the syntactic description of manuscripts.

The Ontology is under development at https://github.com/BetaMasaheft/SyntaxeDuCodex.


dgarijo:
@PietroLiuzzo, 
Note that this will redirect to the ontology, but not serve the it properly with content negotiation

PietroLiuzzo:
Thank you very much!

PR #1056, opened 2018-07-26
Adding additional route for json-ld contexts

cmungall:


biolink/.htaccess
```diff
@@ -1,2 +1,3 @@
 RewriteEngine on
+RewriteRule ^/biocontext/$ https://raw.githubusercontent.com/prefixcommons/biocontext/master/registry/$1 [R=302,L]
```
davidlehn: This should probably be something like:
```
RewriteRule ^biocontext/(.*)$ https://raw.githubusercontent.com/prefixcommons/biocontext/master/registry/$1 [R=302,L]
```

PR #1058, opened 2018-07-29
Added GDPRov and GDPRtEXT projects

coolharsh55:


dgarijo:
@coolharsh55,
why don't you do content negotiation instead of creating a bung of URIs for the same resource? Just curious...

coolharsh55:
@dgarijo these are just replicated from the purl url set for now
Content negotiation does happen on the ontology url, but I couldn’t get it to work at the start using Nginx which is why I went with the silly route of so many urls

PR #1061, opened 2018-07-30
Adding a rule for the Blockcerts 2.1 JSON-LD schema.

lparkerlm:


blockcerts/.htaccess
```diff
@@ -5,4 +5,5 @@ RewriteEngine on
 RewriteRule ^$ https://www.blockcerts.org/ [R=302,L]
 RewriteRule ^v1$ https://www.blockcerts.org/schema/1.2/context.json [R=302,L]
 RewriteRule ^v2$ https://www.blockcerts.org/schema/2.0/context.json [R=302,L]
+RewriteRule ^v2.1$ https://www.blockcerts.org/schema/2.1/context.json [R=302,L]
```
davidlehn: "." is a regex metacharacter so should be escaped like `^v2\.1$`

lparkerlm: Ah, thanks for that.

PR #1069, opened 2018-08-15
Redirect for International Data Spaces project

cmader:


davidlehn:
This was merged rather quickly, but I do have some comments.

The redirects are going to the github UI pages vs to the raw resources.  Not sure if that is what was intended?  Did you mean to have wildcard redirects to a URL with the actual hosted resources?

The context link in the README isn't setup to work.

And a w3id project level consideration is that I wonder if something like "ids" should be reserved for w3id.org itself?  I imagine "id" should be reserved but it's unclear if "ids" could cause any confusion.  I understand it matches the related organization, but "ids" also sounds related to the w3id site itself that is about ids.  Anyone else have thoughts on this?

cmader:
Thanks for the comments! We'll discuss internally about how address these points and create a new pull request soon.

PR #1101, opened 2018-10-02
Add identifier for CRediT taxonomy and ontology

marijane:

PR #1103, opened 2018-10-03
Fix taxonomy and ontology redirects

marijane:
Per suggestion from @davidlehn at https://github.com/perma-id/w3id.org/pull/1101#issuecomment-426775000.

### DIFF
--- a/credit/.htaccess
+++ b/credit/.htaccess
@@ -2,5 +2,5 @@ Options +FollowSymLinks
 
 RewriteEngine on
 RewriteRule ^$ https://casrai.org/credit/ [R=302,L]
-RewriteRule ^taxonomy/(.*)$ https://dictionary.casrai.org/Contributor_Roles/$1 [R=302,L]
-RewriteRule ^ontology#(.*)$ https://raw.githubusercontent.com/data2health/credit-ontology/master/credit.owl [R=302, L]
+RewriteRule ^taxonomy(/.*)$ https://dictionary.casrai.org/Contributor_Roles$1 [R=302,L]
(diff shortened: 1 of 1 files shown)


marijane:
@davidlehn Apologies for the additional request for help, but all three of my redirects are still throwing 500 errors, and I have no idea how to fix them.

credit/.htaccess
```diff
@@ -2,5 +2,5 @@ Options +FollowSymLinks
 
 RewriteEngine on
 RewriteRule ^$ https://casrai.org/credit/ [R=302,L]
-RewriteRule ^taxonomy/(.*)$ https://dictionary.casrai.org/Contributor_Roles/$1 [R=302,L]
-RewriteRule ^ontology#(.*)$ https://raw.githubusercontent.com/data2health/credit-ontology/master/credit.owl [R=302, L]
+RewriteRule ^taxonomy(/.*)$ https://dictionary.casrai.org/Contributor_Roles$1 [R=302,L]
+RewriteRule ^ontology$ https://raw.githubusercontent.com/data2health/credit-ontology/master/credit.owl [R=302, L]
```
davidlehn: Remove the space before the L flag.

davidlehn: I just did the update.

davidlehn:
Try again, the flags had a syntax error.

davidlehn:
I fixed the taxonomy wildcard too.  Needs to be optional.  I think these rules probably do what you want now:
https://github.com/perma-id/w3id.org/blob/master/credit/.htaccess

marijane:
Thank you so much!  It looks like they are all working now.

PR #1109, opened 2018-10-16
Rename htaccess.txt to .htaccess

jmcmurry:


jmcmurry:
Thanks @dgarijo however, something is still not right. I get a server error when trying to resolve both https://w3id.org/loinc2hpo/annotations and https://w3id.org/loinc2hpo. I'm not quite sure what I've done wrong?

davidlehn:
It's a typo.  Remove the space in the flags.  Change `[R=302, L]` to `[R=302,L]`.

PR #1114, opened 2018-10-18
Update .htaccess

jmcmurry:
removing space per instruction from @dgarijo here https://github.com/perma-id/w3id.org/pull/1109#issuecomment-430344124; however neither of the redirects works (neither annotations nor base route) and the base route rule has no space in ````[R=302,L]```` so not sure that this will fix it


davidlehn:
A single syntax error will cause everything in the file to fail.  Looks like it works now.

PR #1124, opened 2018-10-24
Perma ID

fsteinemann:
Hello, I would like to create a permanent ID


davidlehn:
Hi, I'm guessing you are the one that emailed the mailing list the other day.  Can you make some updates to this pull request:
- Please remove the space from the directory name.  It would work, but it's not normal and I'm guessing a mistake.
- Please add a README.md file to that directory with at least some contact info (like email or github id).
- Did you change your mind about the `digital-reference` name from the email?  Looks like you added and removed it.
- The `.htaccess` file you have doesn't really need the `AddType` rules if it's just redirecting everything to one URL.
- It would be nice if you could squash all these commits into just one.

dgarijo:
@fsteinemann,
 I think the space in the directory is still there

fsteinemann:
Hello, I deleted the directory with the space in its name and created a new directory
Thank you

davidlehn:
- This is only going to redirect the one top level URL to a login protected page.  Is that what you want?
- These commits really need to be squashed.  There are 18 when it would be better as 1.  You can either do that yourself and force push to your branch or we can do it while merging, but that will make the upstream branch diverge from your local one.

fsteinemann:
Yes, that is what I want.
Can you please squash the commits while merging?

Thank you

Issue #1063, opened 2018-07-31
Unable to find valid certification path


megankatsumi:
I am using w3id.org a means of creating persistent, de-referenceable namespaces for a set of ontologies. Recently, I have encountered the following error when attempting to load the ontologies in Protege via the w3id url:
sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target

This is the first time I've encountered such an issue, (and have not recently updated Protege); the ontologies have been loading successfully via the w3id url prior to this. They also still load successfully via their current url on GitHub.

Any suggestions to fix this?

davidlehn:
I don't know what that error means.  Our prior cert was expiring recently and we switched w3id.org over to letsencrypt.  We haven't had other reports of problems though.  Which url is it?  Do you have a more detailed error?  Is w3id.org cert failing or did that pass and the target it redirected to is having problems?

megankatsumi:
Hi David,
The error seems to be occurring with all of the urls that I am using w3id.org for (within the /icity directory). It looks to me like the issue is with the wi3d.org cert, as the redirect targets open without issue when I use those urls instead. The stack trace that is provided by Protege is appended below.

sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target


Full Stack Trace
-----------------------------------------------------------------------------------------

org.semanticweb.owlapi.io.OWLOntologyCreationIOException: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
	at uk.ac.manchester.cs.owl.owlapi.OWLOntologyFactoryImpl.loadOWLOntology(OWLOntologyFactoryImpl.java:207)
	at uk.ac.manchester.cs.owl.owlapi.OWLOntologyManagerImpl.actualParse(OWLOntologyManagerImpl.java:1099)
	at uk.ac.manchester.cs.owl.owlapi.OWLOntologyManagerImpl.loadOntology(OWLOntologyManagerImpl.java:1055)
	at uk.ac.manchester.cs.owl.owlapi.OWLOntologyManagerImpl.loadOntologyFromOntologyDocument(OWLOntologyManagerImpl.java:1011)
[...]

PR #1126, opened 2018-10-25
update 

shekarpour:
Dear Admin

Could you please update

### DIFF
--- a/cevo/.htaccess
+++ b/cevo/.htaccess
@@ -16,9 +16,10 @@ RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
 RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
 RewriteRule ^$ https://shekarpour.github.io/cevo.io/ 
 
-# Rewrite rule to serve RDF/XML content if requested
+# Rewrite rule to serve RDF/XML OWL content if requested
 RewriteCond %{HTTP_ACCEPT} application/rdf\+xml [OR]
 RewriteCond %{HTTP_ACCEPT} application/owl\+xml
+RewriteCond %{HTTP_ACCEPT} ^.owl
 RewriteRule ^$  https://raw.githubusercontent.com/shekarpour/cevo.io/master/cevo.owl
 
 
--- a/empathi/.htaccess
+++ b/empathi/.htaccess
@@ -19,4 +19,5 @@ RewriteRule ^$ https://shekarpour.github.io/empathi.io/
 # Rewrite rule to serve RDF/XML content if requested
 RewriteCond %{HTTP_ACCEPT} application/rdf\+xml [OR]
 RewriteCond %{HTTP_ACCEPT} application/owl\+xml
+RewriteCond %{HTTP_ACCEPT} ^.owl
 RewriteRule ^$  https://raw.githubusercontent.com/shekarpour/empathi.io/master/empathi.owl


shekarpour:
Dear Admin

Could you please update

dgarijo:
@shekarpour,
I think there are still errors in your .htaccess. For example, you haven't added an [OR] in the line before last on your condition. In addiiton, I am not sure what you intend with the rewrite rule.
You are redirecting to your HTML first thing, so the second rule will never be triggered...

I recommend testing everything in a local server in your laptop before submitting your PR. Or have a look at some examples for redirection to vocabularies from the repository. Thanks!

cevo/.htaccess
```diff
@@ -16,9 +16,10 @@ RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
 RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
 RewriteRule ^$ https://shekarpour.github.io/cevo.io/ 
 
-# Rewrite rule to serve RDF/XML content if requested
+# Rewrite rule to serve RDF/XML OWL content if requested
 RewriteCond %{HTTP_ACCEPT} application/rdf\+xml [OR]
 RewriteCond %{HTTP_ACCEPT} application/owl\+xml
+RewriteCond %{HTTP_ACCEPT} ^.owl
```
davidlehn: I don't think this will work.  `Accept` can't be both `application/owl+xml` and start with `.owl`.  We can help with the rules if you'd like to explain what you are trying to do.

shekarpour:
Dear David

The point is that we can not open https://w3id.org/cevo or https://w3id.org/empathi in protege
meaning the redirect to owl file does not work,

Initially, I thought the problem is from the URI, then I changed to:
https://raw.githubusercontent.com/shekarpour/cevo.io/master/cevo.owl

still does not work, so probably the problem is from the rule

I appreciate if you can guide us

Thank you

davidlehn:
This may be a Protege issue related to https://github.com/perma-id/w3id.org/issues/1063.  It doesn't like the new Let's Encrypt certificate for some reason.  Someone with Protege and knowledge of Java certificates may need to debug this problem.

shekarpour:
David,  I am not sure it is protege 's fault, 
because I can successfully open the following link in protege:
https://raw.githubusercontent.com/shekarpour/cevo.io/master/cevo.owl

Thus, it can indicate that redirect rule does not perform well, do you have any solution as how can we fix that?

Thanks

davidlehn:
The problem is probably that Protege and/or whatever Java runtime it's using are having trouble with the Let's Encrypt cert that w3id.org is using.  So it fails to even get a response from w3id.org.  Tests I ran with just basic java on debian worked fine.  Do you have a reproduceable procedure that someone could use to see this problem?  What versions of protege, java, OS, etc are showing this issue?

dgarijo:
@shekarpour,
as David indicates, the content negotiation is well done. If you execute:
curl -sH "accept:application/rdf+xml" https://w3id.org/cevo"

You obtain:

<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>301 Moved Permanently</title>
</head><body>
<h1>Moved Permanently</h1>
<p>The document has moved <a href="https://w3id.org/cevo/">here</a>.</p>
<hr>
<address>Apache/2.4.7 (Ubuntu) Server at w3id.org Port 443</address>
</body></html> 

Which is the desired answer. It's a problem of certificates and Java.

davidlehn:
What's the status of this?

davidlehn:
No response. Closing. Reopen if needed.

PR #1133, opened 2018-11-01
Pull request for redirect to new vocabulary

lodewijk81:
Hi, 

Thanks for reviewing my redirect file.

Best wishes,
Lodewijk

### DIFF
--- a/pnv/.htaccess
+++ b/pnv/.htaccess
@@ -17,7 +17,7 @@ RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
 RewriteCond %{HTTP_ACCEPT} text/\* [OR]
 RewriteCond %{HTTP_ACCEPT} \*/\* [OR]
 RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
-RewriteRule ^pnv$ http://www.lodewijkpetram.nl/vocab/pnv/  [R=303,NE,L]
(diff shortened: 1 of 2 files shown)


pnv/.htaccess
```diff
@@ -0,0 +1,26 @@
...
+RewriteEngine on
+
+#Rewrite rules for pnv vocabulary 
+RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+xml|text/\*|\*/\*)
+RewriteCond %{HTTP_ACCEPT} text/html [OR]
+RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
+RewriteCond %{HTTP_ACCEPT} text/\* [OR]
+RewriteCond %{HTTP_ACCEPT} \*/\* [OR]
+RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
+RewriteRule ^pnv$ http://www.lodewijkpetram.nl/vocab/pnv/pnv.ttl  [R=303,NE,L]
```
davidlehn: These rules are relative to the current directory so would result in `https://w3id.org/pnv/pnv`.  You probably mean to use `^$`.

lodewijk81: Yes, you're right. I updated the .htaccess.

Sorry for the extra work,
Lodewijk

PR #1156, opened 2018-11-19
zora actions ontology

Fspiga13:


zoraActions/.htaccess
```diff
@@ -0,0 +1,5 @@
+Header set Access-Control-Allow-Origin *
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^$ https://github.com/Fspiga13/zoraActionsOntology [R=302,L]
+RewriteRule ^(.+)$ https://github.com/Fspiga13/zoraActionsOntology/$1 [R=302,L]
```
davidlehn: Did you mean to point to the raw content?  The above will go to the github UI for a resource.

Fspiga13: Yes, I think it's better to point to the raw content.

dgarijo: Then please update your PR and we'll merge it :)

PR #1165, opened 2018-11-25
added SSL certificates in .htaccess

ilin-t:


dgarijo:
hello @ilin-t,
I do not understand very well what you are trying to achieve with this change. w3id is https, but your URL is http. What do you gain by forcing everyone to be http? Isn't that the behavior by default?

ilin-t:
Hello Daniel,

I have the issue that is addressed in Issue #1063. With this change i tried
to tackle the issue, and add the required certificates so i can open the
resources via URL in Protege.

dgarijo:
I will merge the request to see if it works, but I don't think it will, as you still go from https to http. I have been wanting to research that issue for a while but got no time. Please let me know if this resolves it.

ilin-t:
Unfortunately, it didn't.

dgarijo:
Actually, this change broke it even more, because all URLs come from https and does not redirect. I am going to revert it

ilin-t:
Yeah, i saw, i made a new request with the old rules.

dgarijo:
Ok, now it doesn't load in protege, but redirects fine. This issue is very annoying. I addressed it in the past by doing my own content negotiation (if You do a CURL command, you can download the ontology and open it in Protege)  curl -sH  "accept:application/rdf+xml"  https://w3id.org/ontopbm-aquatic -L

ilin-t:
Yes, that is one way.  You can also open the ontology by entering the url
after the redirect in protege.

PR #1168, opened 2018-11-26
Update .htaccess

fpc-rescuer:


fpc-rescuer:
We would like to update the permanent links in htaccess file.

PR #1170, opened 2018-11-27
changed to github.io

muddymudskipper:


muddymudskipper:
changed /music-circle location to github.io

PR #1174, opened 2018-11-29
update htaccess

ncarboni:
updated and corrected version of htaccess


davidlehn:
It's kind of odd checking if user agent accepts html then returning pdf.

PR #1178, opened 2018-12-01
RC (Result-Context) ODP

iesnaola:


dgarijo:
@iesnaola,
why not use the URLs you have in your github pages instead of raw.githubusercontent ones?
For instance https://iesnaola.github.io/RC/ontology/ontology.ttl instead of https://raw.githubusercontent.com/iesnaola/RC/master/ontology/ontology.ttl

iesnaola:
That would certainly simplify the URLs. Thanks for the suggestion

PR #1193, opened 2018-12-12
Added Podchaser redirect (pc namespace)

bslinger:
To be used as an XML namespace for the additions to the RSS podcast spec used by podchaser.com


bslinger:
Thankyou!

PR #1196, opened 2018-12-17
music-circle: changed .n3 to .nt

muddymudskipper:


muddymudskipper:
comment should be: 
"music-circle: changed .n3 _to_ .nt"

PR #1204, opened 2018-12-21
The redirection for yxl works and korkut ontology has been created

bosforox:
Please pull my changes to add my ontology uri on your repo


dgarijo:
Glad it helped!

PR #1205, opened 2018-12-21
updated htaccess

ncarboni:
completely anew. should fix everything and be more readable.
Thanks for spotting the error on the last one!

### DIFF
--- a/sari/.htaccess
+++ b/sari/.htaccess
@@ -9,15 +9,15 @@ RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+
 RewriteCond %{HTTP_ACCEPT} text/html [OR]
 RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
 RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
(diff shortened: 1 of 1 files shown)


sari/.htaccess
```diff
@@ -1,23 +1,20 @@
...
+# In case of accept header <text/turtle>
+RewriteCond %{HTTP_ACCEPT} ^.*text/turtle.* 
+RewriteRule ^$ https://schema.swissartresearch.net/ontology/sari.ttl [R=303,NE,L]
 
-# Rewrite rule to serve RDF/XML content if requested
-RewriteCond %{HTTP_ACCEPT} application/rdf\+xml [OR]
-RewriteCond %{HTTP_ACCEPT} application/owl\+xml
-RewriteRule ^$ https://swissartresearch.net/ontology/sari.rdfs.xml [R=303,L]
+# If suffix ttl, redirect to turtle version
+RewriteRule ^sari.ttl$ https://schema.swissartresearch.net/ontology/sari.ttl [R=302,NE,L]
```
dgarijo: Here are you expecting https://w3id.org/sari/sari.ttl?
Or https://w3id.org/sari.ttl ? Because you are specifying the former

PR #1215, opened 2019-01-07
update htaccess for ttl

ncarboni:
should be fine now

### DIFF
--- a/sari/.htaccess
+++ b/sari/.htaccess
@@ -11,10 +11,10 @@ RewriteCond %{HTTP_ACCEPT} ^.*text/turtle.*
 RewriteRule ^$ https://schema.swissartresearch.net/ontology/sari.ttl [R=303,NE,L]
 
 # If suffix ttl, redirect to turtle version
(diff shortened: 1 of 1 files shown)


sari/.htaccess
```diff
@@ -1,23 +1,20 @@
...
 
-# Rewrite rule to serve RDF/XML content if requested
-RewriteCond %{HTTP_ACCEPT} application/rdf\+xml [OR]
-RewriteCond %{HTTP_ACCEPT} application/owl\+xml
-RewriteRule ^$ https://swissartresearch.net/ontology/sari.rdfs.xml [R=303,L]
+# If suffix ttl, redirect to turtle version
+RewriteRule ^$ https://schema.swissartresearch.net/ontology/sari.ttl [R=302,NE,L]
 
+# If suffix html, redirect to html version
+RewriteRule ^$ https://schema.swissartresearch.net/ontology/index.html [R=302,NE,L]
```
dgarijo: You are redirecting to html and ttl with the same rule here, these are contradictory

ncarboni: right. forgot about the RewriteCond

PR #1219, opened 2019-01-09
added an htaccess redirect for the FAIR Evaluator URIs

markwilkinson:
Thank you for this service!

### DIFF
--- a/FAIR_Evaluator/.htaccess
+++ b/FAIR_Evaluator/.htaccess
@@ -1,0 +1,5 @@
+Header set Access-Control-Allow-Origin *
+Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^ http://linkeddata.systems:3000%{REQUEST_URI} [R=302,L]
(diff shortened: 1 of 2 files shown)


dgarijo:
Hi @markwilkinson,
could you please add who will be the maintainer for this w3id?
A simple github id or email is enough. It help us track who is responsible for a w3id.
Thanks!

markwilkinson:
Daniel, I notice that my HTTP POST data isn't passing through the redirect... (message body is empty when it arrives).  Can you think of a reason this would be true, or suggest a fix?

dgarijo:
I have not experienced this, but maybe https://softwareengineering.stackexchange.com/questions/99894/why-doesnt-http-have-post-redirect
and
https://devcentral.f5.com/questions/how-to-preserve-data-in-a-http-302-redirect-of-a-post
help.
Please let me know if the issue is solved, so we can help more people in the future.

dgarijo:
Also, I recommend testing locally with your own apache server (e.g., using XAMPP) before doing multiple tries here, it is quicker :)

markwilkinson:
What is the time-delay between pull merge and when the changes go "live"?

markwilkinson:
Good news!  307 response code fixed the problem.  The message body is re-POSTed to the remapped URL.

Thanks for your time!  I looked at dozens of pages and never saw that solution!

Cheers Daniel!

dgarijo:
Excellent!

PR #1221, opened 2019-01-10
Create .htaccess

LeonardoForero:


dgarijo:
Sorry, we don't allow reserving w3ids. Can you please add the link to the redirection?

LeonardoForero:
Excuse me, I'm a beginner!

RewriteRule ^test-a# https://github.com/LeonardoForero/Knowls-Ontologies/blob/master/test-a.owl [R=303,NE,L]

davidlehn:
- You should add that to the .htaccess file.  Look at other files for examples.
- That github resources doesn't exist.
- If it did exist, that redirect would go to the github UI for that file vs serving it directly.  Consider the raw resource URLs, github pages hosting, or other hosting.

LeonardoForero:
I've updated the ".htaccess" file. Please find the ajustements!

davidlehn:
Can you also please add a short README.md with basic info like a title, maybe a reference url, and a contact?

If you just need simple redirects you could simplify the .htaccess file by removing the AddType and RewriteCond parts.  That's more useful when redirecting to different resources based on request accept headers.

dgarijo:
@LeonardoForero,
I don't recommend adding your phone number, your github id or your email is enough.

dgarijo:
Also, you are doing a 303 redirection, typically used for semantic-web resources. But you only serve one resource. If so, why not a 302?

PR #1229, opened 2019-01-18
Update .htaccess

angelosalatino:
Updated with another link.


davidlehn:
You might consider just replacing all the rules in that file with one wildcard redirect.

PR #1233, opened 2019-01-23
IP change

firmao:
please update


davidlehn:
If your IP will change often you should consider changing to a hostname and manage the IP mapping with a DNS provider.

PR #1248, opened 2019-02-10
Update .htaccess

cgueret:
Added cgueret to http://cgueret.github.io/


davidlehn:
You could use https, and the entry was *almost* in order, but no big deal.

PR #1250, opened 2019-02-12
Changed domain of OMG

AnnaWagner:
Before making changes, I copied the current status of the htaccess page. Now only the domains should be different.

Thanks!


davidlehn:
Please resolve the conflicts.

PR #1252, opened 2019-02-13
Add rewrite rules for /fair

tkuhn:


tkuhn:
This should have been two pull requests, one for /fair and one for /np. I didn't expect them to get merged by GitHub. I hope that's not a problem.

PR #1256, opened 2019-02-20
Add version specific redirects.

IC-Standards:


dgarijo:
Could you please add a short description for all these w3ids you are creating in your existing w3id file?
Is the same maintainer associated with all of them?
Thanks in advance!

IC-Standards:
Yes all of the files are maintained by the group I chair. The descriptions are harder there is comment code in the htaccess file but the actual what is at the destination is mainly reserved for the people we choose to share with. All the new URLs are behind a password protected wall so regular people can't get there anyway.


Did that help?


________________________________

dgarijo:
I understand. However, I see redirections that are not yet completed, e.g., 
RewriteRule ^DED/?$   [NE,NC,R=302,L]
does not redirect to anything?
As far as I know, so far we haven't allowed reserving w3ids. @davidlehn, thoughts?

davidlehn:
I'm not even sure what it does without a target.  I'm guessing it doesn't just break the file?  If so, you get to keep both pieces. ;-)  It would be better to add rules that do something rather than placeholders, but if you need to do it in your own file, that's fine.  Reserving top level ids is another matter.

I'll also note that the "Suffix" comments don't match up with these new "file", "repo", etc rules.  No big deal I guess, just confusing.

IC-Standards:
Sigh the without a target ones are ones that I do need to fix but can't do that right away, assuming they don't break the file none of our users should have been given them yet. I noticed those after I did the pull request.  And the comments don't deal with the "new" file and repo have to update the XSL that generates the .htaccess to fix those. Hopefully next commit with have both those updates.


If neither of those break things it would be great if this could be merged since the working URLs are ones that are blocking some of the users. The pre file/repo goes to sharepoint and the sharepoint admin has not been updating the pages to have the new content. File and repo get users to the content while we wait for sharepoint to work.


________________________________

PR #1259, opened 2019-02-25
Add rewrite rules for maturity_indicator under /fair

tkuhn:


fair/.htaccess
```diff
@@ -4,7 +4,9 @@ RewriteEngine on
 
 RewriteRule ^$ https://www.go-fair.org/ [R=302,L]
 
-RewriteRule ^metrics/(.*)$ https://github.com/FAIRMetrics/Metrics/#$1 [R=302,NE,L]
+RewriteRule ^maturity_indicator/$ https://github.com/FAIRMetrics/Metrics [R=302,L]
```
davidlehn: Might want to use `^maturity_indicator/?$` if you want the trailing `/` to be optional.

tkuhn: Right, thanks for pointing that out! I might do that for the next version of the rewrite rules.

PR #1261, opened 2019-02-26
Update .htaccess

ncarboni:
changed from rdfs to ttl and updated the htaccess

### DIFF
--- a/vir/.htaccess
+++ b/vir/.htaccess
@@ -1,23 +1,25 @@
-## Turn off MultiViews
+Options +FollowSymLinks
+RewriteEngine on
+
+# Turn off MultiViews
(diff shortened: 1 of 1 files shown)


vir/.htaccess
```diff
@@ -1,23 +1,25 @@
...
+# Turn off MultiViews
 Options -MultiViews
 
-# Rewrite engine setup
-RewriteEngine On
-			
+AddType text/turtle .ttl
+
+# In case of accept header <text/turtle>
+RewriteCond %{HTTP_ACCEPT} ^.*text/turtle.*
```
dgarijo: Is this correct? Why not just text/turtle?

PR #1276, opened 2019-03-14
Re Ontology URI redirection service

mazimweal:
Hi
 my path was not redirecting the URI .... guess I had made a mistake. I have made changes and corrections in the path .... I hope you accept the changes. will be glad to hear from you.
Allan


davidlehn:
You should probably update the README and redirects to use https urls.  w3id.org will redirect all http requests to https anyway.  And github.io can use https.

PR #1287, opened 2019-03-21
Neural Network Ontology

TobiasWeller:


davidlehn:
That target URL doesn't load.

TobiasWeller:
Ok., we uploaded the target file and changed the htaccess. Should work now.

PR #1292, opened 2019-03-23
RealEstateCore: added content negotiation

hammar:


dgarijo:
I recommend using a github page instead of raw.githubusercontent to serve your ontology :)

hammar:
> I recommend using a github page instead of raw.githubusercontent to serve your ontology :)

We're trying to figure out a workflow that allows us to link directly using ontology version IRI:s into the corresponding tags in GitHub, without having to manually move/copy anything. That way we can just draft a new release in GitHub, assign a version tag, and everything works and resolves automatically. Haven't figured out how to do that, if it is at all possible, using Pages..?

dgarijo:
Well, all contents of the master branch are zipped whenever you do a release in Github, so I don't see very well how each version would be served. Since I like to have all my versions available, what I do is that every time I "release" a stable version of the ontology, I create a folder with the appropriate documentation. This goes against the spirit of releases in Github, but does the trick. See an example: https://github.com/mintproject/Mint-ModelCatalog-Ontology/tree/master/modelCatalog/release and its corresponding https://w3id.org/mint/modelCatalog/
I generate the contents of each folder automatically with WIDOCO (plus a little of editing), so it's not that painful. But if you manage to link it to releases in GitHub, please let me know. I was interested in implementing something with OnToology (http://ontoology.linkeddata.es/), but never got the time to fully develop the idea.

PR #1304, opened 2019-04-02
Creation of URI 

fsteinemann:
Hi,

I want to create the URI `www.w3id.org/ecsel-dr-sn-SSP`.

Could you please squash the two commits into one?

Thank you very much


dgarijo:
The ontology does not seem to be accessible because it's behind a login page. Are you aware of this issue?

fsteinemann:
Yes, I am.

PR #1307, opened 2019-04-03
Redirects for the Timebank Ontology

tobiasgrubenmann:
The Timebank Ontology is an ontology for p2p service exchange.


davidlehn:
Please rename `.htaccess.txt` files to `.htaccess`.

tobiasgrubenmann:
Thanks for spotting this. I will fix it and create a new pull request

PR #1316, opened 2019-04-05
Add won folder for webofneeds project

fkleedorfer:


dgarijo:
@fkleedorfer,
if you are not adding content negotiation, why not use just 301/302 redirect?

fkleedorfer:
Yep you're right, I'll change it to 301. I was thinking about content negotiation, but for the time being, we won't do it.

davidlehn:
You are welcome to use 301, but that might be cached such that a user agent wouldn't see a change on w3id.org.  Is that ok?

fkleedorfer:
Well it sounded like a good idea at the time... but if we want to change the redirect later (e.g., add content negotiation), which we might, 301 may hide that. Not what we want, thanks for pointing that out. So I'll go for 302 then.

PR #1319, opened 2019-04-08
change the default response

saidfathalla:
change the default response to ttl file and owl file location to our server


saidfathalla:
I don't know what is making the conflict? could you please check?

davidlehn:
I squashed all the commits when doing that last merge.  Looks like they are here again.  Not sure how that happened.  Probably these new commits are based off of your branch vs the upstream with the squashed commit?  Yet I see that commit.  I dunno.  Can you hit the "Resolve conflicts" button, and fixup the code to what you want it to be?  This PR will need to be squashed again as well.  Next time you should base changes off the latest upstream master branch.

PR #1326, opened 2019-04-10
Adding permanent identifier for AIRM-O project

xtoph85:
We're publishing a developed ontology and would like to assign permanent identifiers to this project.


dgarijo:
You are not adding redirections to any ontology yet, but I will accept the PR

PR #1330, opened 2019-04-12
Need to update Permanent Identifier

khancynthia28:
We submitted a request earlier for a permanent identifier which got approved as https://w3id.org/AEOLUSsmartAPI. However, the Conference Paper which we submitted to ISWC 2019, describing this resource, contains the URI as https://w3id.org/AEOLUSsmartApi. Therefore we are requesting to change our redirect URI to a lower case 'Api' instead of 'API'. The correct Permanent Identifier Link should be https://w3id.org/AEOLUSsmartApi. Please accept our change. Thanks.


davidlehn:
While some file systems can handle this, others get upset trying to checkout git repo directories with names that are similar other than case.  Redirect rules would also be duplicated and more likely to become unsynced.  We can add a special top level .htaccess rule for the "Api" version to redirect to the "API" dir (or the other way around).  I'd be nice to avoid, but we already had to do that for a "un" prefix.  Would that work?  Which naming would you prefer?

davidlehn:
If you intended to just remove the other version and stick with the new name, that works too.  If so, just add the file removals to this PR.

khancynthia28:
Deleting the incorrect Permanent Identifier

khancynthia28:
Hello,

I have added the file removals to the PR for the older name (AEOLUSsmartAPI). Please merge the PR so that the new name (AEOLUSsmartApi) can come into effect. The Conference paper is currently under review and the new URI being active is vital for the process. Thank you.

dgarijo:
I will merge the PR so it works for the author. @davidlehn please revert if you don't consider this appropriate

PR #1336, opened 2019-04-15
Update .htaccess

saidfathalla:
change files urls


saidfathalla:
@davidlehn could you please merge?

PR #1338, opened 2019-04-16
update htaccess

saidfathalla:
change redirect of documentation.


saidfathalla:
@davidlehn when doing content negotiation, it returns 301 then from the 2nd request it returns 303. Do you know how to solve this? N.B. it is working with the url https://w3id.org/seo/ not https://w3id.org/seo#

saidfathalla:
@davidlehn  I found the problem is seo is in the root directory of w3id, but other ontologies, such as coil ontology, have their htaccess file in a sub directory. do you have a solution to this problem? i cannot change the right now.

davidlehn:
If I understand your issue, then yes, it's going to do redirects like that.  It's an artifact of this service using apache.  What problem does that cause?  If you know of an apache configuration that would help with this, we can discuss that.

saidfathalla:
@davidlehn could you please merge this changes first? the problem is when doing the content negotiation with vapour, it fails the test because it returns and empty response with 301. I tried other ontologies which have URL https://w3id.org/XXX, e.g. BCI-ontology

davidlehn:
Can't merge this yet.  It has changes to the 'un' dir (again).  Fix this PR so it only has the changes you want.  Check here:
https://github.com/perma-id/w3id.org/pull/1338/files
Due to the 45(!) commits, this will be squashed when merging unless you fix that yourself first.

PR #1347, opened 2019-05-01
Change ip address of target server

matt-wallis:


dgarijo:
@matt-wallis,
any possibility that you store files on a more permanent URL?  51.15.69.110 doesn't look very permanent. For example, since most are files, you could have them in a github page

matt-wallis:
@dgarijo ,
I'm not sure I understand what you are asking. Are you asking me to change something?

dgarijo:
@matt-wallis I am suggesting you to use URLs that are more permanent than an IP such as 51.15.69.110. Otherwise we will have to be changing the w3id every time the IP changes.

matt-wallis:
I have no expectation that the ip address will change - I am paying for a fixed IP address. The change in this pull request was due to me moving to a new server. I forsee no reason to change it again.

PR #1352, opened 2019-05-04
permanent identifier for Costume Core documentation and ontology

ArdenKirkland:


dgarijo:
@ArdenKirkland,
I think there is an error in one of the redirections. You are creating a w3id to https://w3id.org/costumeCore/ontology/ontology
instead of 
https://w3id.org/costumeCore/ontology/

You should change the RewriteRule ^ontology$  to be RewriteRule ^$

ArdenKirkland:
Thanks so much, @dgarijo! Good catch. Thanks also for your series of blog posts which have helped me to figure out (mostly) how to do this!

dgarijo:
Happy to help!

PR #1355, opened 2019-05-08
Rule fixing

anuzzolese:
This pull request fixes some issues associated with the resolution of
URIs for alignment ontologies and the following controlled vocabularies:
- classifications-for-public-services/life-business-event/life-event
- classifications-for-public-services/life-business-event/business-event
- territorial-classifications/countries/italy


davidlehn:
Haven't seen anyone here use env vars to pick extensions like that before.  Interesting technique.

PR #1364, opened 2019-05-14
Patch 14

luigi-asprino:


ecodigit/.htaccess
```diff
@@ -1,5 +1,20 @@
...
 RewriteEngine on
 RewriteRule ^$ http://ecodigit.dtclazio.it [R=302,L]
 RewriteRule ^(.+)$ http://ecodigit.dtclazio.it/$1 [R=302,L]
+
+# Rewrite rule to serve HTML content
+RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+xml)
+RewriteCond %{HTTP_ACCEPT} text/html [OR]
+RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
+RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
+RewriteRule ^ontology(.*)$ http://wit.istc.cnr.it/arco/lode/extract?url=https://raw.githubusercontent.com/ecodigit/workspace/master/ontology/$1/latest/$1.owl [R=303,L]
```
davidlehn: Rules are checked in order so the `^(.+)$` above will match before these `ontology` ones.  You probably should move the catch all wildcard rule to the bottom.

luigi-asprino: Thanks

PR #1380, opened 2019-05-22
need to keep paths only

markwilkinson:


FAIR_Training_LDP/.htaccess
```diff
@@ -1,4 +1,4 @@
 Options +FollowSymLinks
 RewriteEngine on
-RewriteRule ^ http://training.fairdata.solutions/DAV/home/LDP/fair%{REQUEST_URI}/ [R=307,L]
+RewriteRule ^FAIR_Training_LDP/(.*) http://training.fairdata.solutions/DAV/home/LDP/fair/$1/ [R=307,L]
```
davidlehn: Patterns are relative to current directory. You probably want this:
```suggestion
RewriteRule ^(.*) http://training.fairdata.solutions/DAV/home/LDP/fair/$1/ [R=307,L]
```
You might end up with double trailing `/` with the above.  Maybe want to remove the redir URL one.

markwilkinson: nope.  What I requested is what I want.

markwilkinson: I guess removing the trailing slash would be fine too, though the current behaviour doesn't break me.  For now, let's just leave it as it is.

davidlehn: Ok, that means urls will be like `https://w3id.org/FAIR_Training_LDP/FAIR_Training_LDP/...`

PR #1384, opened 2019-05-28
update .htaccess for rdf content negotiation

kabulkurniawan:
thanks :)

### DIFF
--- a/sepses/.htaccess
+++ b/sepses/.htaccess
@@ -8,8 +8,6 @@ AddType text/turtle .ttl
 AddType application/n-triples .n3
 AddType application/ld+json .json
 
-RewriteRule ^$ http://sepses.ifs.tuwien.ac.at [R=302,L]
-RewriteRule ^(.*)$ http://sepses.ifs.tuwien.ac.at/$1 [R=302,L]
 
 #Rewrite rules for sepses vocabulary 
 RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+xml|text/\*|\*/\*)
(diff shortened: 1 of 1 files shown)


sepses/.htaccess
```diff
@@ -1,4 +1,36 @@
...
 RewriteEngine on
+
+AddType application/rdf+xml .rdf
+AddType application/rdf+xml .owl
+AddType text/turtle .ttl
+AddType application/n-triples .n3
+AddType application/ld+json .json
+
 RewriteRule ^$ http://sepses.ifs.tuwien.ac.at [R=302,L]
 RewriteRule ^(.*)$ http://sepses.ifs.tuwien.ac.at/$1 [R=302,L]
```
davidlehn: Rules are checked in order, so this wildcard rule will catch everything.  You should move it to the bottom of the file.

kabulkurniawan: done.. thankyou

kabulkurniawan: Hi I still got a problem when I open this link  http://w3id.org/sepses/vocab/ref/cve/ontology.ttl via protege, the error message is :
sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target

but when I directly open this link: http://sepses.ifs.tuwien.ac.at/vocab/ref/cve/ontology.ttl, it works fine.

do you see that this is a problem from w3id?

dgarijo:
This is was Protege problem.
They have fixed it in the Protege version 5.5. I just opened
https://w3id.org/sepses/vocab/ref/cve/ontology.ttl fine.

AnnaWagner:
I just updated the htaccess of BOT while I was fixing my htaccess files, should work fine now. Just noticed during copying the rewrite rules that BOT is not provided in any serialisation but TTL, maybe this could be supplemented by RDF, JSON-LD and NT?

PR #1385, opened 2019-05-29
Added new link

xtoph85:


dgarijo:
Note that there is no RDF content negotiation, so this ontology will not work in an ontology editor

PR #1393, opened 2019-06-06
Update .htaccess

mariapoveda:
undo


dgarijo:
In the future, please do both changes in the same PR :)

PR #1400, opened 2019-06-13
SARO Ontology permanent namespace

scerrisimon:
The namespace of the ontology (available at the rewrite rule URI) will be changed to:

w3id.org/saro#

### DIFF
--- a/saro/.htaccess
+++ b/saro/.htaccess
@@ -4,10 +4,10 @@ AddType text/turtle .ttl
 Header set Access-Control-Allow-Origin *
 Options +FollowSymLinks
 RewriteEngine on
-RewriteRule ^/?$ https://elisasibarani.github.io/SARO/ [R=302,L]
+RewriteRule ^/?$ https://elisasibarani.github.io/SARO/ [R=303,L]
 
 RewriteCond %{HTTP_ACCEPT} ^.*application/rdf\+xml.* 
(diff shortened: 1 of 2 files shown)


dgarijo:
@scerrisimon, you are not redirecting to the right file if you plan to serve ttl/html. I suggest you use a github page for this.

dgarijo:
@scerrisimon,
same thing applies for the rdfs/turtle files.
Also, shouldn't the html redirect be 303?

scerrisimon:
Fixed issue with 302-303, changed rdfs to rdf file extension. Set up html as a page and also ttl and rdf files are directly accessible. I hope this is now according to standard!

scerrisimon:
Updated Readme file with contact info. Awaiting further instructions..

dgarijo:
In line 13 you have "© 2019 GitHub, Inc.", which is not valid syntax.
Also, in line 7 why do you have RewriteRule ^/?$ https://elisasibarani.github.io/SARO/ [R=303,L]
instead of just RewriteRule ^$ https://elisasibarani.github.io/SARO/ [R=303,L]
?

scerrisimon:
Thnks for your patience. I was using a 'template' (or so I thought) from another w3id example. Removed the issues and committed!

davidlehn:
@scerrisimon Rules are processed in order.  The default one listed first will be used before the type detection ones are checked.  You should move that rule to the bottom.

PR #1407, opened 2019-06-19
additional redirect because people are forgetting the case-sensitivit…

markwilkinson:
additional redirect because people are forgetting the case-sensitivity of the original AmIFAIR

### DIFF
--- a/certainty_nanopubs/README.md
+++ b/certainty_nanopubs/README.md
@@ -1,0 +1,7 @@
+## Redirect for the Certainty Nanopublications
+
+From the thesis of Mario Prieto come these exemplar Nanopuiblications
+including an annotation of the degree of certainty being expressed
+in each assertion.
+
+Contact:  mark.wilkinson@upm.es with any concerns.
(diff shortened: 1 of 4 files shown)


davidlehn:
We need to avoid making files or directories that only differ in case.  Some systems have problems with that and can't even check out the code if such things exist.  If it's really a problem then the solution is to put a fix in the top-level .htaccess.  We did that for one other project, but would prefer to avoid that when possible.  What would you like to do?

markwilkinson:
I will back-out the request.  
Thanks!

certainty_nanopubs/README.md
```diff
@@ -0,0 +1,7 @@
+## Redirect for the Certainty Nanopublications
+
+From the thesis of Mario Prieto come these exemplar Nanopuiblications
```
TallTed: typo --
`Nanopuiblications` -> `Nanopublications`

davidlehn:
This is confusing. :-)  There's the lowercase patch, a unrelated nanopub patch with unfixed spelling issue, and a revert of the first commit.  In the future please use a fresh branch and PR for new changes.  Thanks.

PR #1412, opened 2019-06-24
Add swot folder for redirection

fr4ncidir:
Please add this redirection to swot ontology. Thanks.

### DIFF
--- a/swot/README.md
+++ b/swot/README.md
@@ -1,4 +1,4 @@
-# IOMUST: internet of musical things ontology
+# SWOT: Semantic Web of Things Ontology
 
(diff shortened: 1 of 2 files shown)


swot/README.md
```diff
@@ -0,0 +1,5 @@
+# IOMUST: internet of musical things ontology
```
davidlehn: Looks like this be about SemanticWoT.  I'm assuming this was a copy & paste from previous PR.

fr4ncidir: It's totally a copy paste error. Sorry about that, I'll fix the problem.
Thank you a lot.

PR #1413, opened 2019-06-25
Add IDS folder with htaccess

vemonet:
For the Institute of Data Science at Maastricht University http://maastrichtuniversity.nl/IDS


davidlehn:
We had a request for "ids" before, and I had a concern then that I still have.  The name "ids" on a service dealing with ids could lead to confusion.  We probably should explicitly reserve names like that.  Thoughts?  Is there another name that would work for you?
https://github.com/perma-id/w3id.org/pull/1069
https://github.com/perma-id/w3id.org/pull/1070

vemonet:
No problem, I was actually surprised that those 3 letters (ids) were not already taken (especially with identifiers services)
We will look into a another URI then, thanks

PR #1417, opened 2019-06-27
Add permanent identifier for MAKG

michaelfaerber:


dgarijo:
@michaelfaerber,
http://ma-graph.org/ontology leads to a 404. I think the intended redirect is http://ma-graph.org/ontology.owl

PR #1423, opened 2019-07-01
Update .htaccess

HadleyKing:
Versioning of links


dgarijo:
@HadleyKing,
the options are served in order. If you add
RewriteRule ^(\d+.\d+.\d+)
before   ^1.2.0$, then it will never get to  ^1.2.0$

PR #1426, opened 2019-07-03
Added CRI association namespace

NSeydoux:
CRI is an independant research association, and part of its activity will resort to the publication of vocabularies, hence the need of a permanent namespace.

### DIFF
--- a/cri/.htaccess
+++ b/cri/.htaccess
@@ -1,0 +1,8 @@
+Header set Access-Control-Allow-Origin *
+Options +FollowSymLinks
+RewriteEngine on
+
+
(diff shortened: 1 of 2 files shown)


dgarijo:
@NSeydoux,
you are creating w3ids like https://w3id.org/cri/cri 
I think you intend to do a redirection like this:
RewriteRule ^$ https://cri-association.fr/ [R=302,L]
RewriteRule ^ns/obema$ https://zwifi.eu/voc/obema/ [R=302,L]

NSeydoux:
Hi @dgarijo, 
Thank you for the feedback, and I'm very sorry I did not come back to make this fix earlier !

PR #1455, opened 2019-08-13
changed ysedira for riccardotommasini

riccardotommasini:


davidlehn:
@jpcik Is this change ok?  Looks like old urls are gone.

dgarijo:
@davidlehn, @riccardotommasini is a coauthor on the spec, so the change seems appropriate.

davidlehn:
Ok, I was just looking at the commits here.

PR #1460, opened 2019-08-25
Updated the htaccess with the new rewrite rules

essepuntato:


essepuntato:
Hi David,

I’ve tested them here:

https://htaccess.madewithlove.be/ <https://htaccess.madewithlove.be/>

And they seemed correct indeed – even if the online tester actually seems to have a small bug since it seems that the URL changed as a consequence of a rule is not used as input of the following one, something that should be guaranteed by the htaccess spec.

In practice, following the guidelines for using the LODE service available at https://essepuntato.it/lode <https://essepuntato.it/lode> (a website which is now handled via GitHub pages), the idea is that a possible user specifies an URL like

https://w3id.org/lode/owlapi/lang=it/imported/http://purl.org/spar/cito

And the URL must be translated as follows, so as to call the service in the server where it is currently available (that in principle may change, but using w3id PURL will guarantee the correct working):

http://150.146.207.114/lode/extract?owlapi=true&lang=it&imported=true&url=http://purl.org/spar/cito <http://150.146.207.114/lode/extract?owlapi=true&lang=it&imported=true&url=http://purl.org/spar/cito>

If you try to put the previous URL on a browser, the service will be called and it will return an HTML as intended.

I hope this clarify the rules.
Have a nice day :-)

S.




----------------------------------------------------------------------------
Silvio Peroni, Ph.D.
Assistant Professor, Department of Classical Philology and Italian Studies
University of Bologna, Bologna (Italy)
Director, OpenCitations
Figshare Ambassador
E-mail: silvio.peroni@unibo.it
Web: https://www.unibo.it/sitoweb/silvio.peroni/en
Twitter: @essepuntato

PR #1467, opened 2019-08-30
CYBELE model

rapw3k:
Adding space for CYBELE models. See: https://www.cybele-project.eu/


dgarijo:
@rapw3k, there is no redirection here. By default we don't allow reserving URIs if there is no redirection. I suggest adding a small htaccess redirecting https//w3id.org/cybele to https://www.cybele-project.eu/

rapw3k:
Thanks @dgarijo , I added a small htaccess

PR #1469, opened 2019-09-02
Add namespace for Schneider-Electric

jbulow:


davidlehn:
You are pointing at github interface urls.  Did you mean to point somewhere that the raw files will be served?  Also we prefer to add these rules when there are actual files being used rather that just people reserving names.

jbulow:
Hi,

We are in the process of standardising the process of our creation and use of ontologies. My proposal was to use permanent links instead of random, private and non-retrievable UR:s. So yes, in this phase it is just a name space reservation because that is the best I can do right now.

Is there a better way to do this given the circumstances?

jbulow:
I've updated the redirection to a github pages placeholder.

PR #1476, opened 2019-09-07
Add so-ontology

marioscrock:
The survey-ontology helps in publishing surveys and answers collected as linked data research objects.


dgarijo:
@marioscrock,
the redirections point to urls that are local paths, and won't work because they are not part of the PR (and we don't store files in this repo). Instead of files, add the URL to where those files are hosted, please. For example, instead of
RewriteRule ^$ index.html [R=303,L]
do something like:
RewriteRule ^$ YourURL/index.html [R=303,L]

marioscrock:
I set the base path on line 14 with RewriteBase in the .htaccess file. It should be fine or am I wrong? Thank you

dgarijo:
@marioscrock,
have you tried this file locally? Because I have and I get an error (500).  I think the rewriteBase is not intended to be used this way...

marioscrock:
I used rewriteBase in the wrong way, sorry. I changed the .htaccess file and tested it locally, it should be ok. Thank you again

PR #1478, opened 2019-09-09
new redirect for execution technique ontology

PietroLiuzzo:
This is a redirect for a w3id url we would like to use for our execution technique ontology, which is part of the efforts towards a general modelling of epigraphic documents.


PietroLiuzzo:
my plan is to then replace the namespaces and updated the url for all concepts in the ontology and in the vocabulary, using https://w3id.org/executionTechnique/ontology

PietroLiuzzo:
thank you!

PR #1480, opened 2019-09-10
Update .htaccess

luigi-asprino:


dgarijo:
@luigi-asprino,
you are aware you are doing 302 redirects, right? Doing 301 may be better if your redirections are permanent

PR #1489, opened 2019-09-17
added readme files and removed empty htaccess file

JoepvanGenuchten:
added readme files and removed empty htaccess file


dgarijo:
@JoepvanGenuchten,
you left a new 2.txt that I think should not be there

davidlehn:
One readme at the top level would be fine if they are all going to be the same.

JoepvanGenuchten:
I removed the access files and added a new folder  and htaccess file in this commit

PR #1490, opened 2019-09-18
Added amlo namepace

ecow:
We are in the process to write a new project publishing a public ontology, so we need a permanent namespace. Please add my new permanent id. Thank you.

### DIFF
--- a/amlo/.htaccess
+++ b/amlo/.htaccess
@@ -1,0 +1,14 @@
+Header set Access-Control-Allow-Origin *
+Options +FollowSymLinks
+RewriteEngine on
+SetEnvIf Accept ^.*application/rdf\+xml.* SYNTAX=rdf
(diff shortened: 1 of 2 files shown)


dgarijo:
@ecow,
this is the first time I see the redirects made this way. Have you tried this locally? 
Also, you seem to need login to access some of these resources. Not sure if the redirection will work appropriately. If you confirm you have tried it locally I will merge

ecow:
Hi @dgarijo ,
I confirm that the redirects work locally. I confirm also that the redirects, now, point to a protected repository. Our project is at an early stage and is not yet stable, but we need a flexible solution to work with real namespaces. This is more or less a placeholder that I will update as soon as the ontologies names and contents will be more stable. About the redirection: for .htaccess I used a flexible skeleton easy to maintain that you should already have seen in https://github.com/perma-id/w3id.org/blob/master/italia/onto/.htaccess where I got the idea.

PR #1514, opened 2019-10-16
redirects for GOM ontology

mathib:
Hi!

This redirect contains a proposal for the registration of https://w3id.org/gom for the GOM or Geometry Metadata Ontology. Additional references can be found in the readme.

Best regards,

Mathias Bonduel

### DIFF
--- a/gom/.htaccess
+++ b/gom/.htaccess
@@ -1,0 +1,40 @@
+Options +FollowSymLinks
+RewriteEngine on
+
+# Turn off MultiViews
+Options -MultiViews
+
+AddType text/turtle .ttl
+AddType application/rdf+xml .rdf
+
+# Rewrite rule to serve HTML content from the vocabulary URI if requested
+RewriteCond %{HTTP_ACCEPT} text/html [OR]
+RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
+RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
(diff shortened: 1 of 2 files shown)


dgarijo:
@mathib,
a small comment:  the first rule: RewriteRule ^$ https://mathib.github.io/gom-ontology/ [R=302,NE,L]
will void the last one (default). Why not make this one a 303?

mathib:
Hi, thanks for merging! I actually just copied the code from another working example and didn't notice :). I think the first rule was meant as a redirect if html is explicitly is asked and the last one reacts as a default if nothing specifically is asked?

dgarijo:
@mathib,
yes, you are correct. But lines 14 and 22 should be a 303 redirection instead of a 302 redirection, because you are doing content negotiation. See: https://www.w3.org/TR/swbp-vocab-pub/#recipe2

mathib:
OK, I'll have a look at it! Thanks for the pointer.

BTW, the w3id URI does not yet exist? If I request `https://w3id.org/gom` / `https://w3id.org/gom/gom.ttl` / `https://w3id.org/gom/gom.html` / ... I get a 404. Maybe I'm too impatient, but the main readme of the w3id repo states that "Once the pull request is merged, the changes go live immediately"

dgarijo:
This should be working immediately. @davidlehn, any hints of what the problem could be?

davidlehn:
Fixed. See https://github.com/perma-id/w3id.org/issues/1516.

mathib:
Great, thanks :)

PR #1519, opened 2019-10-18
Update .htaccess

GeorgFerdinandSchneider:
Revision and addition of rewriting rules for IBP namespace and adoption of Widoco documentation


davidlehn:
That file is getting rather huge and I imagine is hard to manage.  Certainly too large to review.  Looks like many of the rules are similar.  Maybe capture groups might help for common patterns?

May be able to collapse lots of those sections into just rules like this:
```
RewriteRule ^(.*)$ https://technicalbuildingsystems.github.io/Ontologies/$1/ontology.xml [R=303,L]
```

If that doesn't work in your use case, no big deal.

PR #1521, opened 2019-10-20
Smetzger

stefanmet:


dgarijo:
@stefanmet,
some of the changes override existing redirections to repositories that are not public, and I cannot verify them (e.g., https://github.com/NEONScience/eddy4R) Are you making them private, or is there a problem in the redirection?

stefanmet:
Thanks @dgarijo for checking!

It's a bit of a chicken and and egg challenge - we are preparing our online resources for conjoint publication with a manuscript in which these PURLs are used to provide the linkage.

The manuscript is being typeset, for which we have updated three redirects to their final targets (they pointed to temporary targets during the manuscript review period). One of the redirects (eddy4R GitHub repo) will be switched public only on the day when the manuscript is released. In the meantime, I am happy to check myself that this redirect works.

Please let me know if you have any concerns.


Thanks - Stefan.

dgarijo:
@stefanmet gotcha. I merged the pr

stefanmet:
I see that the wiki redirect now also points to the (still) private eddy4R repo, which is correct. So there should only be two redirects that you are not currently able to access. This at the same time as the manuscript comes out in November. Hope this helps - S.

stefanmet:
Thanks @dgarijo! S.

PR #1522, opened 2019-10-21
prefixed all links with "https://" to fix redirect errors; added documentation

stefanmet:
- updated redirect documentation


davidlehn:
That's a lot of files.  You could probably rearrange as one README and put all the redirect rules in one top level file rather than one per dir.   Might be easier to manage, but ok either way.

stefanmet:
Thanks @davidlehn for suggesting to consolidate! Do you have an example that we could build off? Previously I was under the impression that a separate folder is required for each redirect. I agree that a single file would be easier to manage. Thanks! S.

PR #1536, opened 2019-11-04
Added obelisk namespace

NSeydoux:
This namespace is used for a simple vocabulary describing obelisks. This vocabulary is used as an example for a tutorial on https://solidproject.org/, and its publication under a permanent IRI is consistent with the good practices identified in said tutorial.

### DIFF
--- a/obelisk/.htaccess
+++ b/obelisk/.htaccess
@@ -1,0 +1,6 @@
+Header set Access-Control-Allow-Origin *
+Options +FollowSymLinks
+RewriteEngine on
+
+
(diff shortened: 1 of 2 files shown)


dgarijo:
@NSeydoux,
https://zwifi.eu/voc/obelisk/ returns 404. Is this the correct link?

NSeydoux:
@dgarijo, my bad, when accessed though a browser the IRI redirects to an HTML documentation that I forgot to include. The link is correct, and I fixed that so now it should be okay.

PR #1537, opened 2019-11-04
Plan Executions (PlEx) Ontology

jonimoreira:
FAIR Workflows 
Profile reusing predicates from workflow ontologies.


dgarijo:
@jonimoreira,
we don't store content files here. Instead, you should add the redirection to the server/repository with the right file. Please find some examples in the rest of the folders. For example:  https://github.com/perma-id/w3id.org/blob/master/ac-ontology/.htaccess

jonimoreira:
Hi @dgarijo ,
Sorry for the mistake and thanks for the support. 
Can you check now if everything is ok, please? 
Thank you.

dgarijo:
@jonimoreira,
is the w3id you want  https://w3id.org/fair/plex/aco ?? (Maybe the "aco" is a mistake??)

jonimoreira:
It was a mistake, removed. Thnks

dgarijo:
@jonimoreira,
now you have set it up to be  https://w3id.org/fair/plex/plex

PR #1538, opened 2019-11-06
created fso folder and added 2 files

alikucukavci:


alikucukavci:
Fluid System Ontology (fso) folder is created together with htaccessfile and readme

dgarijo:
@alikucukavci,
https://w3c-lbd-cg.github.io/fso.ttl is returning a 404. Is the link correct?

alikucukavci:
This pull request should work. 
I have updated the .htaccess file. 
Sorry for that :)

PR #1543, opened 2019-11-11
adding ontology for OEB datasets

sthhher:
we want to connect json and ontologies with a permanent id. This is for the project obenebench.
https://openebench.bsc.es/html/dashboard

### DIFF
--- a/oebDatasets/.htaccess
+++ b/oebDatasets/.htaccess
@@ -1,0 +1,15 @@
+Options +FollowSymLinks
+
+Options -Indexes
+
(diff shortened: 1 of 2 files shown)


dgarijo:
@sthhher,
is your second redirect correct? You are creating the w3id: https://w3id.org/oebDatasets/(latest).
Also, why a 307 redirect instead of 302?

sthhher:
You are right. I have correct the mistakes and I have push it again. Thank you

PR #1545, opened 2019-11-13
adding oebDataFormats

sthhher:


sthhher:
oebDataFormats is like oebDatasets but linked to an other ontology for the same project. We have imported the ontology of DataFormats in the ontology of Datasets and we need a permanent id for both.

oebDataFormats/.htaccess
```diff
@@ -0,0 +1,15 @@
...
+
+Options -Indexes
+
+RewriteEngine on
+
+#Unversioned
+RewriteRule ^$ https://raw.githubusercontent.com/inab/OEB-ontologies/master/dataFormats.owl [R=302,L]
+
+#Latest version
+RewriteRule ^latest https://raw.githubusercontent.com/inab/OEB-ontologies/master/dataFormats.owl [R=302,L]
```
davidlehn: FYI, You might consider using using a trailing `$` in these patterns, like `^latest$`.  Otherwise a request path like `/oebDataFormats/latestfoobar` would also match.  But maybe that's ok.  Same below.

sthhher: Yes, you are right. I have already corrected it. Thank you!

PR #1547, opened 2019-11-13
Adding version 0.1.2

jonimoreira:
Hi @dgarijo ,
Thanks for all the support.
I wonder if this way is a good practice for PIDs of ontology versions or do you recommend another approach? 
Gracias!
Joao


dgarijo:
Your approach is valid but a little bit hard maintain if you plan to do this for each version. What I usually do is that I have the redirect to the latest version and then the link to any version. This is an example:  https://github.com/perma-id/w3id.org/blob/master/okn/o/sd/.htaccess
I cannot guarantee this is the best practice, but it's the one which worked best for me.

PR #1548, opened 2019-11-13
adding at the end of rewrites $ and change the name of github using oebDataFormats at the first of link

sthhher:


dgarijo:
I would recommend to use a guithub page instead of raw.githubusercontent.

sthhher:
We need raw.githubcontent instead of github page, thanks!

PR #1552, opened 2019-11-14
Redirects for RO-Crate JSON-LD context

stain:


stain:
Thanks, Dani!

dgarijo:
No problem! I am looking forward to do a pass to the spec and give some comments, even if they don't make it to this version..

PR #1563, opened 2019-11-22
update chainpoint for v4 spec

jacohend:


chainpoint/.htaccess
```diff
@@ -5,3 +5,4 @@ RewriteEngine on
 RewriteRule ^$ https://chainpoint.org/ [R=302,L]
 RewriteRule ^v2$ https://chainpoint.org/contexts/chainpoint-v2.jsonld [R=302,L]
 RewriteRule ^v3$ https://chainpoint.org/contexts/chainpoint-v3.jsonld [R=302,L]
+RewriteRule ^v4$ https://chainpoint.org/contexts/chainpoint-v4.jsonld [R=302,L]
```
davidlehn: Consider making this just one generic rule, something like:
```
RewriteRule ^v([0-9]+)$ https://chainpoint.org/contexts/chainpoint-v$1.jsonld [R=302,L]
```
Could adjust to be more restrictive as needed, or just let your server reject unknown versions like `01`.  This would let you update your side without needed to update rules here.

PR #1566, opened 2019-11-28
Initial commit for /tern

edmondchuc:


dgarijo:
@edmondchuc,
please note that this is not a repository for hosting files besides the redirections to your URLs. Please remove all the non .htaccess or readme files from the PR.

edmondchuc:
@dgarijo I have removed all files except for the `.htaccess` and `README.md` as requested.

dgarijo:
I recommend doing 303 redirections if you want to do content negotiation, instead of 302. See https://www.w3.org/TR/swbp-vocab-pub/#recipe3

davidlehn:
@dgarijo It's better to get the unwanted files out of the commit history before merging.  Adding a removal commit will still leave those extra files here are in the history.

edmondchuc:
Thanks for the tip @dgarijo. I've made another pull request using 303 redirects instead of 302 redirects. Cheers.

PR #1570, opened 2019-12-02
Documentation new ontology

ncarboni:
Creating an id for a new ontology.


cpr/.gitignore
```diff
@@ -0,0 +1 @@
+.DS_Store
```
davidlehn: This file isn't needed since that ignore rule is in the top level .gitignore.

PR #1588, opened 2019-12-17
Digital Construction Ontologies - Reorganization

seppotorma:
Hi, 

We are preparing a large ontology suite for digital construction, both for new construction and renovation projects. Since there are going to be a lot of internal imports between ontologies and our development is decentralized, it would be easier if we could import ontologies from permanent addresses at w3id.org/digitalconstruction.  

Currently there are only couple of ontologies published in the redirected addresses at digitalconstruction.github.io (Entities and Agents). Three others will be before the next week (Information, Processes and Contexts) and the rest five ontologies by mid January. All ontologies have been designed and initially formalized in OWL.

We are currently also discussing future research efforts that would utilize these ontologies. 

Best regards,
Seppo Törmä

### DIFF
--- a/digitalconstruction/.htaccess
+++ b/digitalconstruction/.htaccess
@@ -1,1 +1,1 @@
-Redirect 301 index.html https://digitalconstruction.github.io/index.html
+Redirect /index.html https://digitalconstruction.github.io/index.html
--- a/digitalconstruction/.htaccess
+++ b/digitalconstruction/.htaccess
@@ -1,1 +1,3 @@
-Redirect /index.html https://digitalconstruction.github.io/index.html
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^$ https://digitalconstruction.github.io/index.html [R=302,L]
--- a/digitalconstruction/.htaccess
+++ b/digitalconstruction/.htaccess
@@ -1,0 +1,1 @@
+Redirect 301 index.html https://digitalconstruction.github.io/index.html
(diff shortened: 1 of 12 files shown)


dgarijo:
@seppotorma,
 the following htaccess redirect is wrong: 
Redirect 301 index.html https://digitalconstruction.github.io/index.html
Also, note that you are not doing proper content negotiation on the html and you return it by default in most configurations. For example, if you ask for n-triples, you will still receive html.

seppotorma:
> @seppotorma,
> the following htaccess redirect is wrong:
> Redirect 301 index.html https://digitalconstruction.github.io/index.html

I changed this to:
    Redirect /index.html https://digitalconstruction.github.io/index.html
> Also, note that you are not doing proper content negotiation on the html and you return it by default in most configurations. For example, if you ask for n-triples, you will still receive html.

I included n-triples and tried to make the rewrites more specific. 

Is it ok now? (Or is another pull request needed?)

Seppo

dgarijo:
@seppotorma,
no, it would be like the other htaccess. For example, something like:
Options +FollowSymLinks
RewriteEngine on
RewriteRule ^$ https://digitalconstruction.github.io/index.html [R=302,L] 
If you want w3id.org/digitalconstruction redirect to https://digitalconstruction.github.io/index.html
As for the default, it is not wrong to return html, but I usually like to add the following rule, which is html specific:
# Rewrite rule for latest version.
RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+xml)
RewriteRule ^$ YOUR ONTOLOGY DOC URL [R=303,L]

However, that is up to you.

seppotorma:
I made the fixes you suggested. I hope it is ok now. 
Thanks for advice!
Seppo

seppotorma:
Thank you very much! 

However, I found a small problem in an xml redirect. I fixed, commited and pushed it. 
Do I need to make another pull request or can you merge that?

Seppo

dgarijo:
I cannot merge it because it has already been merged. You'll have to submit a new PR

PR #1596, opened 2020-01-09
Added trailing slash for all resources.

edmondchuc:


tern/.htaccess
```diff
@@ -3,86 +3,98 @@ RewriteEngine On
...
-# https://w3id.org/tern/ontologies/org
-######################################
+#######################################
+# https://w3id.org/tern/ontologies/org/
+#######################################
 
 # Turtle
 RewriteCond %{QUERY_STRING} ^_format=text/turtle$ [OR]
 RewriteCond %{HTTP:Accept} text/turtle [NC]
+RewriteRule ^ontologies/org/$ https://raw.githack.com/ternaustralia/ontology_tern-org/master/docs/tern-org.ttl [R=303,L]
```
davidlehn: You can simplify all these and use one rule that accepts both forms.  Just make the `/` optional:
```
RewriteRule ^ontologies/org/?$ https://...
```

edmondchuc: Thanks for the tip. I've removed the redundant rule and added the regex optional `?`.

davidlehn: Merged, but then noticed this change and one below still have the redundant rule.  Feel free to change later, should work either way.

PR #1597, opened 2020-01-09
created readme and htaccess

vojtechhuser:
Hi. Please add the CONSIDER statement

### DIFF
--- a/CONSIDER/.htaccess
+++ b/CONSIDER/.htaccess
@@ -1,0 +1,3 @@
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^$ https://github.com/lhncbc/CDE/tree/master/CONSIDER [R=301]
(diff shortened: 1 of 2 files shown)


CONSIDER/.htaccess
```diff
@@ -0,0 +1,3 @@
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^$ https://github.com/lhncbc/CDE/tree/master/CONSIDER [R=301]
```
davidlehn: This is going to the github interface directly, is that what you wanted?

vojtechhuser: Yes, we wanted that.. We hope to move the stuff later to either our own website or to a website maintained by National Library of Medicine. But for now, that is the best target URL and what we intended.

PR #1609, opened 2020-01-21
permalinks for Griebel_et_al_2020_GRL paper

AnneGriebel:
First time user, hope this works.


davidlehn:
We've been suggesting personal pages go in the `people/` subdir.  Would you mind that?  Also could you add info on who the maintainer(s) are for the files to a README.md or as .htaccess comments.  Thanks.

dgarijo:
@AnneGriebel, you don't need to add a readme file per subfolder; I think it's fine with one at the highest level. I will merge your PR, but that will make it easier to maintain your w3ids in the future :)

PR #1618, opened 2020-01-30
One target link was wrong

d3cebal:
Do I always have to open a pull request when I change something?


davidlehn:
Yes, PRs are how you make changes at the moment.  Usage of this service has been targeted at redirections that change rarely, so this has worked out ok so far.

PR #1628, opened 2020-02-09
Added redirects to AtomGraph's open-source project ontologies

namedgraph:
Under the `/atomgraph` namespace

### DIFF
--- a/atomgraph/README.md
+++ b/atomgraph/README.md
@@ -1,0 +1,15 @@
+# AtomGraph namespace
+## Ontologies of AtomGraph's open-source projects
+
+Homepage:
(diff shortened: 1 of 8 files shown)


namedgraph:
@dgarijo does the test hook do anything? Because it passed but some of my links are not working.
I guess I should avoid folders and collect all URLs into one `.htaccess`?
But the testing is useless then

dgarijo:
@namedgraph the check hasn't worked well for a long time, so I wouldn't pay much attention to it right now. You may have multiple folders, that shouldn't be an issue. 
Which links are not working for you? The ones I tried work fine

PR #1653, opened 2020-02-27
Update DID v1 context file location

BigBlueHat:
The CCG's context file is missing things like `RsaVerificationKey2018` (probably due to changes happening now that DID is a WG). So, this should get the w3id.org URL pointing to the correct DID v1 context file (afaik).

Thanks!
🎩

### DIFF
--- a/did/.htaccess
+++ b/did/.htaccess
@@ -2,5 +2,5 @@ Header set Access-Control-Allow-Origin *
 Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
 Options +FollowSymLinks
 RewriteEngine on
-RewriteRule ^$ https://w3c-ccg.github.io/did-spec/ [R=302,L]
-RewriteRule ^v(.+)$ https://w3c-ccg.github.io/did-spec/contexts/did-v$1.jsonld [R=302,L]
+RewriteRule ^$ https://www.w3.org/ns/did/ [R=302,L]
+RewriteRule ^v(.+)$ https://www.w3.org/ns/did/v$1 [R=302,L]


davidlehn:
We should also add a readme with links and maintainers.

did/.htaccess
```diff
@@ -2,5 +2,5 @@ Header set Access-Control-Allow-Origin *
 Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
 Options +FollowSymLinks
 RewriteEngine on
-RewriteRule ^$ https://w3c-ccg.github.io/did-spec/ [R=302,L]
-RewriteRule ^v(.+)$ https://w3c-ccg.github.io/did-spec/contexts/did-v$1.jsonld [R=302,L]
+RewriteRule ^$ https://www.w3.org/ns/did/ [R=302,L]
```
davidlehn: This link doesn't work.  Should it point at latest did-core spec?
```suggestion
RewriteRule ^$ https://w3c.github.io/did-core/ [R=302,L]
```

davidlehn:
Currently it is just hosting two files, v0.11 and v1:
https://github.com/w3c-ccg/did-spec/tree/gh-pages/contexts
For backwards compatibility there should probably be a initial check for v0.11 since that won't exist if wildcarding to the w3.org namespace.

I'm also unsure if maybe this should explicitly only support v0.11 and v1 rather than wildcards.  That would force future links to use the official w3.org namespace links.

Here's how to see the v1 context and headers diffs to make sure this is ok:
```sh
diff -u <(curl https://w3c-ccg.github.io/did-spec/contexts/did-v1.jsonld) <(curl https://www.w3.org/ns/did/v1)
diff -u <(curl -I https://w3c-ccg.github.io/did-spec/contexts/did-v1.jsonld | sort) <(curl -I https://www.w3.org/ns/did/v1 | sort)
```

did/.htaccess
BigBlueHat: Good question. I'd forgotten that the W3C doesn't catch directory requests.

I'd just remove it...but it's up to y'all.

BigBlueHat:
@msporny no worries! We're implementing DIDs atm, and just had this (old?) URL in use. We're switching to the W3C one anyhow, but wanted to make y'all aware that things were out of sync. Handle it how you will. 😄

OR13:
This remains blocked by the painfully slow pace of the DID WG.

OR13:
`https://w3id.org/did/v1` -> `https://www.w3.org/ns/did/v1` (PASS)

`https://w3id.org/did/v0.11` -> `https://www.w3.org/ns/did/v0.11` (FAIL)

This PR is blocked by the W3C setting up support for v0.11

OR13:
+1 to closing this PR

dgarijo:
Closing the PR due to inactivity.

PR #1658, opened 2020-03-05
Remove redirects to give possession to IEEE

HadleyKing:
The https://gitlab.com/IEEE-SA/2791/ieee-2791-schema/-/raw/master/2791object.json redirect has been removed so that IEEE will be able to own the redirect repo under their own space. 
@joshgay


dgarijo:
@HadleyKing,
you are not redirecting to anything anymore. Is this correct?

HadleyKing:
Yes, I think so. Thank you. I did not see anything in the docs about how to do this but I am actually trying to transfer ownership of the redirects so I thought the best way was to delete my redirects and then have the other org commit with their redirects. LMK if there is a better/more appropriate way.

dgarijo:
@HadleyKing,
what I would do is the following: 
 - Add them (the other organization) to the readme file as maintainers.
 - Do not modify the redirection; if they are maintainers, then they will add the right link. when they consider appropriate

We don't support reserving empty w3ids, so if you delete yours, there is nothing preventing anyone else claiming it for themselves. 
I hope this helps

HadleyKing:
> @HadleyKing,
> what I would do is the following:
> 
> * Add them (the other organization) to the readme file as maintainers.
> * Do not modify the redirection; if they are maintainers, then they will add the right link when they consider appropriate
> 
> We don't support reserving empty w3ids, so if you delete yours, there is nothing preventing anyone else claiming it for themselves.
> I hope this helps

@dgarijo 
Thanks you, that does help. Please delete it. We do not need the current URL. IEEE (@joshgay) will manage it on their own.

dgarijo:
@HadleyKing, 
then please update your PR to remove the readme file as well and I will merge it.

PR #1663, opened 2020-03-15
Add the trek namespace

vemonet:
Add the trek namespace. Used by a clinical data KG for the NIH NCATS Translator project

Redirect to https://trek.semanticscience.org


dgarijo:
@vemonet,
the link  https://trek.semanticscience.org is returning a connection refused error. Is it supposed to work?

vemonet:
@dgarijo sorry, I mistakenly used HTTPS instead of HTTP, thanks for noticing it

PR #1671, opened 2020-03-19
modified for the new configuration at Maastricht

markwilkinson:


FAIR_Tests/.htaccess
```diff
@@ -1,4 +1,3 @@
 Options +FollowSymLinks
 RewriteEngine on
-RewriteRule ^ https://fair-tests.137.120.31.101.nip.io%{REQUEST_URI} [R=307,L]
-
+RewriteRule FAIR_Tests/(.*) https://fair-tests.137.120.31.101.nip.io/$1 [R=307,L]
```
davidlehn: Rules are relative to current directory.  This will result in http://w3id.org/FAIR_Tests/FAIR_Tests/...

PR #1676, opened 2020-03-23
Add namespace for Fraunhofer

mrf-ipt:
I would like to add a namespace for ontologies of Fraunhofer institutes.
A first use for this namespace are our ontologies for the lighthouse project EVOLOPRO.

### DIFF
--- a/fraunhofer/Readme.md
+++ b/fraunhofer/Readme.md
@@ -1,0 +1,4 @@
+# Fraunhofer namespace
+This namespace is for ontologies developed at Fraunhofer institutes.
+
+https://www.fraunhofer.de/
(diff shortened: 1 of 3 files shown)


dgarijo:
@mrf-ipt,
while we do not require a readme for each individual w3id, we would like to know who is the general maintainer (name, email and github id) of the main w3id. Could you please add that in the main readme file?

mrf-ipt:
Hi @dgarijo, it's a bit difficult because Fraunhofer is a very large organization and each institute is more or less independent. There is no central department which would be the right contact for this. I added a colleague and myself as contact but this might change every now and then. I hope that's OK.

PR #1694, opened 2020-04-07
updated BOT for more mediatypes

maximelefrancois86:


dgarijo:
@maximelefrancois86, you are redirecting to bot and index.html, two files that are not present in the system. Have you tested this redirection locally? I think it may not work

maximelefrancois86:
Hi Daniel,

Yes I tested locally.

Those redirections are internal first, and not final (no [L] tag)

The final redirect rule is applied after this in all cases:

```
RewriteRule ^(.*)$ https://w3c-lbd-cg.github.io/bot/$1 [R=302,L]
```

PR #1708, opened 2020-04-27
added paths to defined vocabulary

sroertgen:
Hello,

I added paths to vocabulary we defined.


openeduhub/.htaccess
```diff
@@ -1,3 +1,16 @@
 RewriteEngine on
 RewriteRule ^vocabs/(.*) https://vocabs.openeduhub.de/$1 [R=302,L]
+RewriteRule ^vocabs/category/(.*) https://vocabs.openeduhub.de/w3id.org/openeduhub/vocabs/category/$1 [R=302,L]
```
davidlehn, line 3: Rules are checked in order so the first "^vocabs/(.*)" will match before these new rules.  Probably just need to move the new ones so they are checked first.

sroertgen: I'm sorry, I did not know that.
I corrected as suggested and will check beforehand next time. Thank you!

PR #1709, opened 2020-04-27
fix rule for France

pyrog:


davidlehn:
@cKlee Merging this since it looks correct.  @pyrog isn't listed as a maintainer in the README.md though.  Let us know if that's a problem.

The style is a tad different than the other rules that capture (but don't use) the country prefix part.  Same effect either way.

pyrog:
Thank you very much 🌞

cKlee:
Thanks!

PR #1730, opened 2020-05-13
HTTP ontology

mthl:
Hello,

I would like to register the **http://w3id.org/http** namespace for an OWL ontology of HTTP interactions. Please let me know if you need more information.

Thanks

### DIFF
--- a/http/.htaccess
+++ b/http/.htaccess
@@ -1,0 +1,5 @@
+Header set Access-Control-Allow-Origin *
+Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
+Options +FollowSymLinks
+RewriteEngine on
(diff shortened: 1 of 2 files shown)


OR13:
Interesting... any information about labs.nereide.fr / how this ontology will be maintained / used...

mthl:
@OR13 This is early research work so maintenance and usage is pretty limited to that scope currently. `labs.nereide.fr` is the Gitlab CE instance of [Néréide](https://nereide.fr/) where this ontology is developed.

@dgarijo Thanks for merging.

davidlehn:
I do have a concern about use of a common name like "http".  Looks like this currently being used only as a way to fetch the ontology file?  Was the plan to define ids in this namespace too or just use the w3c one?  In any case, I imagine it'll be ok since the current use looks appropriate.

mthl:
Yes the plan is to define ids in this namespace instead of reusing the one from the W3C as it is currently done. I am in the process of migrating those ids.

PR #1742, opened 2020-05-22
Updated .htaccess

ecow:
Now redirects to gitlab pages to allow better serving of mime types and better perrformances


dgarijo:
https://mopso.gitlab.io/amlo/ returns a 404. Is the link correct?

ecow:
https://mopso.gitlab.io/amlo/  is not a valid url.

A valid url is  https://mopso.gitlab.io/amlo/core/core.rdf (for instance)

See redirects rules

PR #1750, opened 2020-05-29
updated .htaccess

JoaoMFCardoso:
Changed the url to point to the semantics.id/ns/dcso


davidlehn:
Note that the old and new targets are all returning 404 now.

PR #1771, opened 2020-06-09
Updated the entry for QUDT

steveraysteveray:


dgarijo:
@davidlehn @KMax @steveraysteveray,
I don't know very well what should be done here.

On the one hand, @KMax created the w3id a long time ago (5years), and @steveraysteveray's PR is not conflicting with that w3id. But @KMax is the owner of this w3id. On the other hand, qudt.org existed before this w3id, and in fact the work by @KMax extends it. I think it makes sense because many people are using qudt. 

@KMax would you be ok if we add @steveraysteveray's redirect to the root of this w3id? 

Another question @davidlehn is: can someone create w3ids of domains they don't directly own? For example, should we allow a w3id called "google"?

KMax:
@dgarijo Yes, of course, I'm okay and I'm more than glad to see this PR :)

davidlehn:
Ah, good.  Let people sort it out themselves is always the easiest solution. ;-)

I find the "update as of" comments a bit odd.  Seems better to have more declarative info in readmes, but no big deal.

@dgarijo Re using names you don't "own", we'll just have to deal with that as it comes up and do our best.

steveraysteveray:
Thanks everyone. I just stumbled on this repo following a meeting where someone mentioned it. I'm CEO of QUDT.org, so I guess I'm the owner anyway!

I intentionally left the extension language by @KMax because I didn't want to interfere with his extension work. However, much has changed in the last 5 years. Please feel free to get in touch if you would like to discuss.

dgarijo:
Ok, then I will merge

PR #1773, opened 2020-06-09
Added redirects for SXACML project.

mdrozdo:


mdrozdo:
Sorry, spotted a mistake. Will correct promptly.

PR #1786, opened 2020-06-29
Digitalconstruction: redirects to new ontology modules

seppotorma:
As a result of refactoring and refinements on digital construction ontologies, we would need to create redirects to modules Activities, EnergyEfficiency, Lifecycle, and Variables. 

- Seppo

### DIFF
--- a/digitalconstruction/Activities/.htaccess
+++ b/digitalconstruction/Activities/.htaccess
@@ -1,0 +1,37 @@
+Options +FollowSymLinks
+RewriteEngine on
+Options -MultiViews
+AddType text/turtle .ttl
+
+# Redirect based on file suffix (ttl, xml, owl, json, jsonld, nt, html)
+RewriteRule ^ontology.ttl$ https://digitalconstruction.github.io/Activities/ontology.ttl [R=302,NE,L]
+RewriteRule ^ontology.xml$ https://digitalconstruction.github.io/Activities/ontology.xml [R=302,NE,L]
+RewriteRule ^ontology.owl$ https://digitalconstruction.github.io/Activities/ontology.xml [R=302,NE,L]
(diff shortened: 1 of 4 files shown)


davidlehn:
I didn't look too close, but it appears all these dirs are the same in how the redirect to github?  You might be able to simplify all the rules down into the top level and use pattern replacement to cover current and future names.

seppotorma:
Yes, I believe so too, but I'm somewhat unfamiliar with how to write the
redirects in a correct way. I have had some bad experiences when trying to
be too smart with redirects.

If you could point me to an example .htaccess that would work correctly in
at top level, I could try to replace current versions. Are there other
ontologies in w3id that have multiple different modules, and use a
top-level .htaccess?

Seppo

ma 29. kesäk. 2020 klo 22.23 David I. Lehn (notifications@github.com)
kirjoitti:

davidlehn:
Not sure what to point at... there are probably many pattern replacement uses.  It would basically be moving all the rules from one subdir up to your root .htaccess and changing all the lines to be generic and capture the path pattern and replace it in the redirect.  In theory if all those subdir rules are the same, and the dir names match the associated ones on github, they would be redundant and could be removed.  There may be edge cases or things I don't know about your use case, but the below might help.  Would be best to test such things locally as this is untested.

```diff
-RewriteRule ^ontology.ttl$ https://digitalconstruction.github.io/EnergyEfficiency/ontology.ttl [R=302,NE,L]
￼+RewriteRule ^(.*)/ontology.ttl$ https://digitalconstruction.github.io/$1/ontology.ttl [R=302,NE,L]
```
```diff
-RewriteRule ^$ https://digitalconstruction.github.io/EnergyEfficiency/ontology.ttl [R=303,NE,L]
￼+￼RewriteRule ^(.*)/?$ https://digitalconstruction.github.io/$1/ontology.ttl [R=303,NE,L]
￼```

PR #1788, opened 2020-06-29
Update .htaccess

luigi-asprino:


dgarijo:
@luigi-asprino,
you are trying to  modify a w3id on which you are not listed as a maintainer. The last commit was done by @anuzzolese. Can he clarify whether you are a maintainer of this w3id?  

If so, please add yourself to the readme file.

luigi-asprino:
Hi @dgarijo 
@anuzzolese added me as a contributor of its repository, I thought that this was enough to give me the rights to update the w3id.
By the way, yes I'm a maintainer of this w3id and I've updated the README

Thanks

arco/.htaccess
```diff
@@ -17,7 +17,11 @@ RewriteRule ^ontology/([a-zA-Z-_0-9]+)(\-aligns)(/?)$ %{ENV:ROOT_URL}/ontologie/
 RewriteCond %{HTTP_ACCEPT} ^.*text/html.*$
 RewriteRule ^ontology/([a-zA-Z-_0-9]+)(/)?$ %{ENV:LODE_URL}%{ENV:ROOT_URL}/ontologie/$1/$1.owl [R=303,L]
 
+RewriteCond %{HTTP_ACCEPT} ^.*text/html.*$
+RewriteRule ^ontology/([a-zA-Z-_0-9]+)/([0-9].[0-9])(/)?$ %{ENV:LODE_URL}%{ENV:ROOT_URL}/ontologie/$1/$2/$1.owl [R=303,L]
+
 RewriteRule ^ontology/([a-zA-Z-_0-9]+)(/)?$ %{ENV:ROOT_URL}/ontologie/$1/$1.owl [R=303,L]
+RewriteRule ^ontology/([a-zA-Z-_0-9]+)/([0-9].[0-9])(/)?$ %{ENV:ROOT_URL}/ontologie/$1/$2/$1.owl [R=303,L]
```
davidlehn, line 24: Is the change two lines up doing the same match?  Seems like this line wouldn't be hit.  Similar with the line above matching an earlier line.  Also seems the html RewriteConds are mostly redundant?  Maybe I'm missing some tricks here.

davidlehn:
@luigi-asprino We're just trying to be more strict about adding contributors so we don't get confused about who should be changing files.

anuzzolese:
I confirm that @luigi-asprino should be listed among the maintainers of this w3id.

arco/.htaccess
luigi-asprino: I have to admit that I'm not expert, I was trying to define a rule matching uris with a trailing number (e.g. ontology/arco/1.0).
The lines that I added match the same URIs but the first rule should be conditioned to the content type required.

davidlehn: I'm no expert either. ;-)  I just try rules until things work.  And apparently I get easily confused looking at other peoples rules.  I think this is fine.  It'll use that one rule if it's html, else the owl rule.

PR #1794, opened 2020-07-03
Update .htaccess

luigi-asprino:
Let's try again.


davidlehn:
Not sure what you're trying to do.  We can probably help if needed.  Rules sometimes get tricky.

dgarijo:
You can also try them locally (I usually try with XAMPP, which includes an apache server).

luigi-asprino:
hi thanks, it seems working now :-)
I just wanted to attach a postfix to the ontology uri for specifying the version number, e.g. 

https://w3id.org/arco/ontology/arco resolves to the current version of the ontology which is the 1.1
https://w3id.org/arco/ontology/arco/1.0 resolve to the version 1.0

sorry for bothering you with multiple pull requests

Thanks for your help

PR #1795, opened 2020-07-06
Addingnewperson

pantoner:
Added a new person to repo


dgarijo:
@pantoner,
can you please add your github id as a maintainer? From the email you used I cannot know that you were resposible for creating the w3id, and I need that to be clear in case you . Thanks in advance.

PR #1798, opened 2020-07-11
Create .htaccess

adrioro:
"You Can Do It (round 2)" Painting by Adriana Cerrotti


dgarijo:
@adrioro,
you are trying to modify a w3id that you didn't author, and where you are not listed as a maintainer.
Can we get the ok to merge by @pantoner ?
And if we do, it would be great to add yourself as a maintainer in the ids so I don't have to ask again :)
Thanks in avdance.

pantoner:
@adrioro, is ok to merge.

PR #1806, opened 2020-07-22
Fix vocabulary URL in CIT vocab.

msporny:


davidlehn:
This doesn't seem to be needed.

msporny:
Ignore... Chromium caching issue that didn't go away w/ force reload.

PR #1826, opened 2020-08-18
Add three additional redirections to w3id.org/hdgi

madhawap:
This will add the following URL redirections to https://w3id.org/hdgi
1. https://w3id.org/hdgi/v0.1 ->  https://w3id.org/hdgi
2. https://w3id.org/hdgi/mappings-apis -> GitHub repo
3. https://w3id.org/hdgi/mappings-docs -> GitHub repo

### DIFF
--- a/hdgi/.htaccess
+++ b/hdgi/.htaccess
@@ -44,13 +44,16 @@ RewriteCond %{HTTP_ACCEPT} .+
 RewriteRule ^$ https://madhawap.github.io/human-device-gesture-interaction-ontology/v0.1/406.html [R=406,L]
 
 # Redirect https://w3id.org/hdgi/v0.1 to https://w3id.org/hdgi
-Redirect /hdgi/v0.1 https://w3id.org/hdgi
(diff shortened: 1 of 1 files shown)


hdgi/.htaccess
```diff
@@ -43,6 +43,15 @@ RewriteRule ^$ https://madhawap.github.io/human-device-gesture-interaction-ontol
 RewriteCond %{HTTP_ACCEPT} .+
 RewriteRule ^$ https://madhawap.github.io/human-device-gesture-interaction-ontology/v0.1/406.html [R=406,L]
 
+# Redirect https://w3id.org/hdgi/v0.1 to https://w3id.org/hdgi
+Redirect /hdgi/v0.1 https://w3id.org/hdgi
```
davidlehn, line 47: These are relative to the current `hdgi` directory, so you don't want the `/hgdi/` prefix.  Usually should be something more strict like `^v0.1$`.  Similar below.

madhawap: Changed Redirect to RewriteRule with regex and tested locally.

PR #1838, opened 2020-08-23
Add redirect for security/v3

OR13:
@dlongley @tplooker @davidlehn @msporny This will allow us to start using security/v3 with "default" document loaders.

It will also help with automated linting / documentation tooling for properties that are not not in sec-v2.

### DIFF
--- a/security/.htaccess
+++ b/security/.htaccess
@@ -5,4 +5,4 @@ RewriteEngine on
 RewriteRule ^$ https://w3c-ccg.github.io/security-vocab/ [R=302,L]
 RewriteRule ^v1$ https://w3c-ccg.github.io/security-vocab/contexts/security-v1.jsonld [R=302,L]
 RewriteRule ^v2$ https://w3c-ccg.github.io/security-vocab/contexts/security-v2.jsonld [R=302,L]
-RewriteRule ^v3$ https://w3c-ccg.github.io/security-vocab/contexts/security-v3-unstable.jsonld [R=302,L]
+RewriteRule ^v3-unstable$ https://w3c-ccg.github.io/security-vocab/contexts/security-v3-unstable.jsonld [R=302,L]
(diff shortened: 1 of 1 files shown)


security/.htaccess
```diff
@@ -5,3 +5,4 @@ RewriteEngine on
 RewriteRule ^$ https://w3c-ccg.github.io/security-vocab/ [R=302,L]
 RewriteRule ^v1$ https://w3c-ccg.github.io/security-vocab/contexts/security-v1.jsonld [R=302,L]
 RewriteRule ^v2$ https://w3c-ccg.github.io/security-vocab/contexts/security-v2.jsonld [R=302,L]
+RewriteRule ^v3$ https://w3c-ccg.github.io/security-vocab/contexts/security-v3-unstable.jsonld [R=302,L]
```
msporny, line 8: We should name this v3-unstable... if we don't, people will ship with an experimental context and blame us.

OR13: @msporny yep I agree... where do we write that decision down? 

something to the effect of "URLs that contain unstable may have their resources altered without notice".

"URLs that do not contain unstable will never be updated.

OR13, line 8:
```suggestion
RewriteRule ^v3-unstable$ https://w3c-ccg.github.io/security-vocab/contexts/security-v3-unstable.jsonld [R=302,L]
```

OR13: I have updated the unstable URL, it will be:

`https://w3id.org/security/v3-unstable`

davidlehn:
@dgarijo I think we were still discussing this. ;-)  Probably should delay on merging PRs when there's active discussion with reviews requested.  Probably not a big deal here as we can adjust rules with another PR if needed.

OR13:
agree with @davidlehn however, this case its even safer given we warned people in the URL that its unstable :)

dgarijo:
@davidlehn @OR13 
Oh sorry, I thought the discussion was resolved because the latest change addressed the request.

davidlehn:
The rules here could be simplified to just let the target repo handling naming conventions.  I'm having dejavu this may have been previously discussed and rejected though?  Anyway, could be something like this:
```apache
RewriteRule ^v(.+)$ https://w3c-ccg.github.io/security-vocab/contexts/security-v$1.jsonld [R=302,L]
```

OR13:
+1 to the suggestion @davidlehn

PR #1842, opened 2020-08-25
Update .htaccess

rapw3k:
I logged the requests from Protege (Mac). It tries multiple times using two different requests, with the headers below. So, no turtle is even considered, and it seems text/html has higher weight than text/turtle. see issue: https://github.com/protegeproject/protege/issues/973 . In meantime, adding rdf+xml sending to ttl to try make it Protege to load.

accept: 'application/rdf+xml, application/xml; q=0.7, text/xml; q=0.6, text/plain; q=0.1, /; q=0.09', 'accept-encoding': 'xz,gzip,deflate', 'user-agent': 'Java/1.8.0_121'

accept: 'text/html, image/gif, image/jpeg, *; q=.2, /; q=.2',


rapw3k:
redirection from rdf+xml to ttl didnt work, I created rdf/xml version

PR #1870, opened 2020-09-24
Dictionary 00

stcruy:
– The name is explained in detail in 00/README.md.

### DIFF
--- a/00/.htaccess
+++ b/00/.htaccess
@@ -1,0 +1,3 @@
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^(.*)$ https://vsm.github.io/dict/00/$1 [R=302,L]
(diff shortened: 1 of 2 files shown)


davidlehn:
Hard to tell what's going on here. ;-)  I'll just merge it.  If programs start dereferencing all numbers though w3id.org, that seems rather inefficient!  Numbers seem like a primitive concept that could be special cased.  Maybe a urn scheme that implementations would be required to know about.  In any case, fine for now, enjoy.

stcruy:
Thanks!
And indeed :-) Programs would have to know about the generative nature of a CV like this, and not look them up!

The only intention is to have a permanent base-URI. Like that, numeric concepts can occasionally be used (and contextualized, see VSM) in a "knowledge structure" setting, not in a "data sheet" setting. – Other such CVs I can think of could 'generate' concept IDs for years, or months, etc.
Thanks for now :)

PR #1890, opened 2020-10-13
add new project: Hieratische Paläographie DB

nakamura196:


hpdb/.htaccess
```diff
@@ -0,0 +1,7 @@
+Options +FollowSymLinks
+RewriteEngine on
+
+RewriteCond %{REQUEST_URI} /api/
+RewriteRule ^(.*)$ https://hp-db.github.io/$1\.json [L,R=301]
```
davidlehn, line 5: Is this going to use redirects for an application api?  That may be a bit out of scope for what w3id.org should be used for.  Probably not very efficient either.  If this is just a small project, then no big deal, but if any significant amount of traffic will happen, I'd consider some other way to do this.

nakamura196: Thank you for your advice.
This is not for the redirect for API, though the pathname contains the string "api".

I would like to access to the json file on github pages without suffix.

thanks.

PR #1900, opened 2020-10-20
added CCOnto as w3id-links

amanallahrafat:


dgarijo:
@amanallahrafat,
you may want to change:

`^(.*)/cconto` to `^cconto`; otherwise you will not be doing the redirection you expect in your readme file. This applies to both of your  redirects.

dgarijo:
Sorry, you also need the trailing "$": `^cconto$`

PR #1921, opened 2020-11-11
Update portal redirect

HadleyKing:


dgarijo:
I think the second redirection generalizes the first one. Since both resolve to the same link, the first one is not necessary

PR #1931, opened 2020-11-21
Update .htaccess

luigi-asprino:
Hi,

I'm involved in the SPICE project so I have the rights to update this perma-id.
@essepuntato will confirm it

Best


essepuntato:
I confirm

PR #1966, opened 2021-01-07
Correctes rules in htaccess file

tanktoo:


dgarijo:
@tanktoo,
You are creating the w3id: https://w3id.org/iot/fd/fd
Is this intended (note the double `fd`)?

tanktoo:
Hm, for https://w3id.org/iot/qoi/ it is working with the same .htaccess file. Do I miss something here?

dgarijo:
If you want https://w3id.org/iot/fd, you rewrite rule should be `RewriteRule ^$` instead of `RewriteRule ^fd$`. Note that you are already in the `fd` folder

tanktoo:
Ok, did a new merge request. Closed this one. Thanks for help!

PR #1973, opened 2021-01-15
digitalconstruction - rules fixed

seppotorma:
Fixing minor bugs in the rules


davidlehn:
It looks like you might be able to replace all those files with one top-level file that uses pattern replacement.  If so, it would be easier to maintain.

PR #1980, opened 2021-01-18
Add subcontext splitting redirects

OR13:
This PR implements context splitting for https://w3id.org/security 

If this PR is merged, we should remove:

- https://w3id.org/pgp/v1
- https://w3id.org/jws/v1
- https://w3id.org/bbs/v1

And instead use:

- https://w3id.org/security/pgp/v1
- https://w3id.org/security/jws/v1
- https://w3id.org/security/bbs/v1

If this PR is not merged, the above will continue to work :)


OR13:
ping @davidlehn @msporny @dlongley @tplooker the goal here is to make it easier to pull "only what you need" in terms of context definitions, while also centralizing around a shared namespace, so that its easy to take contributions regarding security vocab.

PR #1984, opened 2021-01-21
Adding devops-infra catalogue of ontologies and the first ontology, to check that redirections work well

ocorcho:


devops-infra/.htaccess
```diff
@@ -1,5 +1,48 @@
...
+AddType application/rdf+xml .rdf
+AddType application/rdf+xml .owl
+AddType text/turtle .ttl
+AddType application/n-triples .n3
+AddType application/ld+json .json
+# Rewrite engine setup
+RewriteEngine On
+
+#If no specific ontology is requested, return the HTML ontology catalogue 
+RewriteRule ^.*$ https://oeg-upm.github.io/devops-infra/index.html [R=302,L]
```
dgarijo, line 18: If you have this rule first, it will superseed all subsequent rules (you are matching anything). You should have this one at the end

dgarijo:
@ocorcho please have a look at my comment. I can merge the PR as it is, but it will probably not work as intended.

ocorcho:
You were right. thanks!! Change is now made

PR #1994, opened 2021-01-28
ID for the digital twin of a crane

juusoautiosalo:
Created an identifier for the digital twin of an overhead crane, along with a folder "twins" which can be used for more twin IDs.

This is relates to an ongoing development of a digital twin framework that needs an external ID provider.


dgarijo:
@juusoautiosalo,
if you plan to create multiple folders, it is not necessary to add a readme file on each of them. One at the root level is fine :)

PR #1997, opened 2021-01-29
Adding more ontologies to be redirected in devops-infra

ocorcho:


dgarijo:
@ocorcho,
this will not work as intended, because the rewrite condition applies to only the first rewrite rule below.

According to https://stackoverflow.com/questions/7218164/multiple-rewriterules-for-single-rewritecond-in-htaccess you can tie multiple rewrite rules to a single rewrite condition, but I have not tried it myself:

```
RewriteCond  %{HTTP_HOST}  !^www.mydomain.com$
# skip rules if NOT within domain - only way to tie multiple rules to one cond
RewriteRule  .?  -  [S=3]
RewriteRule  ^path1(/.*)$  /otherpath1$1
RewriteRule  ^path2(/.*)$  /otherpath2$1
RewriteRule  ^path3(/.*)$  /otherpath3$1
``` 
I personally find easier to maintain creating individual folders for each ontology (in particular when different versions are involved), but I will leave that to you.

davidlehn:
@ocorcho It also looks like you have very regular redirects.  You may be able to replace almost everything with simple pattern replacement rules.   As a bonus those would simplify the conditionals issue.

ocorcho:
Thanks for all your suggestions. 

I have now changed everything to make it folder oriented, which simplifies things as I am maintaining them. I will continue working on some files that I need to generate, so there will be a few more changes in the near future, but I would appreciate it if you can accept these ones (which should work now)

dgarijo:
@ocorcho,
if you want to follow the best practices for Semantic Web resources, your vocabulary redirects should be 303, not 302. I will merge, because they will work. But I suggest changing them

devops-infra/def/network/.htaccess
```diff
@@ -0,0 +1,49 @@
...
+RewriteCond %{HTTP_ACCEPT} text/\* [OR]
+RewriteCond %{HTTP_ACCEPT} \*/turtle 
+RewriteRule ^.*$ https://oeg-upm.github.io/devops-infra/ontology/network/ontology.ttl [R=302,L]
+
+RewriteCond %{HTTP_ACCEPT} .+
+RewriteRule ^network$ https://oeg-upm.github.io/devops-infra/ontology/network/406.html [R=302,L]
+RewriteRule ^software$ https://oeg-upm.github.io/devops-infra/ontology/software/406.html [R=302,L]
+RewriteRule ^database$ https://oeg-upm.github.io/devops-infra/ontology/database/406.html [R=302,L]
+RewriteRule ^hardware$ https://oeg-upm.github.io/devops-infra/ontology/hardware/406.html [R=302,L]
+RewriteRule ^certificate$ https://oeg-upm.github.io/devops-infra/ontology/certificate/406.html [R=302,L]
```
davidlehn, line 48: I know this has been merged, but just noting these rules differ from your other files.

davidlehn:
@ocorcho When I was noting how regular the files are, that still stands.  You might be able to just have a devops-infra/def/.htaccess file that uses wildcard matching and replacement to have one set of rules that cover all the def/X file use cases.

PR #2004, opened 2021-02-02
Fixing RewriteRule for KDSF-FFK

doerners:
This pr fixes the RewriteRule. Thanks @acka47


davidlehn:
Is this intentional?  Links are https://w3id.org/kdsf-ffk/kdsf-ffk/* and redirect to a similarly doubled up URL?  Did you want to just have one level?

dgarijo:
That was my previous suggestion @davidlehn, but they have reverted it to the double id in the URI, so I guess they want it this way

PR #2011, opened 2021-02-04
feat: add vaccine credentials vocab

tplooker:
Adds Vaccine Certificate Vocabulary for usage with Verifiable Credentials

### DIFF
--- a/credentials/.htaccess
+++ b/credentials/.htaccess
@@ -3,8 +3,3 @@ RewriteEngine on
 #RewriteRule ^$ https://opencreds.org/vocabs/credentials [R=302,L]
 #RewriteRule ^v1$ https://opencreds.org/contexts/credentials-v1.jsonld [R=302,L]
 RewriteRule ^v1$ https://web-payments.org/contexts/credentials-v1.jsonld [R=302,L]
-
-# https://w3id.org/credentials/vaccination
-
-RewriteRule ^vaccination$ https://mattrglobal.github.io/vaccination-certificate-vocab/ [R=302,L]
-RewriteRule ^vaccination/v1$ https://mattrglobal.github.io/vaccination-certificate-vocab/context/v1/ [R=302,L]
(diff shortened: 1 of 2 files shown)


OR13 (approved):

davidlehn:
I'm not sure how we want to handle this.  Usually projects use their own top-level namespace.  That at least helps with managing who has edit control.  The `credentials` namespace is so old we haven't even documented who should be able to update it.  But in this case, I think it would be the CCG.  Is this request for an official CCG project and effectively under CCG control?  If so then it seems like here would be ok.  If not then it may be easier to manage in its own namespace.  The worry is everyone throwing projects in here and it being difficult to manange.

@dlongley @msporny Thoughts?  Who else should

tplooker:
Ping @dlongley and @msporny, @OR13 and I were talking about whether we want to establish a pattern for name-spacing vocabs associated to VCs under the credential route, rather than having them in the root?

OR13:
imo, it would be better to take this approach, because the alternative will be filling up w3id.org... and end of the day, its the remote host that matters... This probably should have been pointed at the latest vc-data-model ages ago...

tplooker:
Appreciate the context @davidlehn, r.e the intent of the project, i do intend to propose it as a CCG work item just have not had the chance to prepare it yet.

OR13:
I am in favor of merging this PR, and of making the CCG code owners for these redirects similar to how we have been evolving w3id.org/security ... which defines many of the same terms as w3id.org/security/jws for example.

msporny:
> imo, it would be better to take this approach, because the alternative will be filling up w3id.org... and end of the day, its the remote host that matters... This probably should have been pointed at the latest vc-data-model ages ago...

This is not a concern, we shouldn't let the server drive the design. The goal here is short, memorable identifiers. I'm going to suggest the PR be changed to be under global top level directory -- /vaccination/ <-- we can accept that immediately.

tplooker:
> This is not a concern, we shouldn't let the server drive the design. The goal here is short, memorable identifiers. I'm going to suggest the PR be changed to be under global top level directory -- /vaccination/ <-- we can accept that immediately.

Ok happy to change, I will amend now

msporny:
> Ok happy to change, I will amend now

Thank you! :)

tplooker:
ping @msporny @OR13 cannot request review in this repo, but should now be ready

msporny (approved):

OR13 (approved):

PR #2015, opened 2021-02-08
added didcomm

awoie:
added didcomm


dgarijo:
@awoie,
why a 302 redirect? wouldn't it be better a 301?

awoie:
@msporny @davidlehn I tried to use the same pattern as for other DID related specs in W3C and W3C CCG. They all use 302. Is there are specific reason why 301 would not work for JSON-LD context/vocab redirects?

davidlehn:
302 is fine.

PR #2019, opened 2021-02-10
Create README.md

ritamargherita:


dgarijo:
@ritamargherita,
the readme file is empty. You should fill it up with a small description, the name of the maintainer, etc.
Also, you are missing the htaccess file with your redirection. Please see other repositories for reference

PR #2032, opened 2021-02-22
New writing rules for managing html files + error correction

danilo-dessi:


ardo/.htaccess
```diff
@@ -1,4 +1,9 @@
...
-RewriteRule ^$ https://ise-fizkarlsruhe.github.io/ArDO/ [R=302,L]
\ No newline at end of file
+Options +FollowSymLinks 
+RewriteEngine on
+
+RewriteRule ^(.*)\.ttl$ https://ise-fizkarlsruhe.github.io/ArDO/2.0/ArDO.ttl [R=302,L]
+
+RewriteRule ^(.*)$ https://ise-fizkarlsruhe.github.io/ArDO/2.0/index.html#$1 [R=302,L]
+
+RewriteRule ^$ https://ise-fizkarlsruhe.github.io/ArDO/2.0/index.html [R=302,L]
```
davidlehn, line 8: Rules are processed in order.  The above rule will catch everything, so this rule will never be hit.  You would need to have this one appear earlier.

danilo-dessi: The first rule should catch "every string" followed by .ttl
The second and third rule should catch everything and I have already used them in another project aikg

Additionally:
The [L] flag causes mod_rewrite to stop processing the rule set. In most contexts, this means that if the rule matches, no further rules will be processed. This corresponds to the last command in Perl, or the break command in C. Use this flag to indicate that the current rule should be applied immediately without considering further rules. (from: https://httpd.apache.org/docs/2.4/rewrite/flags.html)

davidlehn: I'm saying `^$` will never be used because `^(.*)$` will always catch everything.  It's going to the same place, so it doesn't matter in this case, but the last `^$` rule will never be used.

davidlehn: I'll just merge.  Feel free to fix up the ordering if you'd like.

PR #2039, opened 2021-02-25
Added new rules and directories. Changed redirect to github raw format

ritamargherita:


odissei/np/README.md
```diff
@@ -0,0 +1,3 @@
+Nanopublication
+===============
+
```
dgarijo, line 3: What is the purpose of this file? It does not really add any information, can it be deleted?

dgarijo:
@ritamargherita,
I think that if you are going to be managing the root w3id, you don't need to add a readme file with you as a maintainer for each sub folder. We use the readme in the main folder to know who is the maintainer of the repo. Please have a look at my comment about the empty file and once addressed I will merge

odissei/np/README.md
ritamargherita: I have deleted the README files as recommended

ritamargherita:
Removed README files

dgarijo:
@ritamargherita,
looking at your redirects in detail, I think that the np redirect should be in the subfolder, because the `np` folder is for that redirection. In any case, I promised I would merge, so here it goes.

PR #2041, opened 2021-02-25
new namespace: linkml

cmungall:
We are currently splitting out the linkml namespace, which is currently a sub-namespace of biolink (see #1055) 

Note that we do not have github pages in place yet, but we thought it good to have the w3id redirects in place; apologies if this is premature

thanks cc @hsolbrig

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,7 +1,0 @@
-# w3id.org global rewrites
-
-Options +FollowSymLinks
-RewriteEngine on
-
-# redirect here vs in a UN directory due to case insensitive file systems issue
-RewriteRule ^UN(.*)$ /un$1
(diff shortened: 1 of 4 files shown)


davidlehn:
This removes the root .htaccess.  Please fix that.

davidlehn:
I'm unsure what's going on here.  You added `biolinkml/biolinkml/.htaccess`.  It looks it used to be `biolink/biolinkml/.htaccess`?  If you mean to have a new root `biolinkml/` dir, can you add comments or README with maintainer details?  Might also be good to update biolink with a note about the newer URLs.

cmungall:
Apologies for the confusion, I think it's cleaner if I close and start again.

cmungall:
This is replaced by #2068

PR #2051, opened 2021-03-04
feat(testlink): request w3id for testlink ontology

noelmcloughlin:
This PR introduces new namespace `testlink` which is variant of `csolink` namespace used for testing.

This PR also deletes unused directory under `csolink/` directory.


dgarijo:
@noelmcloughlin,
I am a little confused by this PR:
- You are removing the biolink w3id, but I am not sure this is intended.
- You state you are creating a w3id for <https://w3id.org/cocoon/v0.0.1#> but this is not correct.
- You seem to want to add a w3id for testlink-model/docs, but instead you add one for testlink/vocab

Can you pls clarify?

noelmcloughlin:
This `testlink` directory (namespace) is a derivative of existing `csolink`directory (namespace) so was copied and modified. While modifying I noticed that directory was existing but its actually not used in `csolink`or `testlink` namespaces so should be removed. The readme for "csolink" and "testlink" both reference cocoon  as generic example but I see that is confusing so I have added new commit to address that comment Thanks

testlink/README.md
```diff
@@ -0,0 +1,24 @@
...
+# Homepages
+* https://csolink.github.io/testlink-model/ -- testlink model
+* https://biolink.github.io/biolinkml/ -- Biolink modeling language
+
+# Docs
+* https://csolink.github.io/testlink-model/docs
+* https://biolink.github.io/biolinkml/docs
+
+# Vocabulary Usage
+    @prefix cocoon: <https://w3id.org/cocoon/v0.0.1#>
```
davidlehn, line 14:
```suggestion
    @prefix testlink: <https://w3id.org/testlink/v0.0.1#>
```

noelmcloughlin:
Thanks guys for good review.

testlink/README.md
```diff
@@ -0,0 +1,24 @@
...
+* https://biolink.github.io/biolinkml/docs
+
+# Vocabulary Usage
+    @prefix testlink: <https://w3id.org/testlink/v0.0.1#>
+
+# Contact
+* noel McLoughlin @gmail.com
+
+# Status
+* EXPERIMENTAL
```
davidlehn, line 20: It may be better to omit the status here (and in csolink/README.md).  I assume over time it's likely to be more up-to-date in the target docs and easier to update there.

noelmcloughlin: Good idea, done!!

PR #2060, opened 2021-03-11
Fix trailing slashes in /ga4gh .htaccess

andrewyatz:
Apologies for a fix coming in so quickly on yesterday's PR. A number of URLs had a trailing slash in their rewritten form. This was being interpreted by the rewrite code as an extra slash. We have also moved the capture to a 0 or many wildcard to allow for URLs without an extension to be used. The result of this was many of the URLs failed to resolve unless they included extra parts of a URL, which was not the intended result. URLs have been tested with https://htaccess.madewithlove.be/ in the hope this will reduce the likelihood for future fixes


andrewyatz:
Killing this and coming back for confirmation of errors and not to annoy yourselves

PR #2062, opened 2021-03-14
Update DID Spec links

OR13:
- [x] Update did spec link
- [x] Add warning about stale context
- [x] Add redirect test for did web
- [x] Add readme for section

### DIFF
--- a/did/.htaccess
+++ b/did/.htaccess
@@ -2,5 +2,13 @@ Header set Access-Control-Allow-Origin *
 Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
 Options +FollowSymLinks
 RewriteEngine on
-RewriteRule ^$ https://w3c-ccg.github.io/did-spec/ [R=302,L]
+RewriteRule ^$ https://www.w3.org/TR/did-core/ [R=302,L]
+
+# TODO: fix this once did core is published
+
 RewriteRule ^v(.+)$ https://w3c-ccg.github.io/did-spec/contexts/did-v$1.jsonld [R=302,L]
+
+
+# DID Web
+
+RewriteRule ^did:web:(.+)$ https://did-web.web.app/api/v1/identifiers/did:web:$1 [R=302,L]
--- a/did/.htaccess
+++ b/did/.htaccess
@@ -14,3 +14,9 @@ RewriteRule ^ns$ https://www.w3.org/ns/did/ [R=302,L]
 # https://w3id.org/did/v1 -> https://www.w3.org/ns/did/v1
 
 RewriteRule ^v(.+)$ https://www.w3.org/ns/did/v$1 [R=302,L]
+
+# Old CCG Redirects
+
+# https://w3id.org/did/v0.11 -> https://w3c-ccg.github.io/did-spec/contexts/did-v0.11.jsonld
+
+RewriteRule ^v0.11$ https://w3c-ccg.github.io/did-spec/contexts/did-v0.11.jsonld [R=302,L]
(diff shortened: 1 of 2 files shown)


dgarijo:
@OR13,
you are trying to edit a w3id which you did not create.
Can @dlongley approve this PR?

OR13:
Related option for did web in the future... https://github.com/w3c-ccg/did-method-web/issues/25

davidlehn:
See also: https://github.com/perma-id/w3id.org/pull/1653.
@msporny Would be nice to get your input on this too.

davidlehn:
@OR13 I'm unsure about that did:web change.
- It seems like perhaps it should just be it's own top level thing.  I'm guessing the idea was to prefer a path like /did/did:web:xxx vs /did:web/did:web:xxxx?
- What is the long term intent of that endpoint?  If it might end up as a primary resolver endpoint for a popular did method, there are operational considerations.  There haven't been other use cases like that on w3id.org before (IIRC).

OR13:
@davidlehn most to tests saving us from needing to call kubernetes to get a redirect.... currently the universal resolver supports redirects, but its very expensive to manage, would be nice for w3id.org to help resolve did from methods that register a resolver.

I could make another top level path for this like `r`... `https://w3id.org/r/did:web.....`

that would not use the did namespace, I just thought I would try to start with the existing namespace.

OR13:
@msporny @dlongley since can you please review?

I am happy to take my did resolver namespace idea to a shorter path like `r`.

OR13:
ping @msporny @dlongley @dmitrizagidulin

did/.htaccess
```diff
@@ -2,5 +2,13 @@ Header set Access-Control-Allow-Origin *
...
+RewriteRule ^$ https://www.w3.org/TR/did-core/ [R=302,L]
+
+# TODO: fix this once did core is published
+
 RewriteRule ^v(.+)$ https://w3c-ccg.github.io/did-spec/contexts/did-v$1.jsonld [R=302,L]
+
+
+# DID Web
+
+RewriteRule ^did:web:(.+)$ https://did-web.web.app/api/v1/identifiers/did:web:$1 [R=302,L]
```
msporny, line 14: Let's not mix DID Methods w/ the DID namespace on w3id.org -- can we separate this out to something like: `https://w3id.org/didweb/$1` ?

msporny (changes requested):
LGTM, except for conflation of did:web with DID spec namespace... we shouldn't cross those streams.

did/.htaccess
```diff
@@ -2,5 +2,13 @@ Header set Access-Control-Allow-Origin *
 Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
 Options +FollowSymLinks
 RewriteEngine on
-RewriteRule ^$ https://w3c-ccg.github.io/did-spec/ [R=302,L]
+RewriteRule ^$ https://www.w3.org/TR/did-core/ [R=302,L]
+
+# TODO: fix this once did core is published
+
 RewriteRule ^v(.+)$ https://w3c-ccg.github.io/did-spec/contexts/did-v$1.jsonld [R=302,L]
```
msporny, line 9: I know this isn't in this PR, but it seems wrong, no? We want to point to did-spec-registries now?

OR13:
@msporny  @dlongley @dmitrizagidulin 

please review again :)

dmitrizagidulin (approved):
:+1:

did/.htaccess
msporny: Yes, that's what I meant, thanks.

msporny (approved):

PR #2066, opened 2021-03-16
Digital Construction Ontologies - new .htaccess to support versioning and alignments

seppotorma:
Hi,

The .htaccess file is completely rewritten to support ontology versions and alignments. 

Also the old subdirectories and related .htaccess files have been removed.

I have tested the new redirects with the tool available at https://github.com/madewithlove/htaccess-cli. Rewrites work, but the tool unfortunately does not support the testing of access headers. So let's hope for the best. 

Best,
Seppo


seppotorma:
Obviously something went wrong. When I try to access the ontologies, I just get the error message below.

Is there any useful information in the error logs? I tried to access https://w3id.org/digitalconstruction around 13:50 CET today. 

Seppo

============================
Internal Server Error
The server encountered an internal error or misconfiguration and was unable to complete your request.

Please contact the server administrator at support@w3id.org to inform them of the time this error occurred, and the actions you performed just before this error.

More information about this error may be available in the server error log.

Apache/2.4.29 (Ubuntu) Server at w3id.org Port 443

dgarijo:
@seppotorma,
I think you haven't added the root rule. For example:
RewriteRule ^digitalconstruction$   would redirect to w3id.org/digitalconstruction/digitalconstruction.
Maybe the intention there is just `^$`

seppotorma:
Ok. If that rewrite is what the logs show, then the testing tool that I have been using (https://github.com/madewithlove/htaccess-cli) works in a different way. 

I need to study this a bit. Perhaps the easiest way is to install apache locally to test the redirects. It seems w3id.org is using apache 2.4.29.

Seppo

dgarijo:
@seppotorma,
i have not looked at the logs, as I don't have access. I talk from experience and from what I see in your htaccess.
When I do tests, I run a tomcat locally to test my redirections

davidlehn:
@seppotorma Errors look like this:
```
[Tue Mar 16 12:49:40.530808 2021] [core:alert] [pid X] [client X.X.X.X:X] /X/w3id.org/digitalconstruction/.htaccess: RewriteCond: unknown flag 'AND'
```
And yes the rules are relative to current dir so don't want that extra `digitalconstruction` prefix.

PR #2075, opened 2021-03-19
Creating a permanent identifier for the bop ontology

AlexDonkers:
Dear maintainer, 

Could you please create a permanent identifier for the bop ontology?
I hope I provided you with the correct documents via this way. 

Project: Building Performance Ontology
Prefix: bop
Namespace: https://w3id.org/bop

Best, Alex
a.j.a.donkers@tue.nl

### DIFF
--- a/bop/.htaccess
+++ b/bop/.htaccess
@@ -1,0 +1,47 @@
+# Turn off MultiViews
+Options -MultiViews
+
+# Directive to ensure *.rdf files served as appropriate content type,
+# if not present in main apache config
+AddType application/rdf+xml .rdf
+AddType application/rdf+xml .owl
+AddType text/turtle .ttl
+AddType application/n-triples .n3
+AddType application/ld+json .json
+# Rewrite engine setup
+RewriteEngine On
+#Change the path to the folder here
+RewriteBase /myDocumentation 
+
+# Rewrite rule to serve HTML content from the vocabulary URI if requested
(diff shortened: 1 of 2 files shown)


dgarijo:
@AlexDonkers,
you are copying directly the htaccess by Widoco, which is not correct because we don't host the files in here.
You should host the files in your target GitHUb repository and then copy the URL in the htaccess. 
See the example I did here: https://github.com/perma-id/w3id.org/blob/master/example/.htaccess for reference.

AlexDonkers:
@dgarijo Thank you for your comment. Excuse me, I didn't check the links. They are updated now.

dgarijo:
@AlexDonkers,
please remove the `rewritebase` line, because the `myDocumentation` folder does not exist.

bop/.htaccess
```diff
@@ -0,0 +1,47 @@
...
+# if not present in main apache config
+AddType application/rdf+xml .rdf
+AddType application/rdf+xml .owl
+AddType text/turtle .ttl
+AddType application/n-triples .n3
+AddType application/ld+json .json
+# Rewrite engine setup
+RewriteEngine On
+#Change the path to the folder here
+RewriteBase /myDocumentation
```
dgarijo, line 14: Remove line 14

AlexDonkers: @dgarijo Deleted line 14.

PR #2086, opened 2021-03-31
Basic redirects for ereefs definitions 

dr-shorthair:
with the URI https://w3id.org/ereefs/def/{term}


dgarijo:
307 is for temporary redirects, I recommend using 303 (see other)

dr-shorthair:
Thx @dgarijo - will try to stabilize on this.

PR #2096, opened 2021-04-07
Added .htaccess and README files

cristinacm:


dgarijo:
Just so you know @cristinacm, we just need the reamde for the root folder (in this case, todo). Thanks!

PR #2106, opened 2021-04-12
Updated routing for certification schema

crossoveranx:


bloxberg/.htaccess
```diff
@@ -2,6 +2,7 @@ Header set Access-Control-Allow-Origin *
 Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
 Options +FollowSymLinks
 RewriteEngine on
+RewriteRule ^(bloxberg/schema/research_object_certificate_v1)$ https://certify.bloxberg.org/research_object_certificate_v1 [R=301,L,NC]
```
dgarijo, line 5: This doesn't look like the id you want. You are creating w3id.org/bloxberg/bloxberg/schema/..
Instead, consider: `RewriteRule ^schema/research_object_certificate_v1$`

crossoveranx: Thank you for the suggestion, I will revise and create a new PR. Any suggestions on testing locally before committing would also be appreciated

dgarijo: You can test it by setting an Apache server and then placing your test folder in the `/var/www/html/` folder.

bloxberg/.htaccess
```diff
@@ -2,6 +2,7 @@ Header set Access-Control-Allow-Origin *
 Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
 Options +FollowSymLinks
 RewriteEngine on
+RewriteRule ^schema/research_object_certificate_v1$ https://certify.bloxberg.org/research_object_certificate_v1 [R=301,L,NC]
 RewriteRule ^$ https://www.bloxberg.org/ [R=302,L]
 RewriteRule ^research_object_certificate_v1$ https://certify.bloxberg.org/research_object_certificate_v1 [R=302,L]
-RewriteRule ^(.*)$ https://www.bloxberg.org/$1 [R=302,L]
+RewriteRule ^(.*)$ https://bloxberg.org/ [R=302,L]
```
davidlehn, line 8: Should the above www.bloxberg.org and this bloxberg.org be synced to be the same?

crossoveranx: Yes, they should be synced and redirect to the same. We have a redirect on our end from www.bloxberg.org -> bloxberg.org

bloxberg/.htaccess
crossoveranx: Okay, I believe it is working as intended. Added the .htaccess file in this commit to the `/var/www/html/` folder as suggested, and I receive a redirect when visiting the `/schema/research_object_certificate_v1`. 

However, I didn't have success adding the whole bloxberg folder to the path in order to create the redirects for testing. I assume this folder is prepended to the route when querying (I normally use nginx). If so, then it should work as intended.

PR #2108, opened 2021-04-12
Add /security/suites/secp256k1recovery-2020

clehner:
As discussed here: https://github.com/w3c-ccg/security-vocab/pull/95

### DIFF
--- a/security/.htaccess
+++ b/security/.htaccess
@@ -37,3 +37,5 @@ RewriteRule ^suites/ed25519-2020$ https://digitalbazaar.github.io/ed25519-signat
 RewriteRule ^suites/ed25519-2020/v1$ https://digitalbazaar.github.io/ed25519-signature-2020-context/contexts/ed25519-signature-2020-v1.jsonld [R=302,L]
 RewriteRule ^suites/x25519-2020$ https://digitalbazaar.github.io/x25519-key-agreement-2020-context [R=302,L]
 RewriteRule ^suites/x25519-2020/v1$ https://digitalbazaar.github.io/x25519-key-agreement-2020-context/contexts/x25519-key-agreement-2020-v1.jsonld [R=302,L]
+RewriteRule ^suites/secp256k1recovery-2020$ https://identity.foundation/EcdsaSecp256k1RecoverySignature2020/ [R=302,L]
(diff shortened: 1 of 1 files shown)


dgarijo:
@clehner,
 sorry, you are not listed as a maintainer of this w3id. Please have: @dlongley, @davidlehn, @msporny or @OR13 approve this request

OR13:
Looks good to me!

davidlehn:
I have a concern about the direction of this `/security` endpoint.  It's turning into a bit more of a kitchen sink than most endpoints.  And tracking who can update which sub-rules is going to be a bit more challenging than usual.

This context is defining things off in it's own namespace.  The other contexts here are (I think) all using `https://w3id.org/security#` as a base.  This new one is using `https://identity.foundation/EcdsaSecp256k1RecoverySignature2020#`and redefining properties such as `publicKeyJwk`, `privateKeyJwk`, and others with this new base vs how it's defined in the other jws context.  Seems like this may get confusing or just outright start conflicting if you use both contexts at once.

msporny:
> I have a concern about the direction of this `/security` endpoint. It's turning into a bit more of a kitchen sink than most endpoints. And tracking who can update which sub-rules is going to be a bit more challenging than usual.

To address this concern, we should split this single .htaccess file into multiple directories containing .htaccess files.

> This context is defining things off in it's own namespace.

Yes, the `EcdsaSecp256k1RecoverySignature2020` JSON-LD Context is broken... it'll only blow up on people using `EcdsaSecp256k1RecoverySignature2020`... which is a good thing, and exactly the sort of behavior we want (so this sort of thing doesn't take the rest of the ecosystem down with them).

@clehner, you may want to follow this example instead:

https://raw.githubusercontent.com/digitalbazaar/ed25519-signature-2020-context/master/contexts/ed25519-signature-2020-v1.jsonld

You may also want to avoid providing terms to express private key values in public contexts... private key values shouldn't escape your systems.

msporny:
The registration is fine, however... we should let this through and refactor later. 

@clehner -- please note the ticking time bomb that is your JSON-LD Context. This repo doesn't care about that, but you should (as an implementer). :)

clehner:
Thanks @msporny. Noted:
https://github.com/decentralized-identity/EcdsaSecp256k1RecoverySignature2020/issues/14
https://github.com/decentralized-identity/EcdsaSecp256k1RecoverySignature2020/issues/15

PR #2114, opened 2021-04-16
[idpo] Added idpo ontology

philhag:


idpo/.htaccess
```diff
@@ -0,0 +1,52 @@
...
+# In case of accept header <application/rdf+xml>
+RewriteCond %{HTTP_ACCEPT} application/rdf\+xml
+RewriteRule ^$ https://icdd.vm.rub.de/ontology/idpo/idpo.rdf [R=302,NE,L]
+
+# In case of accept header <application/ld+json>
+RewriteCond %{HTTP_ACCEPT} application/ld+json
+RewriteRule ^$ https://icdd.vm.rub.de/ontology/idpo/idpo.jsonld [R=302,NE,L]
+
+# If suffix ttl, redirect to turtle version
+RewriteRule ^idpo.ttl$ https://icdd.vm.rub.de/ontology/idpo/idpo.ttl [R=302,NE,L]
```
dgarijo, line 30: This will be https://w3id.org/idpo/idpo.ttl Is this what you intend?

dgarijo:
Please see my comment above. Also, 303 redirect is recommended for semantic resources instead of 302 (I see you used it only in the turtle)

idpo/.htaccess
philhag: Right, its intented to serve the turtle file either by requesting using the <text/turtle> header on the base path or specifying the ttl file directly, e.g for weblinks to that file

philhag:
I changed response code to 303 for ontology resources.

PR #2118, opened 2021-04-19
UnderRepresentedWriters Project

marcostranisci:
Permanent Identifier request for the Under-Represented Writers Project


dgarijo:
@marcostranisci,
sorry, we don't host ontologies in this repo. This is only for the redirections. See https://github.com/perma-id/w3id.org/tree/master/example for an example redirection

marcostranisci:
You're right, sorry! I removed the Ontology and left only the url, and my mail as a mantainer

dgarijo:
But now you don't have an htaccess, so you are not redirecting to anything... Please have a look at the example I pointed to for a reference.

marcostranisci:
You're right again. I chose to follow the examples from generic Github pages since the ontology is only a part of my whole project. Thanks for your help

davidlehn:
@dgarijo This should have been squashed to avoid the 6MB 135000 line file in git history.  I'd like to be evil and force push a fix for that, but master branch is protected at the moment and I'm not an admin.  I'm requesting assistance on that.  Trying to do it before other PRs are merged.

@marcostranisci If a squash of your commits is pushed, you'll need to re-checkout before making new changes.

dgarijo:
Apologies, @davidlehn

PR #2137, opened 2021-04-28
Update .htaccess for "faircookbook"

robertgiessmann:
updating the "faircookbook" project .htaccess

sorry to have forgot the project name

link to project readme:  https://github.com/perma-id/w3id.org/blob/master/faircookbook/README.md

### DIFF
--- a/faircookbook/.htaccess
+++ b/faircookbook/.htaccess
@@ -2,7 +2,14 @@ RewriteEngine On
 
 # define individual identifiers for recipes, manually:
 RewriteRule "^FCB000$" https://fairplus.github.io/the-fair-cookbook/content/home.html [R=303,L,QSA]
-RewriteRule "^FCB001$" https://fairplus.github.io/the-fair-cookbook/content/recipes/findability/identifiers.html [R=303,L,QSA]
+RewriteRule ^FCB001$ https://fairplus.github.io/the-fair-cookbook/content/recipes/findability/identifiers.html [R=303,L,QSA]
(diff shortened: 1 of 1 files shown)


dgarijo (approved):

dgarijo:
Why use 303 if these are regular redirects? I would use 301 (or 302 if they are temporal)

robertgiessmann:
Hi @dgarijo , true, 301 would be a possibility. 

We want to use w3id.org as an "identifier minter" for FAIR data, and I have seen "303 See Other" as the "more appropriate" response code. Still no true consensus on that, but I like the "See Other" instead of "301 Moved Permanently". 

We use <http://w3id.org/faircookbook/FCB001> to speak about the recipe. Seeing you're from Ontology Engineering at UPM, I guess you know what I mean?


Thanks and best! 
   Robert

dgarijo:
Yes, I understand. Usually 303 has been agreed to refer to the semantic web descriptions for content negotiation, e.g., you get the OWL  or HTML redirection depending on what you are interested in. Since you only have HTML redirects, I still think a 301 makes more sense. But if you want to mint w3ids for datasets and handle content negotiation (e.g., return an HTML representation of the dataset vs a semantic description of its contents), then 303 redirects would be appropriate. In case you are interested, we wrote a book chapter for publishing FAIR ontologies a while ago:[PDF](https://dgarijo.com/papers/best_practices2020.pdf) [DOI](https://doi.org/10.3233/SSW200034) It overlaps partially with some of your guidelines :)

robertgiessmann:
ah, I didn't know that this was you writing the article! Glad ot meet you here! :) 

We plan to have content negotiation in the future, so we wanted to start like this already. Thanks for commenting!

dgarijo:
no prob!

PR #2151, opened 2021-05-06
Added a new redirect rule

huaning-yang:


iqb/.htaccess
```diff
@@ -4,3 +4,4 @@ RewriteRule ^core(.*)? https://github.com/iqb-berlin/mdc-core [R=302,L]
 RewriteRule ^mdc-core/(.*)? https://iqb-berlin.github.io/mdc-core/w3id.org/iqb/mdc-core/$1 [R=302,L] 
 RewriteRule ^github(.*)? https://iqb-berlin.github.io/ [R=302,L]
 RewriteRule ^(.*) https://www.iqb.hu-berlin.de/ [R=302,L]
+RewriteRule ^verona-modules(.*)? https://raw.githubusercontent.com/verona-interfaces/metadata/master/verona_modules_vocab.jsonld [R=302,L]
```
davidlehn, line 7: Move this up a line to be above the wildcard rule.  Otherwise it will never be used.

huaning-yang: Fixed it, thank you very much!

PR #2153, opened 2021-05-08
Add additional suites

OR13:
This PR fixes the registration style for some already registered suite to match the IRIs added to did core test suite, and also registers some new suites.

### DIFF
--- a/security/.htaccess
+++ b/security/.htaccess
@@ -5,29 +5,25 @@ RewriteEngine on
 RewriteRule ^$ https://w3c-ccg.github.io/security-vocab/ [R=302,L]
 RewriteRule ^v1$ https://w3c-ccg.github.io/security-vocab/contexts/security-v1.jsonld [R=302,L]
 RewriteRule ^v2$ https://w3c-ccg.github.io/security-vocab/contexts/security-v2.jsonld [R=302,L]
+
+# TODO: make v3 "stable"
+
 RewriteRule ^v3-unstable$ https://w3c-ccg.github.io/security-vocab/contexts/security-v3-unstable.jsonld [R=302,L]
 
-# https://w3id.org/security/jws
+# TODO: remove "suites" not rooted under  "suites"
 
 RewriteRule ^jws$ https://w3c-ccg.github.io/lds-jws2020/ [R=302,L]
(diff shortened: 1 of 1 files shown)


security/.htaccess
```diff
@@ -5,30 +5,26 @@ RewriteEngine on
 RewriteRule ^$ https://w3c-ccg.github.io/security-vocab/ [R=302,L]
 RewriteRule ^v1$ https://w3c-ccg.github.io/security-vocab/contexts/security-v1.jsonld [R=302,L]
 RewriteRule ^v2$ https://w3c-ccg.github.io/security-vocab/contexts/security-v2.jsonld [R=302,L]
+
+# TODO: make v3 "stable"
+
 RewriteRule ^v3-unstable$ https://w3c-ccg.github.io/security-vocab/contexts/security-v3-unstable.jsonld [R=302,L]
 
-# https://w3id.org/security/jws
+# TODO: remove "suites" not rooted under  "suites"
```
OR13, line 13: @tplooker @dlongley @dmitrizagidulin 

we probably ought to just bite the bullet and kill these older URIs asap... the longer we wait, the madder everyone will be.

dmitrizagidulin: Yeah that's a good point...

peacekeeper:
Looks good, but could you expain how the individual suite JSON-LD contexts relate to the [Security Vocabulary](https://github.com/w3c-ccg/security-vocab)?

For example, `JsonWebSignature2020` is defined in both:

* https://w3c-ccg.github.io/lds-jws2020/contexts/v1/ (added by this PR), AND
* https://github.com/w3c-ccg/security-vocab/blob/master/contexts/security-v3-unstable.jsonld.

OR13:
@peacekeeper previously, security vocab and credentials context bundled suite types... this lead to problems upgrading them.... sec-v3 it probably going to end up looking more like did core, and suite types are likely to be pulled on an as needed basis... at least that is the direction things seem to be headed.

But its also possible that sec-v3 is will be a kind of kitchen sink, containing recommended suites... sorta like sec-v2...

its not really clear what it will be right now TBH.

peacekeeper (approved):
Very useful update

msporny (approved):

security/.htaccess
OR13: Related: https://github.com/w3c-ccg/security-vocab/issues/108

PR #2161, opened 2021-05-15
Add security/suites/x25519-2019

OR13:
This adds x25519 2019 terms for LD Proofs


davidlehn:
There is a x25519-2020 alias too.  Looks like difference is the year and base58 vs multibase.  Is there a need for this older one as well?  Also maybe the x25519* entries should be sorted next to each other.

OR13:
yes, this registration is required to support backwards compatibility... IIRC the original suite name was "X25519VerificationKey2018" but it was not registered anywhere until 2019.... perhaps we should back date it to align with ed25519... @msporny thoughts?

dmitrizagidulin (approved):
:+1: 
Thanks @OR13!

PR #2165, opened 2021-05-17
seems to need more complex changes.

markwilkinson:


davidlehn:
Please try to use more descriptive commit messages rather than a dialog.  The "Sorry Daniel! ..." message and similar will be strange to future readers of git history.  Better to say "Update duchenne-fdp redirects." for commits and leave the commentary to the PR title and comments.

PR #2174, opened 2021-05-20
Add multikey 2021

OR13:
This PR is sadly necessary to handle cases where there is no JSON-LD support for standard NIST curves like P256.

I did my best not to trigger you @msporny but we can't possible support did:key without this, hopefully its acceptable.

My sense is that the LD Sec WG Suite imp guide will have a large section on best practices in this regard, and we can elaborate on the pro's and con's of `type` there.

### DIFF
--- a/security/.htaccess
+++ b/security/.htaccess
@@ -39,10 +39,10 @@ RewriteRule ^suites/hmac-2019$ https://digitalbazaar.github.io/sha256-hmac-key-2
 RewriteRule ^suites/hmac-2019/v1$ https://digitalbazaar.github.io/sha256-hmac-key-2019-context/contexts/sha256-hmac-2019-v1.jsonld [R=302,L]
 
 
-# http://w3id.org/security/suites/multibase-2021
+# http://w3id.org/security/suites/multikey-2021
 
-RewriteRule ^suites/multibase-2021$ https://ns.did.ai/suites/multibase-2021 [R=302,L]
-RewriteRule ^suites/multibase-2021/v1$ https://ns.did.ai/suites/multibase-2021/v1 [R=302,L]
+RewriteRule ^suites/multikey-2021$ https://ns.did.ai/suites/multikey-2021 [R=302,L]
+RewriteRule ^suites/multikey-2021/v1$ https://ns.did.ai/suites/multikey-2021/v1 [R=302,L]
 
 # http://w3id.org/security/suites/secp256k1recovery-2020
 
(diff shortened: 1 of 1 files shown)


davidlehn:
Should this be merged?  This is ok as far as w3id procedures are concerned.  @msporny was there some issue that needed to be resolved first?  I'm not sure how this is being used, but the name "multibase-2021" seems off since the context looks to be a collection of key and signature types, not various base encodings.  (Is there somewhere better to discuss technical details of this?)

security/.htaccess
```diff
@@ -38,6 +38,12 @@ RewriteRule ^suites/aes-2019/v1$ https://digitalbazaar.github.io/aes-key-wrappin
 RewriteRule ^suites/hmac-2019$ https://digitalbazaar.github.io/sha256-hmac-key-2019-context [R=302,L]
 RewriteRule ^suites/hmac-2019/v1$ https://digitalbazaar.github.io/sha256-hmac-key-2019-context/contexts/sha256-hmac-2019-v1.jsonld [R=302,L]
 
+
+# http://w3id.org/security/suites/multibase-2021
+
+RewriteRule ^suites/multibase-2021$ https://ns.did.ai/suites/multibase-2021 [R=302,L]
+RewriteRule ^suites/multibase-2021/v1$ https://ns.did.ai/suites/multibase-2021/v1 [R=302,L]
```
msporny, line 45: Multibase isn't a cryptography suite anymore than base64 is. The name of the suite is misleading... at the very least, change it to "multikey"... but do that knowing that you're going to have to resolve this with the IPFS community... who have seemed to abandon the concept of multikey a while ago: https://github.com/ipfs/specs/issues/58

You're resurrecting something that's been dead for a while now... I have no issue with you doing that, except that I hope you will do all the due diligence necessary to build bridges between communities when you do this.

Again, I completely disagree with trying to bring the brokenness of JWK into the LD Security community, but it's not illegal to do so.

Please rename to multikey-2021 and I won't stand in the way.

msporny:
> This PR is sadly necessary to handle cases where there is no JSON-LD support for standard NIST curves like P256.

Isn't the only thing necessary there is to create an `EcdsaSignature2021` suite (or similar) and a `Secp256r1VerificationKey2021` (or similar)? We don't have to support all variations, do we?

> we can't possible support did:key without this, hopefully its acceptable.

I don't understand this statement... isn't this conflating multibase/multicodec encoding w/ cryptosuite naming?

OR13:
@msporny I took your proposed name change, the issue is a general one, for all these keys: https://www.iana.org/assignments/jose/jose.xhtml#web-key-elliptic-curve
which we support with JWK, we want to support did:key as well.

You are right that I could create a linked  data suite per key type, but thats not how JOSE really works, and it's also not how JsonWebKey2020 works.

IMO we need to show that all variations can be supported, but we don't need to actually use all of them.

This is part of showing that we are improving things, its hard to argue that a subset of IANA / JWK is better than a superset, especially when it doesn't cover NIST / FIPs crypto.

OR13:
Implemented the needed change on our side: https://github.com/transmute-industries/ns.did.ai/commit/43a5cd31a3b89935ca67d07c3b8a199ec2bc8a8d

This PR is blocking our ability to register our did key implementation in the did core test suite, and our ability to update the dicd core examples.

OR13:
@msporny this PR is blocking our did test suite registration.

OR13:
@davidlehn @msporny eta on merge for this? or are there more changes requested?

msporny (approved):

PR #2175, opened 2021-05-20
Remove all security suites not under /security/suites

OR13:
@msporny @dlongley @tplooker @dmitrizagidulin 

This PR removes any chance that our non suite "suite" registries will continue to survive... I think its worth it, but expect it might break a few things.

### DIFF
--- a/security/.htaccess
+++ b/security/.htaccess
@@ -10,17 +10,6 @@ RewriteRule ^v2$ https://w3c-ccg.github.io/security-vocab/contexts/security-v2.j
 
 RewriteRule ^v3-unstable$ https://w3c-ccg.github.io/security-vocab/contexts/security-v3-unstable.jsonld [R=302,L]
 
-# TODO: remove "suites" not rooted under  "suites"
-
-RewriteRule ^jws$ https://w3c-ccg.github.io/lds-jws2020/ [R=302,L]
-RewriteRule ^jws/v1$ https://w3c-ccg.github.io/lds-jws2020/contexts/v1/ [R=302,L]
-RewriteRule ^pgp$ https://or13.github.io/lds-pgp2021/ [R=302,L]
-RewriteRule ^pgp/v1$ https://or13.github.io/lds-pgp2021/contexts/pgp-v1.jsonld [R=302,L]
-RewriteRule ^bbs$ https://github.com/w3c-ccg/ldp-bbs2020/ [R=302,L]
-RewriteRule ^bbs/v1$ https://w3c-ccg.github.io/ldp-bbs2020/contexts/v1/ [R=302,L]
-RewriteRule ^blockchain$ https://or13.github.io/lds-blockchain2021/ [R=302,L]
-RewriteRule ^blockchain/v1$ https://or13.github.io/lds-blockchain2021/contexts/blockchain-v1.jsonld [R=302,L]
-
 # https://w3id.org/security/suites
 
 # TODO: update registry to reflect this file.


davidlehn:
I'm not sure how to deal with this.  We really don't want to break code or data where people may have been using these. 
 Maybe these should be clearly marked deprecated for a few years(?) first?

OR13:
did core just changed the entire did context... I feel like we can remove these IRIs. but you are right, if these we used by much more than just a few people... I suspect we would be a worse spot.

without access to the logs I can't tell how many times they are accessed, we always cache these things, so we wouldn't even notice for a while...

davidlehn:
Long term logs are not kept, so only data is for last ~2 weeks.  Blind grepping for 'security/X' yeilds: jws=17, pgp=0, bbs=23321, blockchain=0.  There are requesting IPs, but most requests on that server just show generic browser, python, or go clients, so hard to tell what is using the redirects.  And apparently with clusters of the same bbs request every second a thousand times they are systems that are very bad at caching!

It would be nice to use some headers to signal deprecated endpoints, but I imagine getting everyone to look for those would be difficult.  So I'm not sure what the solution is for these situations other than to at a minimum add clear comments saying not to use certain rules.  And at some point in the far future when risk of damage is low, remove them.

OR13:
wow someone using `bbs` has no idea how to securely use JSON-LD....

jws=17 could have been me alone, as I started addressing the fact that did core context has changed.

we could start by removing them 1 by 1 in order to least recently used...

It almost feels appropriate to remove bbs so those people learn how to use a cache.

An even nicer way of educating them would be to return a json warning about the context being deprecated, randomly :)

OR13:
there should a flag in JSON-LD that throws an error when you load a context over the network in production.

alenhorvat:
In addition, when a long-term credential is issued the credential owner should store both the credential and the @context. 
Should the context hash or exact context version be included in the JSON-LD?

OR13:
@alenhorvat excellent question, anyone using JSON-LD should assume the context will change over time (in newer version), but the version bound to the credential should not (once a stable version is achieved).

I client needing to store a context should probably store it like this:

```
iri: string,
digest: string,
created: string,
context: string
```

In our systems today, we use npm to store contexts we use, so `created` and `digest` are stored in both git and npm for a given IRI version, for example:

https://github.com/transmute-industries/verifiable-data/commits/main/packages/did-context/contexts/did-v1.json

^ you can see that did core is not stable yet.

If you are storing contexts outside of a package manager, I would be careful, changes in the context can make signatures on older credentials invalid.

alenhorvat:
Yes, this makes sense. 
However, would including the digest directly into the JSON-LD (VC/VP) help?

Two other points to keep in mind:
1. VCs are shared, hence either the context availability must be ensured or the context should be shared by the holder
2. Offline VC validation (e.g., VC can be shared via protocols like Bluetooth, NFC, ...) and they should be verifiable offline (of course validation is limited as one cannot check if the credential is revoked, ...)
like https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity or @context : "<URL>?<context-digest>"

What do you think?

OR13:
@alenhorvat you can't include the hash of a thing with a signature in the thing with the signature, because the signature does not exist in the thing, so the hash cannot exist....

You can swap out the context IRIs in a VC, with new ones, and as long as no terms are changed, the signatures will still validate.

Thats what you care most about, not "has this context added new terms since the vc was signed".

Adding unrelated semantics to a context should not invalidate credentials produced from it, adding integrity protection for the context to the credential would invalidate credentials if checking that integrity were part of verification, but it's not.

See https://w3c.github.io/vc-data-model/#data-schemas

I would personally not try and use JSON-LD for schemas... use a schema language for that.

the credential is a commitment to the information representation of the credential at the time of issuance by the issuer.

If the issuer changes that, or allows it to change by using unstable dependencies, they are to blame, and you can tell if thats the case if the credential fails to verify because a context shifted.

Don't use that issuer any more.

If this happens enough, nobody will trust JSON-LD VCs and it will solve itself by market forces... :)

alenhorvat:
>  @alenhorvat you can't include the hash of a thing with a signature in the thing with the signature, because the signature does not exist in the thing, so the hash cannot exist....
Of course, you cannot :) I didn't express myself well.

You have a VC in JSON-LD format.
JSON-LD contains a link (or several links) to JSON-LD context. There's no guarantee that the source will be always available. (For example, a month or so ago somebody made a breaking change on the DID document context and updated the new version under the same V1.)
When you want to validate the signature you:
1. obtain the context
2. canonicalize the JSON-LD (take the context, expand all the properties)
3. hash + sign the hash
My point is, if the context is changed for whatever reason, step 2 may result in a different document than the one that was signed.
If the hash of the context is included in the JSON-LD, before the JSON-LD is signed (this of course can be done), one prevents that the wrong context is used for canonicalization.

Of course, adding unrelated semantics won't change the result, however changing existing fields, it would.

Correct, JSON-LD is not used as a schema, the context used by JSON-LD is.

I don't expect the issuer will change the schema, but an attacker would. Let's take an example, where covid credentials are issued as VCs in a JSON-LD format and the context points to an endpoint. If an attacker manages to take down/block/... the endpoint, all credentials suddenly become invalid.
This is why I'm asking if:
a) schema hash should be also included in the JSON-LD?
b) it should be possible to share the schema along with the VC? (this is of course always possible)
(Using another format is also an option)

OR13:
> a) schema hash should be also included in the JSON-LD?

Certainly an option, but an integrity check fail doesn't necessarily mean any of the terms that were signed were changed... were as a signature check failure already tells you this.

> b) it should be possible to share the schema along with the VC? (this is of course always possible)

yes, and ideally, you are not issuing VCs from contexts that are changing... if that ever happens, the issuer is to blame, in the same way they would be to blame for picking a DID Method that broke every 5 seconds  :) 

> I don't expect the issuer will change the schema, but an attacker would. 

From my point of view you don't get anything from embedding the context hash in the VC... because you are already assured of detecting a term change by the LD Proof, and a term addition to the context should not break the schema.

think of `https://example.com/industry/context/v1` as a pointer to `@vendor/industry-context` on npm.

The only way terms under this IRI would change is if `@vendor` updated their package... if they did this outside a major versin release and they broke stuff, don't use them anymore... now replace `vendor` with W3C or Microsoft of Epic or Apple.

The fact that the contexts can be swapped out and the terms remain unchanged is a feature.... it's like being able to publish a non breaking change to a software package...

If you bump a patch and suddenly a breaking change destroys everything, you will likely stop relying on that software provider... in the same way, if the W3C or Microsoft or Epic or Apple decided to push a breaking change outside of a major version, people would rightly argue that they violated the expectations folks had form them as a software provider.

TLDR,

- never trust anyone to map IRIs to terms other than software you control.
- never load a context over the network outside of a testing / package publishing integrity checking use case.
- never rely on software packages you don't trust.
- if you let your users decide which software packaged they trust (like Google and Microsoft do, but Apple does not)... then make it clear t them what allowing those custom contexts / credential format might mean... 

are you sure you really want to install this app -> are you sure you really want to accept this credential?

Giving users the choice to control their own devices and software means given them the choice to disagree with the manufacturer or os vendor.... If the device or os vendor thinks they know better than the user, and wants to create an experience where only approved contexts and credentials can exist, that is their right, and the market will decide if they survive with that approach.

alenhorvat:
I agree. What do you think if the publish the recommendations somewhere (e.g., good practice):

- never trust anyone to map IRIs to terms other than the software you control.
- never load a context over the network outside of a testing / package publishing integrity checking use case.
- never rely on software packages you don't trust.
- if you let your users decide which software packaged they trust (like Google and Microsoft do, but Apple does not)... then make it clear t them what allowing those custom contexts / credential format might mean...

tplooker:
> Long term logs are not kept, so only data is for last ~2 weeks. Blind grepping for 'security/X' yeilds: jws=17, pgp=0, bbs=23321, blockchain=0

Yikes apologies about that I'd say the [jsonld-signatures-bbs test suite](https://github.com/mattrglobal/jsonld-signatures-bbs) is to blame for the majority of these, slipped through when I was updating the context, its now rectified.

Agree with @davidlehn here though, dont agree with simply removing these and breaking things that are working big -1.

tplooker (changes requested):
Just putting this here to make my objection clearer, please do not merge this it will breaking things and is un-necessary.

OR13:
@tplooker this would only break things if you don't cache contexts.... are you saying we can't clean this up because developers don't follow best practices?

Do we leave these 1-offs in here forever?

How can we make it clear that they should not be used?

tplooker:
> are you saying we can't clean this up because developers don't follow best practices?

Following that logic, why would any remote JSON-LD context ever be resolvable? Why not use URN's for all contexts? 

> How can we make it clear that they should not be used?

I think the best tool we have here is education, teaching developers 'why' rather than just breaking things they have come to rely on, rightly or wrongly.

OR13:
@tplooker so you would be in favor of moving these IRIs to the bottom to the block, and surrounding them with warnings, but not removing them.... I can live with that... @msporny @dlongley  thoughts?

davidlehn:
> there should a flag in JSON-LD that throws an error when you load a context over the network in production.

The theory is that a custom document loaders would throw errors for unknown contexts if needed.


> @tplooker this would only break things if you don't cache contexts.... are you saying we can't clean this up because developers don't follow best practices?
> 
> Do we leave these 1-offs in here forever?
> 
> How can we make it clear that they should not be used?

One of the main intents of w3id was to support ids that can last for long periods of time.  More towards decades than days.  Of course things change and people change their minds and projects move on.  The hope was that the redirects could be updated as long as needed to keep old consumers of those endpoints working.  Churn of "1-offs" is a problem when thinking long term.  The w3id best practices are not really documented but would include only minting ids that you intend to use and support for a long time.  No one can force such that though.


> @tplooker so you would be in favor of moving these IRIs to the bottom to the block, and surrounding them with warnings, but not removing them.... I can live with that... @msporny @dlongley thoughts?

I think something like that would work for deprecated ids.  For easier tracking than "git blame" I'd add a date in there too.  And suggestions on what to change to if appropriate.  Hopefully any remaining users would find those notes at some point before the endpoints change or go away and break apps.  If the belief is that there really are zero users of some ids, then this is a lot of trouble and maybe they are better just removed.  That's a judgement call.


> [... context integrity, caching, and related issues ...]

Issues related to context integrity and caching got a bit off topic here.  Such things have come up often in the JSON-LD CG and WG.  Might get more input on the topic in a JSON-LD forum.  I thought we had put some text in the JSON-LD best practices about such things but I'm not sure where it is.  The [hashlink](https://datatracker.ietf.org/doc/html/draft-sporny-hashlink) spec is one attempt at trying to handle some of these issues, perhaps requiring a special document loader.  I think it was too much to add support to JSON-LD 1.1 to handle such things but maybe a future spec will more directly tackle the issues involved.

OR13:
I am going to close this PR and open a new one, that cleanly deprecates the "suites" that are not under "suites"

OR13:
new PR is open here: https://github.com/perma-id/w3id.org/pull/2191

PR #2176, opened 2021-05-21
PUV ontology

dr-shorthair:
@nicholascar maybe look at this one for me?


nicholascar:
I've removed the leading slashes in the patterns ("/puv...") as these are only needed when using apache config files, not .htaccess files.

PR #2179, opened 2021-05-22
Add pattern for ontology elements

dr-shorthair:


env/.htaccess
```diff
@@ -2,16 +2,17 @@ Options +FollowSymLinks
 RewriteEngine on
 
 # PUV ontology
+# match 'puv' followed by optional #{name} or /{name}
 # Turtle representation
 RewriteCond %{QUERY_STRING} ^(_format|_mediatype)=text/turtle$
-RewriteRule ^puv$ https://rawcdn.githack.com/CSIRO-enviro-informatics/PUV-ont/dd2d8762f80a58e618269593e99d3f840de0f150/rdf/puv.ttl [R=303,L]
+RewriteRule ^puv([\/#]\S+)?$ https://rawcdn.githack.com/CSIRO-enviro-informatics/PUV-ont/dd2d8762f80a58e618269593e99d3f840de0f150/rdf/puv.ttl [R=303,L]
```
davidlehn, line 8: Fragments are supposed to be a client-side feature and user agents shouldn't be sending it to the client.  Was this is to fix a case where they were?

dr-shorthair: Doh. 
No, it was a fix for potential future use with fragments using either / or # but I forgot that the # does not go to the server. Will undo that.

dr-shorthair: (thank you for your careful reading!)

PR #2183, opened 2021-05-26
Add six networked twins

juusoautiosalo:
Add six networked twins that will be used in tests for an article.


dgarijo:
@juusoautiosalo,
you don't need to add a readme file per subfolder. Having one in the root folder in your repo is fine.

PR #2187, opened 2021-05-27
ordodata FDP

rajaram5:
Hello,

I have created redirecting config for ordodata. This configuration redirect call for `FAIR Data Point` of ordodata.

Thanks for the help

### DIFF
--- a/orphadata/.htaccess
+++ b/orphadata/.htaccess
@@ -1,0 +1,4 @@
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^/fairdatapoint?(.*) http://95.142.173.26:8090/$1 [R=307,L]
+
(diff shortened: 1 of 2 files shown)


orphadata/.htaccess
```diff
@@ -0,0 +1,4 @@
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^/fairdatapoint?(.*) http://95.142.173.26:8090/$1 [R=307,L]
```
davidlehn, line 3: - That first `/` isn't needed.
- The `?` is going to make `t` optional, is that the intent?

rajaram5: Hi @davidlehn thanks for the quick response. Thanks for finding my bugs. I wanted something like this `^fairdatapoint/?(.*)`. I will push my changes now

PR #2191, opened 2021-05-29
Move deprecated security suites to the bottom of the file

OR13:
Previously https://github.com/perma-id/w3id.org/pull/2175

### DIFF
--- a/security/.htaccess
+++ b/security/.htaccess
@@ -10,17 +10,6 @@ RewriteRule ^v2$ https://w3c-ccg.github.io/security-vocab/contexts/security-v2.j
 
 RewriteRule ^v3-unstable$ https://w3c-ccg.github.io/security-vocab/contexts/security-v3-unstable.jsonld [R=302,L]
 
-# TODO: remove "suites" not rooted under  "suites"
-
-RewriteRule ^jws$ https://w3c-ccg.github.io/lds-jws2020/ [R=302,L]
-RewriteRule ^jws/v1$ https://w3c-ccg.github.io/lds-jws2020/contexts/v1/ [R=302,L]
-RewriteRule ^pgp$ https://or13.github.io/lds-pgp2021/ [R=302,L]
-RewriteRule ^pgp/v1$ https://or13.github.io/lds-pgp2021/contexts/pgp-v1.jsonld [R=302,L]
-RewriteRule ^bbs$ https://github.com/w3c-ccg/ldp-bbs2020/ [R=302,L]
-RewriteRule ^bbs/v1$ https://w3c-ccg.github.io/ldp-bbs2020/contexts/v1/ [R=302,L]
-RewriteRule ^blockchain$ https://or13.github.io/lds-blockchain2021/ [R=302,L]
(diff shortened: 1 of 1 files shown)


security/.htaccess
```diff
@@ -79,8 +67,19 @@ RewriteRule ^suites/secp256k1-2020/v1$ https://ns.did.ai/suites/secp256k1-2020/v
...
 RewriteRule ^suites/secp256k1-2019/v1$ https://ns.did.ai/suites/secp256k1-2019/v1 [R=302,L]
 
-
 # http://w3id.org/security/suites/x25519-2019
 
 RewriteRule ^suites/x25519-2019$ https://ns.did.ai/suites/x25519-2019 [R=302,L]
 RewriteRule ^suites/x25519-2019/v1$ https://ns.did.ai/suites/x25519-2019/v1 [R=302,L]
+
+# 🚧 DEPRECATED as of 04/29/2021 🚧
+# The content of thesse contexts is available under /suites above.
```
OR13, line 76:
```suggestion
# The content of these contexts is available under /suites above.
```

security/.htaccess
```diff
@@ -79,8 +67,19 @@ RewriteRule ^suites/secp256k1-2020/v1$ https://ns.did.ai/suites/secp256k1-2020/v
 RewriteRule ^suites/secp256k1-2019$ https://ns.did.ai/suites/secp256k1-2019 [R=302,L]
 RewriteRule ^suites/secp256k1-2019/v1$ https://ns.did.ai/suites/secp256k1-2019/v1 [R=302,L]
 
-
 # http://w3id.org/security/suites/x25519-2019
 
 RewriteRule ^suites/x25519-2019$ https://ns.did.ai/suites/x25519-2019 [R=302,L]
 RewriteRule ^suites/x25519-2019/v1$ https://ns.did.ai/suites/x25519-2019/v1 [R=302,L]
+
+# 🚧 DEPRECATED as of 04/29/2021 🚧
```
davidlehn, line 75: This is maybe a silly complaint, but would it be ok to just use simple ASCII here?

OR13: the point of the emoji is to make it clear there are deprecations, is reducing the visibility of deprecation warnings an enhancement to this PR?

Certainly if there is an encoding issue that this causes, I would remove it, but if this is just aesthetic preferences, I prefer to see deprecation warnings loud and clear.

OR13: @davidlehn feel free to suggest changes, if lots of other folks prefer your aesthetic preferences, I am happy to accept them.

davidlehn: I thought CAPITALIZATION was enough. :-)  And/or just some `################` delimiters.  I do have a path through some servers that has some console unicode issues, but I'll deal with the `?`s.

TallTed: Keeping in mind that not all emoji are visible on all systems (e.g., the `?`s displayed in some of @davidlehn's usage), I would recommend sticking with ASCII flags in places like this.  If you really want the construction barriers, you could probably find or recreate some ASCII art that delivers them. :-)

PR #2196, opened 2021-06-01
added namespace claim

ComeIn-NRW:
maintainer: https://github.com/ComeIn-NRW

### DIFF
--- a/comein/.htaccess
+++ b/comein/.htaccess
@@ -1,0 +1,3 @@
+RewriteEngine on
+RewriteRule ^vocabs(/)?$ https://comein-nrw.github.io/vocabs/$1 [R=302,L]
+RewriteRule ^vocabs/(.*) https://comein-nrw.github.io/vocabs/w3id.org/comein/vocabs/$1 [R=302,L]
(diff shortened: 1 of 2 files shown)


comein/.htaccess
```diff
@@ -0,0 +1,3 @@
+RewriteEngine on
+RewriteRule ^vocabs(/)?$ https://comein-nrw.github.io/vocabs/$1 [R=302,L]
```
davidlehn, line 2: This could be simplified since the rule is static:
```
RewriteRule ^vocabs/?$ https://comein-nrw.github.io/vocabs/ [R=302,L]
```

comein/README.md
```diff
@@ -0,0 +1,6 @@
+# ComeIn NRW 
+
+The w3id.org URIs are used as permanent URIs for metadata specifications
+
+* Project: [ComeIn NRW](https://www.uni-due.de/comein)
+* Contact: [Manuel Oellers](https://www.uni-due.de/comein/metaportal/)
```
davidlehn, line 6: If possible could you add the github maintainer id here too?  It was in the PR commit message, but we'd check here first.

PR #2201, opened 2021-06-02
Added one more ontology module to w3id.org/devops-infra/

ocorcho:


dgarijo:
@ocorcho, 
why not use 303 redirections for content negotiation?

ocorcho:
This is just in case that there is a final decision in the coming months on moving from github pages as a publication means to another server.

dgarijo:
@ocorcho,
303 is ok in that case too, as they are all redirects. However, 303 is what is recommended for sem. web resources.

PR #2202, opened 2021-06-03
create redirect for ontology for genial project

wawrzik:
Not sure if this works. Not so much experience with .htaccess file
Just want to have https://wawrzik.github.io/GENIALOntologies/index-en.html
this here:  w3id.org/gbo

please make changes if required thank you!
if you have questions please email.


davidlehn:
The `.htaccess.rtf` file needs to be converted to plain text and renamed to `.htaccess`.

wawrzik:
ok now?

gbo/.htaccess
```diff
@@ -0,0 +1,13 @@
...
+
+Options -MultiViews
+
+
+# Rewrite engine setup
+
+RewriteEngine On
+
+
+RewriteRule ^$ https://wawrzik.github.io/GENIALOntologies/index-en.html [R=301,NE,L]
```
davidlehn, line 13: Be aware this is 301 Moved Permanently, so if you change the redirect in the future, that could cause issues.

PR #2203, opened 2021-06-04
changed it to temporary uri instead of permanent

wawrzik:
Thanks for the info. I updated accordingly.


dgarijo:
@wawrzik if you plan to serve the serializations of your ontology with content negotiation, you should use 303. See https://github.com/perma-id/w3id.org/blob/master/example/.htaccess for an example if interested.

PR #2206, opened 2021-06-05
Create agrochemicals-ontology

marco-medici:
add link to the ontology


dgarijo:
@marco-medici,
you are creating a file in the root folder, which goes against how we handle w3ids. You need to create a folder, and inside an `.htaccess` file and a readme file. You may use https://github.com/perma-id/w3id.org/tree/master/example for inspiration in your redirection.

PR #2211, opened 2021-06-08
Add vc-http-api

OR13:


vc-http-api/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+- https://w3c-ccg.github.io/vc-http-api
+- https://github.com/w3c-ccg/vc-http-api
+
+- https://w3c-ccg.github.io/traceability-interop
+- https://github.com/w3c-ccg/traceability-interop
+
+## Maintainers
+
+https://github.com/w3c-ccg/vc-http-api/blob/main/CODEOWNERS
```
davidlehn, line 11: I see what you are trying to do but I find this indirection problematic.  The first issue is it's putting the effort on w3id.org maintainers to chase down and check who has access every time.  This service is still small enough we do this by hand, or maybe just check a files history to see who had access.  In the future there may be metadata to automate things, but I'm not sure how a remote CODEOWNERS would work.  That CODEOWNERS format is specifying who owns file patterns in the vc-http-api repo, not who owns the files here.  What if that file is removed, or changes to some other pattern than '*', then who has access here?  I think it's better to just list the github ids here and update as needed.

PR #2221, opened 2021-06-11
EJP RD services added

rajaram5:
Hello,

I have added redirect rules for services used in the EJP RD project. EJP RD is a EU funded project aim to provide services to facilitate research in rare disease field.

Thanks for the help.


dgarijo:
@rajaram5,
can you please add your github id together with your maintainer name? It's easier for us to check in case another Rajaram makes a change to this w3id.
In addition, do you mind removing the edit to the gitignore?

.gitignore
```diff
@@ -4,3 +4,5 @@
 *.sw[nop]
 *~
 .DS_Store
+
+ejp-rd/README.md.backup
```
davidlehn, line 8: Remove this.  If you think *.backup would be generally useful to ignore, please create another PR for that change.

ejp-rd/.htaccess
```diff
@@ -0,0 +1,6 @@
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^fairdatapoints/bbmri?(.*) https://ejprd.fair-dtls.surf-hosted.nl/bbmri-fdp/$1 [R=307,L]
+RewriteRule ^fairdatapoints/orphanet-catalog-fdp?(.*) https://ejprd.fair-dtls.surf-hosted.nl/orphanet-catalog-fdp/$1 [R=307,L]
+RewriteRule ^fairdatapoint-index?(.*) http://ejprd.fair-dtls.surf-hosted.nl:8070/$1 [R=307,L]
+RewriteRule ^triplestores/graphdb?(.*) http://ejprd.fair-dtls.surf-hosted.nl:7200/$1 [R=307,L]
```
davidlehn, line 6: The `?` in all these patterns is going to make the prior *letter* optional, which I assume is wrong.  Perhaps you meant to have an optional trailing `/` before the trailing text match?  If so, use something like this:

```suggestion
RewriteRule ^fairdatapoints/bbmri/?(.*) https://ejprd.fair-dtls.surf-hosted.nl/bbmri-fdp/$1 [R=307,L]
RewriteRule ^fairdatapoints/orphanet-catalog-fdp/?(.*) https://ejprd.fair-dtls.surf-hosted.nl/orphanet-catalog-fdp/$1 [R=307,L]
RewriteRule ^fairdatapoint-index/?(.*) http://ejprd.fair-dtls.surf-hosted.nl:8070/$1 [R=307,L]
RewriteRule ^triplestores/graphdb/?(.*) http://ejprd.fair-dtls.surf-hosted.nl:7200/$1 [R=307,L]
```

Note that this isn't fully catching all errors on the w3id.org side, as bbmriFOO would get through as bbmri-fdp/FOO. Rules could be adjusted to handle this, but it's easier to just ignore the garbage-in, garbage-out situation unless it's a problem for correct inputs.

rajaram5: @davidlehn Thank you for the review. Yes, I meant to have trailing `/` in the URLs. I missed it.

rajaram5: About `bbmri-fdp/FOO` case it is a bit hard to write rules against it. The FDP software used in this project generates thousands of RDF records which need to have PIDs. I will investigate if the suffix of these records can be fixed.

dgarijo:
The .gitignore is still there despite the most recent changes.

rajaram5:
> The .gitignore is still there despite the most recent changes.

I restored the `gitignore` file to original state now

PR #2238, opened 2021-06-29
ADD asbing 

AnneGoebels:
redirection for ASB-ING Ontology


dgarijo:
@AnneGoebels,
you are feferring to documents which are not in the current repository. You should replace  things like 
```
RewriteRule ^$ index-en.html [R=303,L]
```
with the URL of the server hosting that file.
See https://github.com/perma-id/w3id.org/blob/master/example/.htaccess
for an example

davidlehn:
@dgarijo I think `RewriteBase` will take care of that?

@AnneGoebels You have your base set to the github UI URLs.  You probably want to point at a github pages site or other place where the raw files are hosted.

dgarijo:
True, my bad, i skipped to the redirection section without seeing the rewrite base!

AnneGoebels:
@davidlehn true, sorry, I will change the base URI to a github pages site

PR #2245, opened 2021-07-01
Update .htaccess for dio

nimonika:


dgarijo:
@nimonika one small comment: 
https://github.com/nimonika/ontologies/blob/master/dio.ttl does not redirect to the TTL file, which I think it is what you may want. It redirects to the GitHub page containing the ttl file.
You may use a github page or similar service to serve the TTL directly from github.
If you intend to serve the github html,  then let me know and I will merge the PR.

nimonika:
> @nimonika one small comment:
> https://github.com/nimonika/ontologies/blob/master/dio.ttl does not redirect to the TTL file, which I think it is what you may want. It redirects to the GitHub page containing the ttl file.
> You may use a github page or similar service to serve the TTL directly from github.
> If you intend to serve the github html, then let me know and I will merge the PR.

@dgarijo Hi Daniel, I have added content negotiation through Github pages. Thanks for the tip.

dgarijo:
@nimonika,
now you are adding the redirection for `https://w3id.org/dio/dio` instead of `https://w3id.org/dio`. Is this what you intend? 
If you want the latter, you should remove `dio` in the rewrite rules: `RewriteRule ^dio$` --> `RewriteRule ^$`

nimonika:
> @nimonika,
> now you are adding the redirection for `https://w3id.org/dio/dio` instead of `https://w3id.org/dio`. Is this what you intend?
> If you want the latter, you should remove `dio` in the rewrite rules: `RewriteRule ^dio$` --> `RewriteRule ^$`

Oops!Fixed it now.

dgarijo:
@nimonika,
as you have organized your rules, the first rule will always activate, and the content negotiation will not work properly. You'll probably want to add something like the following for serving HTML only when requested.:

```
RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+xml)
RewriteRule ^$ link to HTML [R=303,L]
```
You can also see a complete redirection example [here](https://github.com/perma-id/w3id.org/tree/master/example)

PR #2250, opened 2021-07-06
iadopt: add redirections

SirkoS:
I added redirects for the I-Adopt framework ontology ( https://i-adopt.github.io/ ).

please let me know, if there is something else you need.

### DIFF
--- a/iadopt/.htaccess
+++ b/iadopt/.htaccess
@@ -1,0 +1,44 @@
+# Turn off MultiViews
+Options -MultiViews
+
+# Directive to ensure *.rdf files served as appropriate content type,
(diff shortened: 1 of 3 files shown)


dgarijo:
@SirkoS 
I merged it, but isn't a bit weird to add 2 redirects for the same ontology? Why don't you redirect one to the other?

SirkoS:
fair point. from my point of view, this would only simplify configuration but not impact the overall result. right?

dgarijo:
It helps maintainability. For example, if you want to publish a new version of your ontology, you only need to change it in one file

PR #2262, opened 2021-07-11
nta8035: Add redirect and Readme

redmer:
Hi 👋, this PR adds `/nta8035` for Dutch national technical agreement NTA 8035, that is hosted at BIM Loket.


dgarijo:
@redmer,
 note that for trig, turtle and rdf/xml you are creating the w3id `https://w3id.org/nta8035` instead of `https://w3id.org/nta8035/def`. I think you may want to change your rewrite rules to `RewriteRule ^def$`

redmer:
Great! I've added it with 5e28d62.

PR #2264, opened 2021-07-12
Update versioning redirects to multiple accepted content types

JohannesLipp:


dgarijo:
@JohannesLipp,
`N3 with versioning ` is redirecting to the turtle representation. Is this intended?

JohannesLipp:
> @JohannesLipp,
> `N3 with versioning ` is redirecting to the turtle representation. Is this intended?

@dgarijo thank you for pointing out this copy-paste error. We just fixed it, thanks.

There is still an error with our backend, which we will fix soon. The redirects should be fine now, thank you.

PR #2266, opened 2021-07-13
g0-Builder-J2RM and local .htaccess updaates

srodriguez142857:
- Adding the identifier for the project "g0-Builder-J2RM".
- Updating local `.htaccess` files


srodriguez142857:
Rename identifier (folder) to "J2RM"

dgarijo:
@srodriguez142857,
note that `303` redirects are usually for semantic web resources. If you are just going to redirect to a page I suggest using 301 or 302

PR #2282, opened 2021-07-28
Make identifier more meaningful before submission.

mu40:


dgarijo:
@mu40,
renaming w3ids is nice, but since you reserved  `f71c4a`  in the first place, are you sure you don't want to leave a redirect? The whole purpose of having permanent w3ids is that they remain permanent, even if you decide to delete or move the content referred to those URLs.

@davidlehn @nicholascar shouldn't we have a policy regarding w3id deprecation/maintenance?

mu40:
I agree with the point of the identifiers being permanent. The original one here was reserved very recently and not distributed yet, so I don't really mind breaking it unless you insist.

dgarijo:
It's not a written rule (yet). If you are in a hurry, I will merge the PR, but I would like to have this as a precedent to decide what we do with these cases

davidlehn:
I think it's up to the maintainers to know what's best, especially for a new id that may have limited or no outside use.  No reason to keep old redirects around that will never be used.  And less weird hex names is fine too. ;-)

I'm not sure about that README.md symlink idea.  Might be better to have a regular file with content that renders. Something like:
```markdown
Moved to [synthmorph](../synthmorph/README.md).
```

mu40:
Thanks, David. I doubt that the old redirect will have any use; should I open a new PR to remove it along with the symlink?

PR #2314, opened 2021-09-03
related-to: new project about providing data about other identifiers

rgiessmann:
this is a new project about providing identifiers which integrate to other identifiers. Happy to discuss the overall scope when time allows, but a quick merge already would be very much appreciated!


davidlehn:
What's the reasoning for this 4500+(!) line `related-to/doi.org/10.5281/zenodo.3978439/.htaccess` file?  Can all those "entry*" lines be simplified into just one single pattern match line?  w3id isn't intended to do URL validation for  arbitrary number ranges.  That's better done with a database rather than rewrite rules.  It's better to handle such things on the target server or at some other application level.

rgiessmann:
Hi @davidlehn , thanks for chipping in! I believe it's good to have this discussion!

The motivation is this: I am trying to make a friend's research data more FAIR. He collected data in a csv file. The data is at some point going to be transferred into a "database", where it will be amended to include more information, and being enriched towards "semantic web" principles. 

We don't have any infrastructure to run a dynamic web service on, nor any application. The goal is to have zero costs, to make this project truly sustainable -- another service he provides continuously fights with the running costs (speaking only about 100$ a year), which is hard to find money for if you don't want to pay it out of your own pocket. And even if you do pay it yourself, your resource will go away some day when you're dead. 

Further, the motivation is to have the possibility to reply with "410 Gone" when one of the entries in the  "database" needs to be deleted, and/or to point to an updated version of the data.

Why do we need those identifiers at all? We are mapping the "entries" (1 to 4.5k) to another person's database, which contains 5k entries, and want to produce exact-match, subset-of, or alike mappings for each individual entry. 


So we are looking for truly persistent identifiers, and I thought that w3id.org would be the best place for this.

I believed that 4.5k entries would not impact the service quality of w3id.org, but I might be mistaken here. 


Regarding alternatives to 4.5k entries: I thought about using fragment identifiers ( ...#entry1, ...#entry2 instead of .../entry1, .../entry2 ), but this would not allow to redirect individual entries to different locations (most likely to happen in the future), nor to reply "410 Gone".


But I do understand if you want to keep w3id.org focused on providing "lean" redirects for a few identifiers per project only; we would need to look into another option then, though. Maybe you have a suggestion on this, too?


I would also be interested to measure the service quality (whether 4.5k entries have an impact or not), but I'm definitely not an expert on this...

What do you think? (pinging @dgarijo here, too, not sure how visible a closed pull request is...) Thanks and best,
   Robert

dgarijo:
@rgiessmann,
I think that you can maybe create one (or multiple) regular expression that captures  `files/TECRDB.csv/entryNNNN` where NNNN is up to the maximum id you are looking for. That way you avoid having so many things in the htaccess and can scope the range of ids to something more manageable

rgiessmann:
hi @davidlehn @dgarijo , I came up with the regex ^([1-9]|[0-9]{2}|[0-9]{3}|[123][0-9]{3}|4[0-4][0-9]{2}|45[0-3][0-9]|454[0-4])$ for matching 1 to 4544. I guess that should do the job. Would you prefer this in the file? I will propose the corresponding PR then asap.

And do you have a strong opinion about the general problem regarding "URL validation" vs. "simple redirection"? Do you have any preference on a way forward to come to a consensus?

Thanks and best! Robert

davidlehn:
Re costs: You're trying to externalize your costs to w3id.org, and w3id.org is trying to externalize costs to you. 
 That's why we try to promote simple redirects and let the target server do any validation needed.

Re identifiers: Lots of redirects here support an arbitrary number of identifiers via wildcards.  That has the advantage of letting validation happen somewhere else, and the rules here are simple to maintain.  A giant list of 4500 entries is difficult to impossible to maintain or check for correctness by hand.  If you intend to update that with any frequency, it seems difficult and error prone.  For instance if that csv file changes.  There's probably another solution for all this, but you're the expert on that.

Re performance: I doubt it will be an issue at all unless you are are sending massive amounts of traffic to the endpoints, which would be an issue with any setup.

Re regex: That regex looks too complex to maintain.  And I think has errors like allowing 0N and 0NN.  Fixable, but it's just an issue going forward if the ranges ever change.  There are regex range generators out there if that helps.

Re rules: I'm not even sure how that file works!  Looks like entryXXX will go to that zenodo link.  I'm not sure where that ENV thing comes from.  At first glance it's weird entry1, 2, and 3-N are different.  From a client viewpoint, it's also going to be difficult to know what to do with everything redirecting to the same file.  It's acting more like an existence test than sending to an appropriate resource identifier.

rgiessmann:
Hi @davidlehn , thanks for your quick reply!

You are true about updating: this is a pain when happening with any frequency, and maybe difficult. Indeed the csv file is not supposed to change, and if so, the w3ids in inside are supposed to stay constant. 

About regex: true, that one was flawed.


About rules / what I am trying to solve here: I want to create a semantic web version of "speaking about rows in a csv file". The csv file sits on Zenodo; Zenodo gives DOIs to the whole package of data + metadata, but does not provide an identifier to the file itself. 

In the file there are rows which correspond to one set of information; I want to address each row individually. 

The client, as of right now, would retrieve the whole file, and would need to lookup the entryN identifier to identify the correct row. If the entryN would be outside of the range, though, the client would immediately get a 404, and would not need to retrieve the file at all. Indeed, this is merely an existence test. 



I now found https://tools.ietf.org/html/rfc7111 -- which introduced fragment identifiers specifically for csv files. I will go with this approach, until we have a proper server / infrastructure running which can resolve identifiers on their side.

PR #2318, opened 2021-09-06
Hysope2 init

zawadzl:


dgarijo:
@zawadzl,
there is no redirection. Can you please add one?

zawadzl:
Sorry @dgarijo, we do not know the exact redirection URL just yet. Do you mind if we add this later this year ?
Or I can just add the file with a "random" URL and change it later. What do you prefer?

dgarijo:
We do not accept reserving w3ids, so I would suggest that if you don't know it yet, then add a temporary URL that you can later change by another one.

zawadzl:
OK I will, sorry about that!
Thanks

zawadzl:
OK, done @dgarijo

PR #2301, opened 2021-08-25
yarrrml: add repository link

DylanVanAssche:
YARRRML's Matey is now publicly available on Github, created a w3id.org perma-id for it.

PR #2344, opened 2021-10-05
yarrrml: fixes & improvements

DylanVanAssche:
- Use HTTPS by default as suggested in #2301 
- Simplify rules as suggested in #2301 
- Add redirect for `tutorial/getting-started` as well.

CC: @pheyvaer


pheyvaer:
Looks good to me!

PR #2346, opened 2021-10-06
dAdd redirectrule for w3id.org/people/michael

carducci:


carducci:
Adding a w3id people identifier for michael

PR #2396, opened 2021-11-08
Add redirect for lrmi-profile

acka47:


acka47:
Thanks for the quick merge, @dgarijo !

PR #2398, opened 2021-11-09
Update .htaccess

maieutike:


dgarijo:
@maieutike,
please add a readme file for this repository adding yourself as a maintainer.
This is a requirement we have started to adopt for all w3ids that require changes, or new w3ids, so we can track who can and cannot edit them. Thanks in advance.

maieutike:
@dgarijo,
Thank you for the kind feedback. Done as requested.

PR #2444, opened 2021-12-18
added my ontology in videogameontology

dangeloandrea14:


dgarijo:
@dangeloandrea14, you are missing a readme file of your ontology. Also, note the potential name conflict with http://purl.org/net/VideoGameOntology#

dangeloandrea14:
Thanks @dgarijo , my bad on the readme. I noticed it was erroneously missing but I wanted to wait for feedback first. I also changed the directory to /videogameontology to /disimgameontology, that should be free. Do I need to make another pull request or are the edits embedded in this one?

dgarijo:
This PR is fine. I am merging your request, but please have a look at the comments I have left you in the PR

disimgameontology/.htaccess
```diff
@@ -0,0 +1,23 @@
+Options +FollowSymLinks
+# Turn off MultiViews
+Options -MultiViews
+
+# Directive to ensure *.rdf files served as appropriate content type,
+# if not present in main apache config
+AddType application/rdf+xml .rdf
+AddType application/rdf+xml .owl
+AddType text/turtle .ttl
```
dgarijo, line 9: If you are not serving ttl, why have these additional types?

disimgameontology/.htaccess
```diff
@@ -0,0 +1,23 @@
...
+
+RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+xml|text/\*|\*/\*)
+RewriteCond %{HTTP_ACCEPT} text/html [OR]
+RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
+RewriteCond %{HTTP_ACCEPT} text/\* [OR]
+RewriteCond %{HTTP_ACCEPT} \*/\* [OR]
+RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
+RewriteRule ^ontology$ https://dangeloandrea14.github.io/  [R=303,NE,L]
+RewriteCond %{HTTP_ACCEPT} ^.*application/rdf\+xml.* 
+RewriteRule ^ontology$ https://github.com/dangeloandrea14/videogameontology/blob/main/videogames273214.owl [R=303,NE,L]
```
dgarijo, line 21: You are adding a rule for rdf/xml, and then you also state that by default the same file should be returned. So having only the second rule should suffice.

dangeloandrea14:
Thank you for your comments. You're helping me immensely because it's the first time I have to deal with rewrite rules. I'm trying to achieve something similar to what you did in your guide, so I'd basically want w3id.org/disimgameontology to redirect to the html of my ontology, but that doesn't seem to happen as of now, only the folder index is showing. Once I figure out what I'm doing wrong I'll fix the .htaccess file entirely

dangeloandrea14:
Actually woops, scratch that. I defined my rewrite rules for disimgameontology/ontology, and that does in fact work. Sorry. (: And thanks again

dgarijo:
@dangeloandrea14,
no worries, it takes time to do it right the first time.
If you want a working example managing ontology versions too, have a look at https://github.com/perma-id/w3id.org/blob/master/example/.htaccess

PR #2458, opened 2022-01-07
redirects for pylode view and turtle download

smrgeoinfo:
add redirects for pylode view and turtle download


smrgeoinfo:
wait on this one-- gotta do some fixes.

PR #2460, opened 2022-01-08
add folders for version number

smrgeoinfo:
the version number is getting rewritten twice, so copy what's in material vocabulary,which is working


dgarijo:
@smrgeoinfo,
just for the record, we only need a readme in the root w3id you are creating, not in every sub folder :)

smrgeoinfo:
@dgarijo thanks for the pointer. I hope more documentation doesn't cause problems?

dgarijo:
Not at all! It's just in case you thought it was a requirement.

PR #2461, opened 2022-01-09
added htaccess files for ontologies m3 and tcm

audunven:


latuli/.htaccess
```diff
@@ -0,0 +1,4 @@
+Options +FollowSymLinks
+RewriteEngine on
+#Rewrite rules for LATULI (no content negotiation)
+RewriteRule ^latuli$ https://audunven.github.io/LATULI/ [R=302,L]
```
dgarijo, line 4: Here you are creating the w3id: https://w3id/org/latuli/latuli. Is this what you intend? Or is there a redundant "latuli"?

dgarijo: if it's a typo, just do: `RewriteRule ^$ https://audunven.github.io/LATULI/`

audunven: Ah, thanks for spotting, should be resolved now

audunven: I cannot access the github pages redirected to from the rewrite rules in latuli/ontology/.htaccess. 
( E.g., RewriteRule ^m3$ https://audunven.github.io/LATULI/ontology/m3/m3Doc/  [R=303,NE,L] )
I´ve probably made some mistake in my .htaccess file in the ontology folder since I get a 500 error?

dgarijo: You need to comment out lines 25 and 40, which are not valid htaccess syntax.
I recommend you try your htaccess files locally before submitting, just to be sure. You can do this with an Apache Tomcat server.

audunven: Ok, thanks for helping out. Commented out lines 25 and 40 as suggested. Should I do a new pull request or revert the merged commit you did yesterday?

dgarijo: A new pull request will suffice

PR #2467, opened 2022-01-13
fix punctuation and typo in README

TallTed:
@noelmcloughlin -- Please approve.

### DIFF
--- a/td/README.md
+++ b/td/README.md
@@ -1,8 +1,8 @@
 # /td/
-This [W3ID](https://w3id.org) provides a persistent URI namespace for Toronto-Domion Bank resources.
+This [W3ID](https://w3id.org) provides a persistent URI namespace for Toronto-Dominion Bank resources.
 
 ## Uses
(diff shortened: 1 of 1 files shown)


noelmcloughlin:
Yes I approve. thank you. pls merge

noelmcloughlin:
Would it be possible to update the commit message to remove my name. 
Otherwise I get pinged when people are working with the commits in their fork

image

davidlehn:
@TallTed In the future please don't use messages like that in commit messages.  That title was fine by itself.

@noelmcloughlin Sorry I didn't notice that message before.  I just read the PR text.  How bad is this issue?  We'd have to edit the message and force push to master which would cause some havoc for everyone that has pulled that change in already.

noelmcloughlin:
I just got pinged from someones branch, so email looks like a new zombie PR, I can live with this, probably get a few pings over time but can ignore it.   You could look at introducing CommitLint, maybe it can catch that, although 1st time I saw this issue.

PR #2472, opened 2022-01-16
bearchaeo

vlombard:
Adding bearchaeo permanent URL


davidlehn:
Can you squash these commits?  Or do you mind if we do before merging?
For two reasons: to avoid the churn and to avoid that owl large file getting in the history.

vlombard:
Sure, please do it. 
Sorry for not having done it myself.
Thanks.
Vincenzo

PR #2492, opened 2022-01-28
Create w3id link for OFO

AlexDonkers:


ofo/.htaccess
```diff
@@ -0,0 +1,63 @@
...
+RewriteRule ^$ https://alexdonkers.github.io/ofo/406.html [R=406,L]
+
+# Default response
+# ---------------------------
+# Rewrite rule to serve the RDF/XML content from the vocabulary URI by default
+RewriteRule ^$ https://alexdonkers.github.io/ofo/ [R=303,L]
+
+
+# If suffix html, redirect to html version
+RewriteRule ^ofo.html$ https://alexdonkers.github.io/ofo/ [R=303,L]
```
dgarijo, line 51: This is creating a rule for w3id.org/ofo/ofo.html. Is this what you intend?

AlexDonkers: Hi, I'm sorry, this rule shouldn't be there. I'll delete it.

dgarijo: Same thing happens for the other serializations: ofo.ttl, jsonld, etc. Is this also an error?

AlexDonkers: Just deleted all these similar rules. Hope the file is fine now.

PR #2504, opened 2022-02-02
rail/topo: add redirect to version ontology

stefanbischof:
- redirect from generic to most recent version, currently v1.0
- fix redirects for ontology version URIs

### DIFF
--- a/rail/topo/.htaccess
+++ b/rail/topo/.htaccess
@@ -13,7 +13,7 @@ RewriteEngine On
 
 # Forward to most recent version
 
-RewriteRule ^$ /v1.0 [L]
+RewriteRule ^$ v1.0 [L]
 
 # Version-specific redirects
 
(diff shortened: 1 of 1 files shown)


rail/topo/.htaccess
```diff
@@ -11,37 +11,43 @@ AddType application/ld+json .jsonld
 # Rewrite engine setup
 RewriteEngine On
 
+# Forward to most recent version
+
+RewriteRule ^$ /v1.0 [L]
```
davidlehn, line 16: That would try to redirect to `https://w3id.org/v1.0`.  I think you can just remove the `/` for it to rerun rules here :
```suggestion
RewriteRule ^$ v1.0 [L]
```

stefanbischof: Good catch, thanks!
worked this way when I tested your suggestion locally.

PR #2510, opened 2022-02-03
book domain ontologies redirect

jucola98:


dgarijo:
@jucola98,
- please add a readme file indicating the maintainers.
- you seem to be adding additional empty folders, not just book
- you are adding a redirection to book/asteroids, which I don't think it's what you intend

See https://github.com/perma-id/w3id.org/tree/master/example for an example

jucola98:
@dgarijo now seems ok, let me know if you can accept the pull request

PR #2524, opened 2022-02-10
Reservoir Risks Ontology

patriciaferreiradasilva:
Added ResRiskOnto, an ontology for reservoir- related risk events in the oil & gas domain.


davidlehn:
Please add a README or comments in the .htaccess file listing the maintainers github ids and/or other contact info. We use that to check who can do future updates. Thanks.

patriciaferreiradasilva:
Added README with ontology description and contact as suggested.

PR #2548, opened 2022-02-21
Upgrade to facilitate dataset directory

docuracy:
Would you please again check my redirects, which I've annotated with examples of what I hope to achieve?

Thanks again for this wonderful service.


docuracy:
I realise that I've used standard regex ^ instead of ! - will revise and try again soon...

PR #2556, opened 2022-02-25
Add entry for smarcql ontology.

realworldobject:
The smarcql.ttl file isn't published yet, but will be soon.

### DIFF
--- a/smarcql/.htaccess
+++ b/smarcql/.htaccess
@@ -1,0 +1,15 @@
+Options +FollowSymLinks
+RewriteEngine On
+
+# Turn off MultiViews
+Options -MultiViews
+
+# Directive to ensure *.ttl files served as appropriate content type,
+# if not present in main apache config
+AddType text/turtle .ttl
+
+# Rewrite rule to redirect 303 from any class or prop URI
+RewriteRule ^smarcql/.+ https://quxfarm.github.io/smarcql/ontology/smarcql.ttl [R=303]
+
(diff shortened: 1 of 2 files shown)


smarcql/.htaccess
```diff
@@ -0,0 +1,15 @@
...
+
+# Directive to ensure *.ttl files served as appropriate content type,
+# if not present in main apache config
+AddType text/turtle .ttl
+
+# Rewrite rule to redirect 303 from any class or prop URI
+RewriteRule ^smarcql/.+ https://quxfarm.github.io/smarcql/ontology/smarcql.ttl [R=303]
+
+# Rewrite rule to serve RDF content from the vocabulary URI
+RewriteRule ^smarcql/$ https://quxfarm.github.io/smarcql/ontology/smarcql.ttl
```
davidlehn, line 15: - These rules are relative to the current dir, not root.  I assume you didn't mean to double up on the name.
- Looks like you renamed yourself from `quxfarm` and the old URLs don't work.  Perhaps you meant to change them here too?
- Since both are going the same place, could probably just use one rule with `.*` as well (not done in this change suggestion).

Maybe something like the following:
```suggestion
# Rewrite rule to redirect 303 from any class or prop URI
RewriteRule ^.+ https://realworldobject.github.io/smarcql/ontology/smarcql.ttl [R=303]

# Rewrite rule to serve RDF content from the vocabulary URI
RewriteRule ^$ https://realworldobject.github.io/smarcql/ontology/smarcql.ttl
```

PR #2571, opened 2022-03-08
Frontier_project

physta:
Ontology


dgarijo:
@physta,
you are adding your ontology directly, which is not supported. You need to add a redirection to wherever your ontology lives. You are also missing a readme file. See https://github.com/perma-id/w3id.org#suitable-pr-content for more information

PR #2576, opened 2022-03-08
Frontier_physta

physta:


dgarijo:
@physta,
you are adding all the ontology documentation files to this repository. Instead, you should point to your own repository. See  https://github.com/perma-id/w3id.org/blob/master/example/.htaccess for an example

dgarijo:
@physta, I am closing this PR because it's adding files that should not be there.

PR #2578, opened 2022-03-08
Update .htaccess

physta:


dgarijo:
@physta, you are not redirecting to the right sources. 
You should point to your github pages url, which should be something like: https://physta.github.io/frontier/OnToology/Frontier.ttl/documentation/doc/index-en.html#$1

physta:
Many thanks for your help Daniel. I have updated the file accordingly. Hopefully now it works.

Best regards, 

Pol.

PR #2588, opened 2022-03-10
Edit the fair-enough namespace

vemonet:
Edit the `.htaccess` file for the `fair-enough` namespace to allow POST data to be passed to the redirects

### DIFF
--- a/fair-enough/.htaccess
+++ b/fair-enough/.htaccess
@@ -2,6 +2,6 @@ Header set Access-Control-Allow-Origin *
 Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
 Options +FollowSymLinks
 RewriteEngine on
-RewriteRule ^$ https://fair-enough.semanticscience.org [R=302,L]
-RewriteRule ^metrics/tests/(.*)$ https://metrics.api.fair-enough.semanticscience.org/tests/$1 [R=302,L]
-RewriteRule ^(.*)$ https://api.fair-enough.semanticscience.org/$1 [R=302,L]
+RewriteRule ^$ https://fair-enough.semanticscience.org [R=302,P]
+RewriteRule ^metrics/tests/(.*)$ https://metrics.api.fair-enough.semanticscience.org/tests/$1 [R=302,P]
+RewriteRule ^(.*)$ https://api.fair-enough.semanticscience.org/$1 [R=302,P]
--- a/fair-enough/.htaccess
+++ b/fair-enough/.htaccess
@@ -2,6 +2,6 @@ Header set Access-Control-Allow-Origin *
 Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
 Options +FollowSymLinks
 RewriteEngine on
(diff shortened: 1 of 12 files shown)


fair-enough/.htaccess
```diff
@@ -2,6 +2,6 @@ Header set Access-Control-Allow-Origin *
 Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
 Options +FollowSymLinks
 RewriteEngine on
-RewriteRule ^$ https://fair-enough.semanticscience.org [R=302,L]
-RewriteRule ^metrics/tests/(.*)$ https://metrics.api.fair-enough.semanticscience.org/tests/$1 [R=302,L]
-RewriteRule ^(.*)$ https://api.fair-enough.semanticscience.org/$1 [R=302,L]
+RewriteRule ^$ https://fair-enough.semanticscience.org [R=302,P]
+RewriteRule ^metrics/tests/(.*)$ https://metrics.api.fair-enough.semanticscience.org/tests/$1 [R=302,P]
+RewriteRule ^(.*)$ https://api.fair-enough.semanticscience.org/$1 [R=302,P]
```
davidlehn, line 7: `mod_proxy` is not enabled so I don't think this will work at all.  What were you trying to do?

vemonet: Hi @davidlehn thanks for the help

I am trying to pass the body of a POST request to my API. I thought it was passed by default, but apparently not:

When using w3id.org the body is not passed:

```bash
curl -L -X 'POST' \            
  'https://w3id.org/fair-enough/metrics/tests/f3-id-in-metadata' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "subject": "https://doi.org/10.1594/PANGAEA.908011"
}'
```

Returns: `{"detail":[{"loc":["body"],"msg":"field required","type":"value_error.missing"}]}` (aka. "body missing")

If I query directly my URL (not through w3id.org) it works fine (it should take a few seconds to complete, that's expected)

```bash
curl -L -X 'POST' \
  'https://metrics.api.fair-enough.semanticscience.org/tests/f3-id-in-metadata' \ 
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "subject": "https://doi.org/10.1594/PANGAEA.908011"
}'
```


Any idea what it could be due to?

davidlehn: You might want to try 307.  The end behavior will probably be what you want, at least with the above curl command. 
 Some info in the HTTP specs, and here:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307

I'm not sure using w3id.org as an API server like that is the best idea.  Being a proxy is out of scope of w3id.org at this time.  And it's not efficient.  Requests will end up sending payload data once to w3id.org, then again to the target server.

davidlehn:
@vemonet Some git strangeness happened here and other files are being edited.  Please fix.  Maybe needs a rebase on top of an updated master branch?

vemonet:
@davidlehn oops, sorry for the inconvenience, I created a fresh pull request to make sure we don't mess with the git history: https://github.com/perma-id/w3id.org/pull/2598

Closing this one then, thanks a lot

PR #2636, opened 2022-04-13
Update and add new W3IDs for the ARK Project

lucymc12:
I have updated and added new W3IDs for the ontologies and terminologies related to the ARK Project.

PR #2637, opened 2022-04-13
Update readme.md

ademarcrotti:
Adding Lucy McKenna as a contributor


davidlehn:
I guessed to use https://github.com/perma-id/w3id.org/pull/2636 instead of this one.  If this email should be used instead, please rebase and we can merge it.  Otherwise I assume this can be closed.

PR #2640, opened 2022-04-14
a database of asd

SilenceSnake:
change name


asdkb/README.md
```diff
@@ -2,7 +2,7 @@ ASD-KG: A Chinese Knowledge Graph for Autism spectrum disorder
 =======
```
TallTed, line 2: It seems like the first line should also change, from —
`ASD-KG: A Chinese Knowledge Graph for Autism spectrum disorder`
— to something like (depending on what your B stands for) —
`ASD-KB: A Chinese Knowledge Base for Autism spectrum disorder`

PR #2644, opened 2022-04-18
Add files via upload

jdmarinj:
Ontology - Energy Efficiency for Industriy


dgarijo:
@jdmarinj, we don't support hosting ontology files directly. Please see https://github.com/perma-id/w3id.org#suitable-pr-content for the types of materials which are suitable for a pull request. And https://github.com/perma-id/w3id.org/tree/master/example with an example on how to do content negotiation on an ontology

jdmarinj:
New folder Ontology ee4i

dgarijo:
@jdmarinj,
you are still leaving all the ontology documentation files in w3id. You need to host them elsewhere, e.g., in a github repository. Please see https://github.com/perma-id/w3id.org#suitable-pr-content for the types of materials which are suitable for a pull request. And https://github.com/perma-id/w3id.org/tree/master/example with an example on how to do content negotiation on an ontology

PR #2646, opened 2022-04-19
Create .htaccess for ride2rail

marioscrock:


davidlehn:
It would have been better to rebase this so those html files were not in the history.  Oh well.

PR #2662, opened 2022-04-26
Redirect to new institute URIs.

awagner-mainz:
After our institute's name has changed, I have opened a new top-level directory and made the old one (301-)redirect to it. The new one contains (302) redirections to the actual data (vocabularies).


davidlehn:
Nice.  Keeping redirects over a rename.  That was always the intent of how to use this service, but so rare to see people do it. :-)

PR #2677, opened 2022-05-04
kg-construct: add survey website

DylanVanAssche:
Add w3id for the survey website.

CC: @dachafra @andimou @pheyvaer


dgarijo:
Since there is an upvote from @pheyvaer, I am merging the request.

PR #2679, opened 2022-05-05
Adding the noria namespace

GenEars:
First release of the *noria* namespace with `README.md` and `.htaccess` files.
Redirect rules tested with the https://htaccess.madewithlove.com/ tool

PR #2684, opened 2022-05-05
noria: correcting target URLs

GenEars:
In relation to https://github.com/perma-id/w3id.org/pull/2679


GenEars:
Thanks @dgarijo . Testings on all RedirectRules are now ok.

PR #2695, opened 2022-05-07
Add OBPD and OBO to perma-id

meta-citizen:
Adding folders and access files to point to projects for Ontology for Building Objects and Ontology for Building Performance Data to the perma-id project.


dgarijo:
Please note the possible ns clash/confusion with the biomedical ontologies: http://prefix.cc/obo

PR #2697, opened 2022-05-09
Add folder for GSC MIxS

sujaypatil96:
Based on the requirement on https://github.com/GenomicsStandardsConsortium/mixs/issues/390 we want to add a folder for the genomic standards consortium MIxS community.

We want to deprecate the [gensc](https://github.com/perma-id/w3id.org/tree/master/gensc) folder we already have registered with w3id.

CC: @ramonawalls @turbomam

### DIFF
--- a/mixs/.htaccess
+++ b/mixs/.htaccess
@@ -1,0 +1,6 @@
+Options +FollowSymLinks
+RewriteEngine on
+
+# Schema elements use github pages
+# Rewrite Base URL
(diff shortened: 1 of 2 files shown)


dgarijo:
@sujaypatil96 if you want to deprecate the other w3id, maybe you can add a redirect to this one?

ramonawalls:
@sujaypatil96 We don't want to deprecate the gensc folder! That is main folder for GSC, and it should be kept, per our discussion at Monday's meeting and https://github.com/GenomicsStandardsConsortium/mixs/issues/233#issuecomment-657666511.

PR #2701, opened 2022-05-10
DigiTwin Updated URI

msbonney:
Changes redirection from HTML version to GitHub .ttl version


dgarijo:
@msbonney note that this is not serving the ttl file, but the github page with the ttl file.

PR #2703, opened 2022-05-12
Add ColActDOnt - Collaborative Actions on Documents Ontology

Muffexx:
Hello everyone.

Please add the ColActDOnt (Collaborative Actions on Documents Ontology) of CEIR (Center for Enterprise Information Research) to perma-id. The publication in our repo follows within the next few weeks.

If changes are required please let me know.

Best
Muffexx

### DIFF
--- a/ColActDOnt/.htaccess
+++ b/ColActDOnt/.htaccess
@@ -1,0 +1,93 @@
+# Allow all (*) origins to access the resource on the server
+Header set Access-Control-Allow-Origin *
+
+# Allow following symbolic links / symlinks / softlinks
+Options +FollowSymLinks
+
+# Turn off MultiViews
+Options -MultiViews
+
+# Ensure files are served as appropriate content type
+# reminder: https://www.w3.org/2001/12/semweb-fin/swlevels.png
+AddType application/ld+json json
+AddType application/rdf+xml owl rdf
+AddType text/turtle ttl
+
+# Required for the redirects
+RewriteEngine On
+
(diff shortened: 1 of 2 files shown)


Muffexx:
@dgarijo

Thanks for your feedback. As stated above the ontology will be published within the next weeks. As soon as I upload the version to our public repo the liks will "activate". However, I thought I could already establish the permanent identifier.

If it is not possible to add ColActDOnt by now I will (re-)inform you via this PR as soon as the files are up and the sites work.

Best
muffexx

Muffexx:
@dgarijo
I have added dummy html files for version "1.0" and "current". Adding dummy files for the ontologies (ttl, owl, etc.) makes no sense from my point of view.

dgarijo:
I will let @davidlehn have a look on this PR so we can have his view. I am against producing w3ids with broken links.

Muffexx:
Okay, I fully understand. However, I can assure you that this is only a preliminary state. The remaining files will follow in the upcoming weeks.

davidlehn:
We're mostly trying to avoid cluttering the system with unused redirects.  People with the best intentions don't always follow through on project ideas (I'm guilty too), and no one ever removes their old unused files.  I doubt anyone is going to jump in and take `ColActDOnt` if that's the worry. :-)  If you really need this added so links work for testing or whatever, we can do that.  Otherwise we do prefer waiting until links are active.  That also makes it more likely paths and servers and so on are more settled on the target end.  Let us know what you'd prefer.

PR #2704, opened 2022-05-13
Update .htaccess

docuracy:
Please reject this Pull Request if proxy requests are not supported.

If they _are_ supported, would I also need to enable `mod_proxy`, and if so would you please point me to an example of appropriate settings?

Thanks again,
Stephen

### DIFF
--- a/peripleo/.htaccess
+++ b/peripleo/.htaccess
@@ -1,4 +1,9 @@
 RewriteEngine on
 
 # Default
-RewriteRule ^(.*)$ https://britishlibrary.github.io/peripleo-lanc/$1 [NE,R=302,L]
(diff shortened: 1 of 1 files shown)


davidlehn:
No, w3id.org does not support proxy features at this time.  What is your use case?

docuracy:
We're using w3id.org to provide PIDs not only for datasets but for their constituent elements too. Because these datasets are hosted (typically) on GitHub pages, we cannot generate indexable pages for individual elements on-the-fly, so want to index masked URLs instead, using JavaScript to inject relevant content.

davidlehn:
Not sure I quite understand.  But sounds like it may be beyond the scope of w3id.org.  You can always fall back to running your own servers if things are too complex.

davidlehn:
Closing due to unsupported proxy usage.

PR #2712, opened 2022-05-18
Updated .htaccess

luca-secchi:
Fix on the root url for redirects.

### DIFF
--- a/tao/.htaccess
+++ b/tao/.htaccess
@@ -2,6 +2,6 @@ Header set Access-Control-Allow-Origin *
 Options +FollowSymLinks
 RewriteEngine on
 
-SetEnvIf Request_URI ^.*$ ROOT_URL=tourism.documentation.linkalab-cloud.com/tao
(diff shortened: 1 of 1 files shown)


tao/.htaccess
```diff
@@ -2,6 +2,6 @@ Header set Access-Control-Allow-Origin *
 Options +FollowSymLinks
 RewriteEngine on
 
-SetEnvIf Request_URI ^.*$ ROOT_URL=tourism.documentation.linkalab-cloud.com/tao
+SetEnvIf Request_URI ^.*$ ROOT_URL=http://tourism.documentation.linkalab-cloud.com/tao
```
dgarijo, line 5: Is it possible to use https? Otherwise with w3id it will go through https -> http, which has problems in some applications

luca-secchi: We don't have https in place right now but would soon move the files to GitHub

PR #2734, opened 2022-06-05
Update .htaccess

Haseebshahzad052:


dgarijo:
You are trying to edit a w3id for which you are not listed as a maintainer. Plus, this is not a valid content negotiation file. Please see https://github.com/perma-id/w3id.org#creating-a-new-identifier

PR #2759, opened 2022-06-20
Update .htaccess

A1ib:


dgarijo:
This is removing the main htaccess, which is incorrect.
Please have a look at https://github.com/perma-id/w3id.org#suitable-pr-content to see what constitutes suitable PR content. You can find an example here: https://github.com/perma-id/w3id.org/tree/master/example.
Since this PR is not valid, I will close it. Please open a new one

PR #2763, opened 2022-06-22
Add Permanent URI To Ontology

sehrishnoreen31:


dgarijo:
@sehrishnoreen31,
you are removing the main htaccess, which is wrong. Please create a folder with your w3id and put your htaccess file inside. You can see an example here: https://github.com/perma-id/w3id.org/tree/master/example

PR #2795, opened 2022-07-08
Update .htaccess

jhaag75:
Update to support new ADL profiles server's restricted IRIs. Pattern matching of IRIs beyond the base profile IRI is no longer supported with the new profile server.


davidlehn:
Please fix conflicts.

PR #2798, opened 2022-07-11
Lexmeta redirection rules

gkirtzou:


gkirtzou:
What? How? Could you please clarify how the /def is being added?

gkirtzou:
Damn it! I didn't notice it. Thanks! I will fix it!

PR #2809, opened 2022-07-20
Update .htaccess

laprus:


lemoustachiste:
@dgarijo it's ok, i actually asked @laprus to handle that for me.
For now it's a one-off so I don't see the need to add him to the list of contributors, but I am aware of that change.

Thanks

PR #2825, opened 2022-08-02
[exdoc] Add exdoc ontology

philhag:


dgarijo:
@philhag,
please add your github user as part of the maintainers, so it's easy for us to confirm who you are if future edits are required.
Also, we recommend not using  raw.githubusercontent links. Please consider using [github pages](https://pages.github.com/) instead

philhag:
thanks for you comment @dgarijo, i changed the redirects to the respective github pages and added link to github user profile

PR #2826, opened 2022-08-02
Updated htaccess for fixing redirection

luca-martinelli-09:


davidlehn:
Can you fix this so there is just one commit instead of 25 for a one line change?  Probably can just rebase off an updated master branch.

luca-martinelli-09:
> Can you fix this so there is just one commit instead of 25 for a one line change? Probably can just rebase off an updated master branch.

I did it but it seems didn't work

PR #2832, opened 2022-08-03
ee4i initial version

jdmarinj:


ee4i/.htaccess
```diff
@@ -0,0 +1,47 @@
...
+# Rewrite rule to serve HTML content from the vocabulary URI if requested
+RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+xml)
+RewriteCond %{HTTP_ACCEPT} text/html [OR]
+RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
+RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
+RewriteRule ^$ index-en.html [R=303,L]
+
+# Rewrite rule to serve JSON-LD content from the vocabulary URI if requested
+RewriteCond %{HTTP_ACCEPT} application/ld\+json
+RewriteRule ^$ ontology.jsonld [R=303,L]
```
dgarijo, line 25: @jdmarinj,
you are redirecting to files that do not exist. You should host the files in a Github repository (or elsewhere) and then redirect them here. See the `example` repository for some insight. In short, you should change `ontology.jsonld` (and the other files) to https://YOUR_URL/ontology.jsonld

dgarijo:
@jdmarinj see the comment I left in your PR. As it is, it won't work. Also, please add your github id to your readme in case you make other changes in the future. Thanks!

ee4i/.htaccess
jdmarinj: Hi, Thanks for your help. I just did it

dgarijo:
@jdmarinj you are almost there. Now you are serving the HTML of the pages where you are hosting your files. Instead, you should have a "github pages" project enabled (see https://pages.github.com/) to serve your content instead of the html. So instead of `https://github.com/jdmarinj/ee4i/blob/main/index-en.html` you want to have `https://jdmarinj.github.io/ee4i/main/index-en.html` or similar.

jdmarinj:
@dgarijo Please ignore Update V2.htaccess and check  **Update V3.htaccess**

stain:
Hi, you should also update each of the content-negotiation bits for RDF, for instance:

```
RewriteCond %{HTTP_ACCEPT} application/n-triples
RewriteRule ^$ https://github.com/jdmarinj/ee4i/blob/main/ontology.nt [R=303,L]
```

will redirect to https://github.com/jdmarinj/ee4i/blob/main/ontology.nt - but that is an HTML page about the N-Triples file, rather than the N-Triples file, and so would not parse in an RDF client. 

Use instead in these as well the github.io links like https://jdmarinj.github.io/ee4i/ontology.nt

```
RewriteCond %{HTTP_ACCEPT} application/n-triples
RewriteRule ^$ https://jdmarinj.github.io/ee4i/ontology.nt [R=303,L]
```

PR #2843, opened 2022-08-05
AllowEncodedSlashes conf, NoEscaping Rewrite flag.

awagner-mainz:
Trying to get redirection of requests where the path contains urlencoded URLs to work. Presumably, the problem occurs when encountering encoded forward slashes, as `curl -I https://w3id.org/mpilhlt/reconcile/https%3A` is working correctly (gives a 307-redirection), whereas `curl -I https://w3id.org/mpilhlt/reconcile/https%3A%2F` does not (gives a 404).

### DIFF
--- a/mpilhlt/.htaccess
+++ b/mpilhlt/.htaccess
@@ -1,5 +1,6 @@
 Options +FollowSymLinks
 RewriteEngine on
+AllowEncodedSlashes On
 
 # Vocabularies webpages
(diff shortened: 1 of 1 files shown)


awagner-mainz:
read about the Nodecode value for the AllowEncodedSlashes conf too late. Will close this PR and fix my fork, then re-submit a PR.

PR #2844, opened 2022-08-05
AllowEncodedSlashed conf, NoEscape Rewrite flag.

awagner-mainz:
Trying to fix redirecting requests that contain full urlencoded URLs in their path.


awagner-mainz:
read about the Nodecode value for the AllowEncodedSlashes conf too late. Will close this PR and fix my fork, then re-submit a PR.

PR #2861, opened 2022-08-11
Update to Qual-o namespace

dcorsar:
The Qual-O ontology was defined with a namespace which is no longer accessible; this update changes the namespace to w3id to simplify reuse of the ontology.

### DIFF
--- a/qual-o/.htaccess
+++ b/qual-o/.htaccess
@@ -1,0 +1,7 @@
+Options +FollowSymLinks
+
+# Rewrite engine setup
+RewriteEngine On
+
+# Redirect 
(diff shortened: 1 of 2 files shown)


dgarijo:
@dcorsar since you are not the author, would it be possible to get the OK from @cbaillie? 
Maybe the ontology is just temporarily down and they would prefer to handle the CN themselves (plus, the URI change)

dcorsar:
Of course; I'll message Chris to get him to give the ok.
The server that did host the ontology was taken down after he completed his PhD, so this update needs to be made anyway.

dgarijo:
Pinging @dcorsar on this PR.

dgarijo:
Closing this PR due to inactivity. @dcorsar if you want it merged please add yourself as a maintainer and reopen it.

PR #2865, opened 2022-08-15
Update semantic question ontology links from v1.0.0 to v1.1.0

hvarg:
Updating links to the web-page and ontologies


hvarg:
Can you check it @dgarijo ?

dgarijo:
If you want to support previous versions of the ontology (which I recommend) you should have a look at https://github.com/perma-id/w3id.org/blob/master/example/.htaccess (see there is latest version vs any other version)

PR #2868, opened 2022-08-22
Minor change in rewriting

ferag:


ferag:
Small change for rewrite rule

PR #2887, opened 2022-09-01
Update URLs to BMO ontology

JYL61:
According to the tutorial "Permanent Identifiers for the Web", I would like to update a permanent identifier of the form https://w3id.org/... on the BMO ontology.


dgarijo:
@JYL61,
we don't host ontologies at w3id, you have to host them in your own repo. See https://github.com/perma-id/w3id.org/tree/master/example for an example

JYL61:
OK. Thank you~

dgarijo:
Closing this PR as it's incorrect. Please submit another one following the suggested example

PR #2896, opened 2022-09-07
new rules

rapw3k:
Hi @dgarijo 
I have a question. 
I need to make a very simple arithmetic operation during the redirection, e.g., 

[https://w3id.org/iliad/api/v1.0/Observations?$top=100&$skip=100](https://w3id.org/iliad/api/v1.0/Observations?$top=100&$skip=100)

 should be  redirected to 

[https://grlc.io/api-git/ILIAD-ocean-twin/api/Observations?page=2](https://grlc.io/api-git/ILIAD-ocean-twin/api/Observations?page=2)

i made rule 
```
RewriteRule ^api/v1.0/Observations\?\$top=(.*)&\$skip=(.*)00$ https://grlc.io/api-git/ILIAD-ocean-twin/api/Observations?page=$2 [R=302,L]
```
(basically with regex i get from skip, 1 from 100,  2 from 200, 3 from 300,....10 from 1000, etc. and then I add 1.

But i dont know how to add 1 to $2 (skip value)

I found it should be possible: https://stackoverflow.com/questions/20164428/how-to-do-arithmetic-operation-in-rewriterule-function  (see also https://httpd.apache.org/docs/2.4/rewrite/rewritemap.html#prg)
but i dont see any example in w3id.org repo

could you please check if this would be possible in w3id ?

thanks
Raul

### DIFF
--- a/iliad/.htaccess
+++ b/iliad/.htaccess
@@ -33,3 +33,4 @@ RewriteRule ^api/v1.0/Observations\((.*)\)$ https://grlc.io/api-git/ILIAD-ocean-
 
 
 
+
--- a/iliad/.htaccess
+++ b/iliad/.htaccess
@@ -32,3 +32,4 @@ RewriteRule ^api/v1.0/Observations\((.*)\)$ https://grlc.io/api-git/ILIAD-ocean-
 
 
 
+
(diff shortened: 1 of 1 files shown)


davidlehn:
While it may be possible to invoke external processes to do special processing, I don't think that will be enabled or allowed anytime soon.  It's beyond the scope of the simple redirection scope of this service.  And there unknown potential DoS problems and other security issues.  I'm not sure what other "math" solutions there are.  I'd suggest passing on all the parameters and implementing the custom logic you need on your grlc.io server, or somewhere else that redirects again to the target server.

rapw3k:
thanks @davidlehn; 
My problem is that is not possible for me to handle on grlc side if i want to expose an API that is 100% compliant with OGC SensorThings API (e.g., https://developers.sensorup.com/docs/#observations_get). 

The API requires the top and skip values in ["@iot.nextLink"], and grlc on the other hand is just a SPARQL behind the scenes. I can send parameter for page to specify from which result to start the response, but I cannot send as parameter the "LIMIT" and "OFFSET"  of the SPARQL query.

I know in this case 
top =100 & skip=100 is page=2
top =100 & skip=200 is page=3
....

I could only provide an incorrect skip value to get
top =100 & skip=200 is page=2 (but in fact this is entries 101-200) 

but not sure any other way of handling this :(

rapw3k:
in any case if all is correct in this PR, please merge!
thanks in advance

PR #2897, opened 2022-09-08
Rename of ontology

michaelfaerber:


dgarijo:
@michaelfaerber, are you sure you don't want to keep the old w3id with a simple redirect to the new one?

PR #2898, opened 2022-09-08
Update rule

rapw3k:
Hi, 
My last commit didnt work as expected. I had these 2 rules:
```
RewriteRule ^api/v1.0/Observations$ https://grlc.io/api-git/ILIAD-ocean-twin/api/Observations?page=1 [R=302,L]
RewriteRule ^api/v1.0/Observations\?\$top=(.*)&\$skip=(.*)00$ https://grlc.io/api-git/ILIAD-ocean-twin/api/Observations?page=$2 [R=302,L]
```
However this URL:
https://w3id.org/iliad/api/v1.0/Observations?$top=100&$skip=300
was redirected to: 
https://grlc.io/api-git/ILIAD-ocean-twin/api/Observations?page=1
 
using the first rule. Not sure why. Do you have any idea?

I am trying now by changing the order of the rules (as below), but not sure if something is not good in the rules? Any help will be appreciated 

```
RewriteRule ^api/v1.0/Observations\?\$top=(.*)&\$skip=(.*)00$ https://grlc.io/api-git/ILIAD-ocean-twin/api/Observations?page=$2 [R=302,L]
RewriteRule ^api/v1.0/Observations$ https://grlc.io/api-git/ILIAD-ocean-twin/api/Observations?page=1 [R=302,L]
```

PR #2899, opened 2022-09-08
Update rule

rapw3k:
I think I found the correct form (testing in https://htaccess.madewithlove.com/ )


davidlehn:
This will only work if both those params are present and in that order.  See my [other comment](https://github.com/perma-id/w3id.org/pull/2898#issuecomment-1241243409) for an alternate approach if you need it.

PR #2914, opened 2022-09-20
updated frost ontology and readme

gatemezing:
Added a new entry for an ontology and subsequent resources

### DIFF
--- a/frost/.htaccess
+++ b/frost/.htaccess
@@ -1,0 +1,70 @@
+RewriteEngine on
+
+# Turn off MultiViews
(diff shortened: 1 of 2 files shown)


dgarijo:
@gatemezing ,
you are creating the w3id: https://w3id.org/frost/frost. Is this what you intend? (note the double 'frost')

davidlehn:
@dgarijo I think this is ok?  Looks like rules are anchored to the dir or are files with extensions.

dgarijo:
You are right.

PR #2918, opened 2022-09-27
feat(sumo): add sumo kb namespace

noelmcloughlin:
This PR reserves sudo namespace for SUMO Ontology (Kb, WordNet).


sumo/README.md
```diff
@@ -0,0 +1,17 @@
+# /sumo/
+
+The Suggested Upper Merged Ontology (SUMO) and its domain ontologies form the largest formal public ontology in existence today. They are being used for research and applications in search, linguistics and reasoning. SUMO is the only formal ontology that has been mapped to all of the WordNet lexicon. Sumo is maintained by https://www.ontologyportal.org
```
TallTed, line 3:
```suggestion
The Suggested Upper Merged Ontology (SUMO) and its domain ontologies form the largest collective formal public ontology in existence today. It is being used for research and applications in search, linguistics, and reasoning. SUMO is the only formal ontology that has been mapped to the entire WordNet lexicon. Sumo is maintained by <https://www.ontologyportal.org>.
```

sumo/wordnet
```diff
@@ -0,0 +1 @@
+wn
```
davidlehn, line 1: I think this is a symlink?  (github makes it hard to tell)  This would be the first and only symlink in use in this project.  (Despite the cargo cult `Options` used everywhere)  I think it would work, but maybe just using `.htaccess` rules is easier to understand than using file system features.  Since this is setting up new redirects, and the readme even says to use `wn`, does this name and/or symlink style need to be used?

noelmcloughlin: Hi @davidlehn  thanks for the comment.  The symlink is not actually needed since this is new W3ID. And I'd prefer to follow existing practices.  I have updated commit to remove the unnecesary symlink. thanks

PR #2939, opened 2022-10-10
Added new redirect rules for jams ontology

andreamust:


davidlehn:
Those `smashub` URLs are 404.  Are those tokens supposed to be secret?

PR #2941, opened 2022-10-10
Update .htaccess

physta:


davidlehn:
- Why do you need the symlink option?  Why did you list it twice?
- Why do you need the rewrite option again that is below?

physta:
Hi, the rewrite is an error. About the symlink, not sure about it. I'm just trying to make my IRI paths to point correctly inside of the ontology. What I mean is that if I copy for example the IRI https://w3id.org/frontier/PersonalDevice in the browser I go directly to the top of my ontology (https://physta.github.io/frontier/OnToology/Frontier.ttl/documentation/doc/index-en.html), but not to the description itself. How I need to modify the htaccess file to make this happend? Also, I've seen in several ontologies hosted here, for example https://comply.cispe.cloud/cispe-definitions/compliance/index-en.html , that they have a WebVOWL visualization included. Do you know how I can include that? Thank you very much in advance.

dgarijo:
@physta,
the WebVowl visualization is a flag inside the documentation tool you are using. It has nothing to do with w3id :)

physta:
Ok thanks. What about my other question:
I'm trying to make my IRI paths to point correctly inside of the ontology. What I mean is that if I copy for example the IRI https://w3id.org/frontier/PersonalDevice in the browser I go directly to the top of my ontology (https://physta.github.io/frontier/OnToology/Frontier.ttl/documentation/doc/index-en.html), but not to the description itself. How I need to modify the htaccess file to make this happend?

dgarijo:
You need to enable hash redirections. See https://www.w3.org/TR/swbp-vocab-pub/#recipe3 for some standard recommendations

frontier/.htaccess
```diff
@@ -1,3 +1,6 @@
+Options +FollowSymLinks
+RewriteEngine on
+Options +FollowSymLinks
```
TallTed, line 3: It seems odd for lines 1 and 3 to be identical...

PR #2948, opened 2022-10-13
Change redirect for publiccode

bfabio:


bfabio:
@dgarijo thanks

cc @ruphy

ruphy:
+1 from me. Thanks @dgarijo for looking after this!

PR #2956, opened 2022-10-19
Fix error in semapv redirect

matentzn:
Sorry about that!

### DIFF
--- a/semapv/.htaccess
+++ b/semapv/.htaccess
@@ -2,4 +2,4 @@ Options +FollowSymLinks
 RewriteEngine on
 
 RewriteRule ^$ https://github.com/mapping-commons/semantic-mapping-vocabulary [R=302]
-RewriteRule ^vocab/?$ https://mapping-commons.github.io/semantic-mapping-vocabulary/# [R=302]
+RewriteRule ^vocab\/(.*)$ https://mapping-commons.github.io/semantic-mapping-vocabulary/# [R=302]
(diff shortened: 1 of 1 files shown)


semapv/.htaccess
```diff
@@ -2,4 +2,4 @@ Options +FollowSymLinks
 RewriteEngine on
 
 RewriteRule ^$ https://github.com/mapping-commons/semantic-mapping-vocabulary [R=302]
-RewriteRule ^vocab/?$ https://mapping-commons.github.io/semantic-mapping-vocabulary/# [R=302]
+RewriteRule ^vocab\/(.*)$ https://mapping-commons.github.io/semantic-mapping-vocabulary/# [R=302]
```
dgarijo, line 5: If you do ^vocab/(.*) you may get the result you are looking for. You don't need to escape slashes

matentzn:
@dgarijo thank you for your help before, something still does not seem to work.. for example

https://w3id.org/semapv/vocab/LinkStripping

now redirects to `https://mapping-commons.github.io/semantic-mapping-vocabulary/%23`.. Any idea what needs changing to have it redirect to `https://mapping-commons.github.io/semantic-mapping-vocabulary/#LinkStripping`?

dgarijo:
It's urlencoding `#` because you put it after the `/`
I guess you will have to do something like: `https://mapping-commons.github.io/semantic-mapping-vocabulary#` if you want to refer to the page in general, and `https://mapping-commons.github.io/semantic-mapping-vocabulary#$1` if you want to go to that anchor.
However, the latter will not work all the time because your anchors do not correspond to your ontology names. For example `backgroundknowledge-basedmatchingprocess` is different from `BackgroundKnowledgeBasedMatching`, used in your ontology

PR #2957, opened 2022-10-20
Update .htaccess

matentzn:
cc @dgarijo

Is this correct then? 

> However, the latter will not work all the time because your anchors do not correspond to your ontology names. For example backgroundknowledge-basedmatchingprocess is different from BackgroundKnowledgeBasedMatching, used in your ontology

Don't worry about that, this has to be addressed by the [developers of the vocabulary documentation framework](https://github.com/RDFLib/pyLODE/issues/86).

### DIFF
--- a/semapv/.htaccess
+++ b/semapv/.htaccess
@@ -2,4 +2,4 @@ Options +FollowSymLinks
 RewriteEngine on
 
 RewriteRule ^$ https://github.com/mapping-commons/semantic-mapping-vocabulary [R=302]
-RewriteRule ^vocab/(.*) https://mapping-commons.github.io/semantic-mapping-vocabulary/# [R=302]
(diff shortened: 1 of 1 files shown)


dgarijo:
the solution (with the corresponding caveat) is the one implemented in https://github.com/dgarijo/Widoco/

matentzn:
@dgarijo -- oooo awesome, I didnt realise you were involved with LODE and ontology documentation! Thank you for pointing that out!

dgarijo:
no worries! We also have https://ontoology.linkeddata.es/, an open service which takes care of everything and issues the doc and other things (diagrams, reports) in a pull request, and takes care of the publication through w3id if you want. But since you have already produced this manually you may not need it for this case. Maybe next time you create an ontology.

matentzn:
The redirect still does not work..



https://w3id.org/semapv/vocab/MappingActivity

------->

https://mapping-commons.github.io/semantic-mapping-vocabulary%23MappingActivity

Will check your resources!

PR #2958, opened 2022-10-20
Update .htaccess for semapv

matentzn:
cc @dgarijo 

Lol sorry I should have done an htaccess tutorial. Looking at Ontoology as well now, looks cool! Trying to find an example deployed somewhere on the web to see how it looks like, for example when deployed with github.io.

### DIFF
--- a/semapv/.htaccess
+++ b/semapv/.htaccess
@@ -2,4 +2,4 @@ Options +FollowSymLinks
 RewriteEngine on
 
 RewriteRule ^$ https://github.com/mapping-commons/semantic-mapping-vocabulary [R=302]
-RewriteRule ^vocab/(.*) https://mapping-commons.github.io/semantic-mapping-vocabulary#$1 [R=302]
+RewriteRule ^vocab/(.*) https://mapping-commons.github.io/semantic-mapping-vocabulary# [R=302]


matentzn:
Thank you @dgarijo, unfortunately it still does not work..

https://w3id.org/semapv/vocab/BlankNormalisation

Is there any way I can test an htaccess redirect before deploying?

dgarijo:
My bad, remove the `#` at the end

dgarijo:
And yes, you can set up a local server (e.g., with Apache) and do local tests.

dgarijo:
as per your question, a test example: https://github.com/dgarijo/test/tree/master/OnToology/test.owl  is published at https://w3id.org/def/test through OntToology. Note that the URLs used in the documentation itself are of another ontology, it's just for testing purposes.

matentzn:
Awesome, thanks for the example! 

I am not a web developer, I was hoping for something like an an online htaccess tester. 

I don't know now, I don't want to do more trial and error and annoy the other people listening to this repo. How can I be certain that 

```
RewriteRule ^vocab/(.*) https://mapping-commons.github.io/semantic-mapping-vocabulary [R=302]
```

will cause 

`https://w3id.org/semapv/vocab/BlankNormalisation`

to redirect to

`https://mapping-commons.github.io/semantic-mapping-vocabulary#BlankNormalisation`

If I remove the hash, wont it just remove to redirect to

```
https://mapping-commons.github.io/semantic-mapping-vocabulary
```

?

dgarijo:
yes, if you remove the  hash it will redirect to the page only. 
To do the former, you need to have `$1` as we first discussed, but some of your anchors won't work (leaving you at the beginning of the page)

matentzn:
Ahh ok.. Got it! Thanks. seems like I would also have to rewrite to lower case.. Another time :) Thank you for your help!

PR #2959, opened 2022-10-20
Hopefully last attempt to fix semapv redirect

matentzn:
I am trying to model the solution after https://httpd.apache.org/docs/2.2/rewrite/flags.html#flag_ne

cc @dgarijo


dgarijo:
This will make some of your terms not work still because the problem is how the anchors are created. So https://w3id.org/semapv/vocab/BlankNormalisation will remain at the top of the page instead of going to https://mapping-commons.github.io/semantic-mapping-vocabulary/#blanknormalisation

dgarijo:
But at least your doc works. I should have mentioned the `NE` bit before, I forgot, sorry about that

PR #2961, opened 2022-10-21
Update idsa context paths from jira to gitlab.

pnlinden:
We moved the hosting of a set of resources from jira to gitlab and would like to update the paths accordingly.


davidlehn:
@pnlinden Can you get one of the other contacts listed in the README to approve this change?  Thanks.

JohannesLipp (approved):
Thank you for asking for review!

PR #2970, opened 2022-10-26
create serdif permanent identifier as w3id.org

navarral:


serdif/README.md
```diff
@@ -0,0 +1,20 @@
+# serdif
+This [W3ID](https://w3id.org) provides a persistent URI namespace for Semantic Environmental and Rare Disease data Integration Framework (SERDIF) resources.
+
+## Uses
+This namespace is broken down into different parts for SERDIF components, sparql endpoint (sparql), user interface (ui) and others.
```
TallTed, line 5:
```suggestion
This namespace is broken down into different parts for SERDIF components, SPARQL endpoint (SPARQL), user interface (UI) and others. 
```

davidlehn:
Missing a `.htaccess` file with the redirects.

serdif/.htaccess
```diff
@@ -0,0 +1,2 @@
+RewriteEngine on
+RewriteRule (.*) https://github.com/navarral/serdif/$1 [R=301,L]
```
davidlehn, line 2: Note this is going to the GitHub UI for those resources.  I see there's nothing there now.  You may consider some other hosting, like github pages, if you want to see the raw resources.

navarral: Thanks for the suggestion. I will consider using GitHub pages. I generated the permanent URL link in advance since I have to include it in a journal paper by the end of the week. I will be adding the content during the next days. Thanks for your guidance.

PR #2972, opened 2022-10-27
Update Mach30 permid .htaccess

j-simmons-phd:
Updated rewrite rule to work for any URI under /Mach30


Mach30/.htaccess
```diff
@@ -2,4 +2,4 @@
 RewriteEngine On
 
 RewriteCond %{HTTP:Accept} text/html [NC]
-RewriteRule ^$ https://github.com/Mach30/ [R=303,L]
+RewriteRule ^Mach30\/(.*)$ https://github.com/Mach30/$1 [R=303,L]
```
davidlehn, line 5: Patterns are relative to current directory.  This matches `https://w3id.org/Mach30/Mach30/...`.  Perhaps you want:
```suggestion
RewriteRule ^(.*)$ https://github.com/Mach30/$1 [R=303,L]
```

j-simmons-phd: Thanks.  I was trying to test it in a .htaccess tester and it probably did not account for the current directory.

PR #2976, opened 2022-11-02
new redirect for CEMontology

federicarollo:


CEMontology/.htaccess
```diff
@@ -0,0 +1,3 @@
+RewriteEngine On
+
+RewriteRule (.*)$ http://crimeanalytics.it/cem-reference/doc/index-en.html$1 [R=302,L,QSA]
```
dgarijo, line 3: I don't think this will work like you intend it. Have you tried it locally? You are trying to convert slashes to hash anchors, so you'll probably need to do something like adding a hash:
`RewriteRule (.*)$ http://crimeanalytics.it/cem-reference/doc/index-en.html#$1 [R=302,L,QSA]` (note the `#`)

dgarijo:
@federicarollo please have a look at the comment whenever you have the chance

federicarollo:
sorry, I just want to obtain an identifier for the ontology available at http://crimeanalytics.it/cem-reference/doc/index-en.html

How do I need to modify the .htaccess file?

dgarijo:
See my comment, I added the required modification for the redirect to work

federicarollo:
ok, I have modified the file

PR #2978, opened 2022-11-03
Update .htaccess

federicarollo:


dgarijo:
You need to add $, or your regular expression will not work. You need to do 
`RewriteRule (.*)$ http://crimeanalytics.it/cem-reference/doc/index-en.html [R=302,L,QSA]`

PR #2980, opened 2022-11-07
update redirection for gaia-x schema

ticapix:
redirect to the official source of reference for gaia-x schema

### DIFF
--- a/gaia-x/.htaccess
+++ b/gaia-x/.htaccess
@@ -14,7 +14,4 @@ Options +FollowSymLinks
 RewriteEngine on
 
 # BASE: Visit homepage
-RewriteEngine on
-
-# BASE: Visit homepage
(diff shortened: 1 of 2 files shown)


ticapix:
Hi @JohannesLipp ,

Could you add me as maintainer ?

Thanks
Pierre

JohannesLipp:
Will speak to my team and accept this ASAP. Should not take too long ✌️

clange (approved):
In advance of the Gaia-X Summit, it is necessary to establish a unified governance of Gaia-X assets in the Gaia-X Registry. This may temporarily break the redirects to the individual assets (ontology, generated HTML documentation), but we will work together on restoring them via the Registry. Resources maintained outside of the scope of the Gaia-X Association, its Committees and Working Groups will be published (and redirected to) in other places.

PR #2998, opened 2022-11-14
agvt

Ahmedreyadh1985sh:


davidlehn:
This PR is editing root files and adding non-redirect rule files.  Please use a subdirectory for your id path and follow other examples for adding redirects.  Closing since a new PR is likely easier than fixing history on this one.  Thanks.

PR #3003, opened 2022-11-17
updated .htaccess

wterkaj:
changed repository

### DIFF
--- a/ontoeng/.htaccess
+++ b/ontoeng/.htaccess
@@ -10,11 +10,20 @@ AddType text/turtle .ttl
 
 RewriteEngine on
 
-#Rewrite rules for the SMO vocabulary  
+#Rewrite rules
 RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+xml|text/\*|\*/\*)
 RewriteCond %{HTTP_ACCEPT} text/html [OR]
 RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
 RewriteCond %{HTTP_ACCEPT} text/\* [OR]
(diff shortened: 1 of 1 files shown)


ontoeng/.htaccess
```diff
@@ -10,11 +10,22 @@ AddType text/turtle .ttl
...
+#Rewrite rules
 RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+xml|text/\*|\*/\*)
 RewriteCond %{HTTP_ACCEPT} text/html [OR]
 RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
 RewriteCond %{HTTP_ACCEPT} text/\* [OR]
 RewriteCond %{HTTP_ACCEPT} \*/\* [OR]
 RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
-RewriteRule ^(.*)$ http://www.ontoeng.com/$1  [R=302,L]
+RewriteCond %{REQUEST_URI} !.owl [NC]
+RewriteRule ^(.*)/$ https://difactory.github.io/repository/ontoeng/$1.owl  [R=302,L]
```
davidlehn, line 21: Is the duplication below added to handle an optional trailing slash?  Consider this instead:
```suggestion
RewriteRule ^(.*)/?$ https://difactory.github.io/repository/ontoeng/$1.owl  [R=302,L]
```

wterkaj: Thanks, I'll apply the suggestion. I had tested this option on another server, but it didn't work as expected. Anyhow, I've seen that it is used in other .htaccess files here in w3id.org, so I'll go with it

PR #3006, opened 2022-11-23
Add redirect `orcidio` for ORCID in OWL

cthoyt:
ORCID in OWL is an ontology that wraps notable ORCID identifiers as named individuals so they can be more easily included in other ontologies (e.g., OBO Ontologies) and empower Protege to show rich label information for `dc:contributor` annotations.

The goal of having an entry in w3id.org is to endow this ontology with a PURL as its ontology IRI.

### DIFF
--- a/orcidio/.htaccess
+++ b/orcidio/.htaccess
@@ -1,0 +1,6 @@
+Options +FollowSymLinks
+RewriteEngine on
+
+# Make ontology artifacts resolvable
+RewriteRule ^orcidio.ofn$ https://raw.githubusercontent.com/cthoyt/orcidio/main/orcidio.ofn [R=302,L]
+RewriteRule ^orcidio.owl$ https://raw.githubusercontent.com/cthoyt/orcidio/main/orcidio.owl [R=302,L]
--- a/orcidio/README.md
+++ b/orcidio/README.md
@@ -1,0 +1,16 @@
+# ORCID in OWL
+
+This [W3ID](https://w3id.org/) provides a persistent URI namespace for ORCID in OWL resources.
+
+## Uses
+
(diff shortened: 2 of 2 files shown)


orcidio/.htaccess
```diff
@@ -0,0 +1,6 @@
+Options +FollowSymLinks
+RewriteEngine on
+
+# Make ontology artifacts resolvable
+RewriteRule ^orcidio.ofn$ https://raw.githubusercontent.com/cthoyt/orcidio/main/orcidio.ofn [R=302,L]
+RewriteRule ^orcidio.owl$ https://raw.githubusercontent.com/cthoyt/orcidio/main/orcidio.owl [R=302,L]
```
dgarijo, line 6: I don't think this is what you intent. This would mean that only the URL `https://w3id.org/orcidio/orcidio.owl` works instead of having `https://w3id.org/orcidio` redirect to your ontology directly

dgarijo:
Please see my comment above

orcidio/.htaccess
cthoyt: Hi @dgarijo, thank you for the quick comment. I would like https://w3id.org/orcidio/orcidio.owl to redirect to https://raw.githubusercontent.com/cthoyt/orcidio/main/orcidio.owl (and similarly for the `.ofn` file). I admit there's a bit of redundancy, but I think inside `https://w3id.org/orcidio/` I will add future alternatives like `https://w3id.org/orcidio/orcidio-obo.owl`, `https://w3id.org/orcidio/orcidio-chemistry.owl`. Have I misunderstood your comment?

dgarijo: ok

cthoyt:
Thank you again @dgarijo :)

Issue #2934, opened 2022-10-06
Slashes and HTTPS redirection


IllidanS4:
I have two general concerns regarding redirection, which (as far as I can tell) are caused by the server running w3id.org and thus are not directly fixable in this codebase:

* Apache handles multiple or encoded slashes in the URI in a special fashion, but this is not desirable for a redirection service. To fix this, these directives have to be used *in the Apache (virtual host) configuration*, not in `.htaccess`:
  ```
  AllowEncodedSlashes On
  MergeSlashes Off
  ```
  Since the target server will most likely have the default behaviour anyway, this is not a security issue. I am considering moving from purl.org to w3id.org, and this is something I would need (purl.org doesn't even merge slashes, it simply outputs "Internal Server Error" and they haven't fixed it yet since I told them months ago).

* w3id seems to redirect HTTP to HTTPS in all situations. I don't think it is necessary, and it doesn't bring much to the security: a potential attacker could hijack the first HTTP request and find out the path, and then determine perfectly where it redirects since this repository is public. Redirecting to HTTPS won't fix anything since the first (any only) request is already insecure.

  I think it would be best to simply honor whichever protocol the client wishes to use, and let it be the concern of the actual redirecting `.htaccess` or the target server to do otherwise. Or, if that is considered inappropriate, you could [decide based on the presence of the `Upgrade-Insecure-Requests` header](https://is4code.blogspot.com/2021/07/https-redirection-in-apache-http-server.html) which is sent by modern browsers anyway, and do not redirect if it is not present (under the assumption that the client wishes to use and continue using HTTP in that case).

  This is not critical to me, but it seems like a good thing to do. TLS versions come and go, and at some point in the future, old TLS versions could stop being supported by this service, making these "persistent" URIs break for clients that can't use the newer version, and using HTTP is more likely to continue working. After all, requiring HTTPS is again the matter of the target server.

davidlehn:
> * Apache handles multiple or encoded slashes in the URI in a special fashion, but this is not desirable for a redirection service. To fix this, these directives have to be used _in the Apache (virtual host) configuration_, not in `.htaccess`:
>   ```
>   AllowEncodedSlashes On
>   MergeSlashes Off
>   ```
> [...]

What are the use cases you are concerned with?  What are some example URLs that would be problematic?

There has been a past issue or two with encoded slashes, but I think those were related to trying to put URLs in a path and having regex issues and they just moved to using query params.

[...]

PR #3010, opened 2022-11-26
uri4uri namespace

IllidanS4:
Contains a workaround for `MergeSlashes On`, but due to `AllowEncodedSlashes Off`, `%2F` cannot be used in an identifier (see #2934). This is a potential issue only for `/uri/`, but the normalized identifiers never use `%2F` in path anyway (it's either `/` or doubly-encoded `%252F`) so it's not an issue if the URI is produced accordingly.

### DIFF
--- a/uri4uri/.htaccess
+++ b/uri4uri/.htaccess
@@ -1,0 +1,9 @@
+RewriteEngine On
+RewriteOptions AllowAnyURI
+
+# MergeSlashes workaround - get the original path from THE_REQUEST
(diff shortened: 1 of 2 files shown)


dgarijo:
@IllidanS4, please add yourself as a maintainer as well in this repository.

IllidanS4:
@dgarijo It's there; I use IS4 wherever possible (can't change it on GitHub sadly since it's already taken).

dgarijo:
ok, apologies

IllidanS4:
No worries, thanks for the merge!

PR #3032, opened 2022-12-14
Update .htaccess

rapw3k:


iliad/.htaccess
```diff
@@ -19,33 +19,74 @@ RewriteRule ^oim$ https://github.com/ILIAD-ocean-twin/oim [R=302,L]
 
 ### API
 # Base
-RewriteRule ^api$ https://grlc.io/api-git/ILIAD-ocean-twin/api/base [R=302,L]
-RewriteRule ^api/v1.0$ https://grlc.io/api-git/ILIAD-ocean-twin/api/base [R=302,L]
-RewriteRule ^api/v1.0/$ https://grlc.io/api-git/ILIAD-ocean-twin/api/base [R=302,L]
+RewriteRule ^api$ https://dev.grlc.io/api-git/ILIAD-ocean-twin/JF-API/base [R=302,L]
+RewriteRule ^api/v1.0$ https://dev.grlc.io/api-git/ILIAD-ocean-twin/JF-API/base [R=302,L]
+RewriteRule ^api/v1.0/$ https://dev.grlc.io/api-git/ILIAD-ocean-twin/JF-API/base [R=302,L]
```
davidlehn, line 24: You can simplify the optional trailing slash as `/?`.
```suggestion
RewriteRule ^api/v1.0/?$ https://dev.grlc.io/api-git/ILIAD-ocean-twin/JF-API/base [R=302,L]
```

iliad/.htaccess
```diff
@@ -19,33 +19,74 @@ RewriteRule ^oim$ https://github.com/ILIAD-ocean-twin/oim [R=302,L]
...
 # Observations
 #RewriteCond %{QUERY_STRING} ^(?:\$|%24)top=(.*)&(?:\$|%24)skip=(.*)00$
 RewriteCond %{QUERY_STRING} (?:^|&)(?:\$|%24)skip=([1-9]+[0-9]*)00
-RewriteRule ^api/v1.0/Observations$ https://grlc.io/api-git/ILIAD-ocean-twin/api/Observations?page=%1 [R=302,L]
-RewriteRule ^api/v1.0/Observations$ https://grlc.io/api-git/ILIAD-ocean-twin/api/Observations?page=1 [R=302,L]
-RewriteRule ^api/v1.0/Observations\((.*)\)$ https://grlc.io/api-git/ILIAD-ocean-twin/api/Observations(id)?id=$1 [R=302,L]
-RewriteRule ^api/v1.0/Observations\((.*)-(.*)-(.*)-(.*)\)/Datastream$ https://grlc.io/api-git/ILIAD-ocean-twin/api/Observations_Datastream?id=$1&location=$1&sensor=$3&species=$4 [R=302,L]
-RewriteRule ^api/v1.0/Observations\((.*)-(.*)-(.*)-(.*)\)/FeatureOfInterest$ https://grlc.io/api-git/ILIAD-ocean-twin/api/Observations_FeatureOfInterest?id=$1&location=$1&sensor=$3&species=$4 [R=302,L]
+RewriteRule ^api/v1.0/Observations$ https://dev.grlc.io/api-git/ILIAD-ocean-twin/JF-API/Observations?page=%1 [R=302,L]
+RewriteRule ^api/v1.0/Observations$ https://dev.grlc.io/api-git/ILIAD-ocean-twin/JF-API/Observations?page=1 [R=302,L]
```
davidlehn, line 30: These rule blocks all have this pattern.  It looks like the first rules will match and the second `page=1` ones will never get used.

rapw3k: about lines 29 and 30, 
first line is to say that if the source URL has parameters including "skip=x00") then redirect to page x, second line is to say of the source URL does not have any parameter at all, then redirect to page 1 by default.

about lines 32 and 33 not sure what is question, these are ending with different resource (Datastream and FeatureOfInterest), isnt that ok?

davidlehn: Oh... I see, the RewriteCond is just for the first rule.  I'm having dejavu that I've made this same comment before.  I'll try to remember in the future. :-)  I'm not even sure what input triggers this.  Can you give an example URL?

rapw3k: sure, so for the condition rule, we have input like:
https://w3id.org/iliad/api/v1.0/Observations?$top=100&$skip=200

and for the default (no inputs)
https://w3id.org/iliad/api/v1.0/Observations

PR #3049, opened 2022-12-29
Common Provenance Model identifier

RudoWittner:
I would like to create a new permatnent identifier for the Common Provenance Model RO Crate profile.


dgarijo:
@RudoWittner please add a valid .htaccess with a redirection, as we don't reserve namespaces.
Thanks!

TallTed:
Also, better to create `README.md` (which GitHub treats specially) than `readme.md`

davidlehn:
Lower case `readme.md` works fine too.

TallTed:
Huh! Something went sideways in another repo, and I thought that was the cause ... but I guess not!

PR #3054, opened 2023-01-03
Adding a new administrator for the icon project

br0ast:
Adding Sofia Baroncini (SofiBar in GitHub) as an administrator for this namespace.

### DIFF
--- a/icon/README.md
+++ b/icon/README.md
@@ -17,12 +17,11 @@ Bologna, italy
 GitHub: [br0ast](https://github.com/br0ast)
 ORCID: [0000-0002-9152-4402](https://orcid.org/0000-0002-9152-4402)  
 
-
 and
 
-**Sofia Baroncini**
-*PhD Student*
-[University of Bologna](https://www.unibo.it)
-Bologna, Italy
-<sofia.baroncini4@unibo.it>
-GitHub: [SofiBar](https://github.com/SofiBar)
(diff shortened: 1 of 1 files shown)


icon/README.md
```diff
@@ -15,4 +15,13 @@ This space is administered by:
 Bologna, italy
 <bruno.sartini3@unibo.it>  
 GitHub: [br0ast](https://github.com/br0ast)
-ORCID: [0000-0002-9152-4402](https://orcid.org/0000-0002-9152-4402)  
\ No newline at end of file
+ORCID: [0000-0002-9152-4402](https://orcid.org/0000-0002-9152-4402)  
+
+and
+
+**Sofia Baroncini** <br>
```
TallTed, line 22: Note that you can use `  ` (two space characters) at the end of a Markdown line to the get same effect as `<br>`.

Also note that while the new info about `Sofia Baroncini` includes `<br>` at the end of each intended line, the previously existing info about `Bruno Sartini` does not include `<br>` at any line end, and two lines end with *one* space, rather than two, so [when the Markdown is rendered to a web browser](https://github.com/perma-id/w3id.org/blob/cb261354aa6b250cdc0d8934cd800fffee15e01d/icon/README.md), there will be no linewrap between `italy` and `<bruno`, nor between `br0ast)` and `ORCID:`.

PR #3068, opened 2023-01-12
Adding rules to OAIS-IP

mmannings:
URI namespace for [https://github.com/fengel/OAIS-IP](https://github.com/fengel/OAIS-IP) and example ontology files: [https://git.tib.eu/terminology/sandbox/oais-ip-ontology-collection/-/tree/master/dev/abox](https://git.tib.eu/terminology/sandbox/oais-ip-ontology-collection/-/tree/master/dev/abox).

### DIFF
--- a/tib/README.md
+++ b/tib/README.md
@@ -1,0 +1,16 @@
+# /tib/
+This [W3ID](https://w3id.org) provides a persistent URI namespace for [OAIS-IP](https://github.com/fengel/OAIS-IP) and examples: [Terminology OAIS-IP](https://git.tib.eu/terminology/sandbox/oais-ip-ontology-collection/-/tree/master/dev/abox).
+
+## Contact
+This space is administered by:  
+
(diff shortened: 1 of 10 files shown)


dgarijo:
@mmannings please remove empty htaccess.

mmannings:
Didn't know, that they might create issues, thanks. 
Done.

dgarijo:
No worries. In the future, only a README in the root folder is needed, unless you have different maintainers for that w3id in subfloders.
You may also avoid creating subfolders by having their path in the root htcaccess. Something like:
`^myfolder$ {http link}`

mmannings:
Great thanks. Will maybe restructure it later, and the other project I'm maintaining here. I was also thinking of adding another rule with strict folder name inside regex, but also liked the idea of having another folder, cause it's quite convenient, even though it will make more mess in a future.

PR #3125, opened 2023-02-17
W3ID for PRObs Lab resources

stefanogermano:
This [W3ID](https://w3id.org) provides a persistent URI namespace for [PRObs Lab](https://github.com/probs-lab) resources.


dgarijo:
@stefanogermano 
can you please add yourself as part of the maintainers in the readme?
Thanks!

stefanogermano:
Thank you for your quick review.

Yes, I can certainly do that.

I have mentioned the whole group because someone else will also edit this W3ID in the future.
But I think we can include additional maintainers/administrators when they modify something in this W3ID.

PR #3136, opened 2023-02-23
Added rdf and turtle urls

PaavoNykanen:
- Added redirect for turle and rdf files

### DIFF
--- a/ldasw/.htaccess
+++ b/ldasw/.htaccess
@@ -17,5 +17,8 @@ RewriteCond %{HTTP_ACCEPT} text/\* [OR]
 RewriteCond %{HTTP_ACCEPT} \*/\* [OR]
 RewriteCond %{HTTP_ACCEPT} ^.*application/rdf\+xml.* 
 RewriteRule ^university$ https://raw.githubusercontent.com/PaavoNykanen/Ontologies/main/LinkedData/UniversityOntology/university.owl [R=303,NE,L]
+RewriteRule ^university$ https://raw.githubusercontent.com/PaavoNykanen/Ontologies/main/LinkedData/UniversityOntology/university.rdf [R=303,NE,L]
+RewriteCond %{HTTP_ACCEPT} ^.*text/turtle.* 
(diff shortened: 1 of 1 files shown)


ldasw/.htaccess
```diff
@@ -17,5 +17,8 @@ RewriteCond %{HTTP_ACCEPT} text/\* [OR]
 RewriteCond %{HTTP_ACCEPT} \*/\* [OR]
 RewriteCond %{HTTP_ACCEPT} ^.*application/rdf\+xml.* 
 RewriteRule ^university$ https://raw.githubusercontent.com/PaavoNykanen/Ontologies/main/LinkedData/UniversityOntology/university.owl [R=303,NE,L]
+RewriteRule ^university$ https://raw.githubusercontent.com/PaavoNykanen/Ontologies/main/LinkedData/UniversityOntology/university.rdf [R=303,NE,L]
```
dgarijo, line 20: This one is not reachable, as the previous one will trigger the rule

dgarijo:
Please have a look at my small comment. Did you try the rules locally?

PaavoNykanen:
> Please have a look at my small comment. Did you try the rules locally?

Sorry which comment? I didn't try this locally since I don't know how.

dgarijo:
Now it should be vissible

ldasw/.htaccess
PaavoNykanen: Do you have a suggestion how to fix this? Maybe adding 'RewriteCond %{HTTP_ACCEPT} ^.*application/owl\+xml.* ' before the owl link and the the rdf condition for the rdf file?

Or should the owl or rdf be moved to the end as a default response?

dgarijo: you have to choose on which file to return when you are requested application/rdf+xml

PaavoNykanen: So if I left the ref+xml condition for the .rdf file, is there a similar condition for owl? Like application/owl+xml?

dgarijo: OWL is not a serialization, right? I would leave only the rdf file. Or if you want to return an .owl file when people ask for rdf, then leave just the OWL file. It's up to you to decide what is being returned.

dgarijo: well, `.rdf` is not a serialization either :)

PR #3139, opened 2023-02-28
Added a redirect for the DSTV Ontology, which represents an owl exten…

kirner-ip:
…sion of the existing DSTV standard for steel construction.

The ontology is part of the project Internet of Construction (see readme) and a future extension which is called BAUFEST 4.0

### DIFF
--- a/dstv/.htaccess
+++ b/dstv/.htaccess
@@ -1,0 +1,16 @@
+Options +FollowSymLinks
+RewriteEngine on
+
+# Turn off MultiViews
+Options -MultiViews
+
+AddType text/turtle .ttl
+
+# Rewrite rule to serve HTML content from the vocabulary URI if requested
+RewriteRule ^$ https://ip.pages.rwth-aachen.de/ioc/dstv-ontologie/
+
+# If suffix ttl, redirect to turtle version
+RewriteRule ^$ https://ip.pages.rwth-aachen.de/ioc/dstv-ontologie/ontology.ttl
+
+# Default response: html
+RewriteRule ^$ https://ip.pages.rwth-aachen.de/ioc/dstv-ontologie/ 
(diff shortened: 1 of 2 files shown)


dstv/.htaccess
```diff
@@ -0,0 +1,16 @@
...
+# Turn off MultiViews
+Options -MultiViews
+
+AddType text/turtle .ttl
+
+# Rewrite rule to serve HTML content from the vocabulary URI if requested
+RewriteRule ^$ https://ip.pages.rwth-aachen.de/ioc/dstv-ontologie/
+
+# If suffix ttl, redirect to turtle version
+RewriteRule ^$ https://ip.pages.rwth-aachen.de/ioc/dstv-ontologie/ontology.ttl
```
dgarijo, line 13: This will not work, as the previous rule superseeds it

dstv/.htaccess
```diff
@@ -0,0 +1,16 @@
...
+
+# Default response: html
+RewriteRule ^$ https://ip.pages.rwth-aachen.de/ioc/dstv-ontologie/
```
dgarijo, line 16: same as above

dgarijo:
@kirner-ip please have a look at my comments. If you add two rules that look the same, only the first one will trigger

dstv/README.md
```diff
@@ -0,0 +1,16 @@
...
+Interconnecting Data in Construction Processes - focus on steel construction and usage of the DSTV-NC standard
+See http://www.internet-of-construction.com/index_en.html
+
+## Contact
+This space is administered by:  
+
+**Lukas Kirner**  
+  *Researcher*  
+Individualized Production, RWTH Aachen, Germany.  
+https://github.com/kirner-ip
```
TallTed, line 14: The double-space-line-end is to preserve the line-break.
```suggestion
https://github.com/kirner-ip  
```

PR #3145, opened 2023-03-02
Add biopragmatics namespace

cthoyt:
This PR adds a new namespace `biopragmatics` that provides a persistent URI namespace for artifacts in the [Biopragmatics project](https://biopragmatics.github.io/).

Initially, the redirect configuration supports redirecting towards artifacts in the https://github.com/biopragmatics/obo-db-ingest, which constitute ontology files in the OBO and OWL formats that have been produced by converting biomedical databases such as the [HUGO Gene Nomenclature Committee (HGNC)](https://www.genenames.org/) and the [Universal Protein Resource (UniProt)](https://www.uniprot.org/) that are not natively ontologies. The ontology community expects ontologies to be referenced via PURLs, so W3ID is a natural fit.

### DIFF
--- a/biopragmatics/.htaccess
+++ b/biopragmatics/.htaccess
@@ -1,0 +1,6 @@
+Options +FollowSymLinks
+RewriteEngine on
+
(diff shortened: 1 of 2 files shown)


cthoyt:
thanks @davidlehn 💯

PR #3154, opened 2023-03-06
Redirection rules for the OntoUML/UFO Catalog

pedropaulofb:
Creation of permanent identifiers for the OntoUML/UFO Catalog. Matching rules tested with success.

### DIFF
--- a/ontouml-catalog/.htaccess
+++ b/ontouml-catalog/.htaccess
@@ -2,13 +2,13 @@ RewriteEngine on
 
 # Redirection Match rule for accessing the catalog repository git page
 
-RedirectMatch 302 ^/?$ https://github.com/unibz-core/ontouml-models
-RedirectMatch 302 ^/git/?$ https://github.com/unibz-core/ontouml-models
+RedirectMatch 302 ^/ontouml-catalog/?$ https://github.com/unibz-core/ontouml-models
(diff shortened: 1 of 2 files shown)


ontouml-catalog/.htaccess
```diff
@@ -0,0 +1,14 @@
+RewriteEngine on
+
+# Redirection Match rule for accessing the catalog repository git page
+
+RedirectMatch 302 ^/?$ https://github.com/unibz-core/ontouml-models
```
davidlehn, line 5: mod_alias RedirectMatch checks against the full path, so you'd need to add /ontouml-catalog at the front of all these.  You could also change to use mod_rewrite's RewriteRule which matches on the relative part.

pedropaulofb:
Dear @davidlehn, thank you very much for your review! I am sorry for the mistakes; it is the first time I am working with htaacess files. I thought the rules were already going to be applied to the specific URL and tested that way using [this website](https://htaccess.madewithlove.com/). There, I set a generic "http://test.com/" as I was expecting it to be substituted for 
"http://w3id.org/ontouml-catalog" (and not only for "http://w3id.org/").

I updated the files according to your information. Please let me know if there is something more I need to do. Once again, thank you!

PR #3178, opened 2023-03-13
Add new rules and information for ontouml ttl schema

pedropaulofb:


davidlehn:
@tgoprince You were the only maintainer listed here, but it looks like you are all working on the same target project, so I'm merging this.  If that's an issue, let me know.

PR #3179, opened 2023-03-14
Add rewrite rule for QA manifest

NSeydoux:
Temporarily redirect the QA manifest to GitHub pages.

### DIFF
--- a/inrupt/.htaccess
+++ b/inrupt/.htaccess
@@ -2,5 +2,5 @@ Options +FollowSymLinks
 
 RewriteEngine On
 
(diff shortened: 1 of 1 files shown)


dgarijo:
I don't think you need the quotes here. Have you tried this redirection locally?

NSeydoux:
Yup, I think the quotes are optional, I just added them because they are present in the documentation, so I assumed it was a best practice :)

NSeydoux:
Thanks for being so quick by the way!

PR #3203, opened 2023-03-23
simplify redirection rules and update fdp

pedropaulofb:


pedropaulofb:
@tgoprince 
@mozzherina
@claudenirmf

PR #3212, opened 2023-03-28
Add https://w3id.org/dmi for the Digital Material Identifier DMI

webhart:


dgarijo:
@webhart please add yourself as a maintainer in the readme file

PR #3225, opened 2023-04-04
Add redirect `rorio` for ROR in OWL

cthoyt:
ROR in OWL is an ontology that wraps [Research Organization Registry (ROR)](https://ror.org/) identifiers as named individuals so they can be more easily included in other ontologies (e.g., OBO Ontologies) and empower the [Protégé](https://protege.stanford.edu/) ontology editor to show rich label information for `dc:contributor` annotations.

The goal of having an entry in w3id.org is to endow this ontology with a PURL as its ontology IRI.

This PR is similar in concept to https://github.com/perma-id/w3id.org/pull/3006 but for a different resource (i.e., ROR instead of ORCID)

### DIFF
--- a/rorio/README.md
+++ b/rorio/README.md
@@ -1,0 +1,16 @@
+# ROR in OWL
+
+This [W3ID](https://w3id.org/) provides a persistent URI namespace for [Research Organization Registry (ROR)](https://ror.org) records in OWL resources.
+
+## Uses
+
+ROR in OWL is an ontology that wraps ROR identifiers as named individuals so they can be more easily included in other ontologies (e.g., OBO Ontologies) and empower the [Protégé](https://protege.stanford.edu/) ontology editor to show rich label information for `dc:contributor` annotations.
+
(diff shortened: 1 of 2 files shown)


rorio/README.md
```diff
@@ -0,0 +1,16 @@
...
+ROR in OWL is an ontology that wraps ROR identifiers as named individuals so they can be more easily included in other ontologies (e.g., OBO Ontologies) and empower the [Protégé](https://protege.stanford.edu/) ontology editor to show rich label information for `dc:contributor` annotations.
+
+The artifacts referenced by this entry are versionioned in the following GitHub repository: https://github.com/cthoyt/rorio
+
+## Contact
+
+Charles Tapley Hoyt
+Email: cthoyt@gmail.com
+GitHub: [@cthoyt](https://github.com/cthoyt/)
+ORCID: [0000-0003-4423-4370](https://orcid.org/0000-0003-4423-4370)
```
TallTed, line 16: Note typo: `versionioned` -> `versioned`

Also note line wraps are not preserved when MD is rendered.  Need to end each wrapping line with either two spaces (`  `) or an HTML BReak (`<br>`)

PR #3226, opened 2023-04-04
Creating airo 

DelaramGlp:
As w3id urls are case-sensitive, I have created this request for https://w3id.org/airo. 

https://w3id.org/AIRO will be redirected to this new url.


airo/.htaccess
```diff
@@ -0,0 +1,3 @@
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^$ https://https://delaramglp.github.io/airo [R=302,L]
```
davidlehn, line 3: URL typo

PR #3237, opened 2023-04-11
Add /ark:57802/

dwinston:
This [W3ID](https://w3id.org) provides a persistent URI namespace for resources under the [archival resource key (ARK)](https://arks.org/about) name assigning authority number (NAAN) 57802. That is, this W3ID service acts as an ARK name mapping authority (NMA).

## Uses

This namespace may be broken down using so-called ["shoulders"](https://www.ietf.org/archive/id/draft-kunze-ark-36.html#name-optional-shoulders) at the .htaccess level (i.e. without slashes `/`.), as per the ARK scheme.

Example name: https://w3id.org/ark:57802/dw0/agu (current redirect to https://ns.polyneme.xyz/ark:57802/dw0/agu) retrieves a representation of the American Geophysical Union (AGU) bibliographic index terms as a SKOS ConceptScheme. https://w3id.org/ark:57802/dw0/agu?_mediatype=text/html explicitly requests a text/html representation.

## Contact

I was assigned the NAAN 57802, and I administer it. I've provided contact info in the README as part of this PR.

### DIFF
--- a/ark:57802/.htaccess
+++ b/ark:57802/.htaccess
@@ -1,0 +1,3 @@
+RewriteEngine On
+
+RewriteRule (.*) https://ns.polyneme.xyz/ark:57802/$1 [R=302,L,QSA]
--- a/ark:57802/README.md
+++ b/ark:57802/README.md
@@ -1,0 +1,22 @@
+# /ark:57802/
+
+This [W3ID](https://w3id.org) provides a persistent URI namespace for resources under the [archival resource key (ARK)](https://arks.org/about) name assigning authority number (NAAN) 57802. That is, this W3ID service acts as an ARK name mapping authority (NMA).
+
+## Uses
+
+This namespace may be broken down using so-called ["shoulders"](https://www.ietf.org/archive/id/draft-kunze-ark-36.html#name-optional-shoulders) at the .htaccess level (i.e. without slashes `/`.), as per the ARK scheme.
+
+Example name: https://w3id.org/ark:57802/dw0/agu retrieves a representation of the American Geophysical Union (AGU) bibliographic index terms as a SKOS ConceptScheme. https://w3id.org/ark:57802/dw0/agu?_mediatype=text/html explicitly requests a text/html representation.
+
+## Contact
+
+This space is administered by:
+
+**Donny Winston**  
+_Metadata Infrastructure Architect_  
+[Polyneme LLC](https://polyneme.xyz/)  
+New York, New York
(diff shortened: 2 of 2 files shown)


davidlehn:
That ark spec is long.  Looks like it explicitly requires `:` in the names.  I have a concern about reserved characters in windows filenames.  Is that `:` going to cause problems when windows uses checkout this repo?
https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words

davidlehn:
Also there's a concern if 57802+ ARK NAAN owners wanted to use w3id.org as a resolver service. :-)  The infrastructure here is not currently setup to scale like that.

dwinston:
As a possible remedy to both concerns, I noticed when checking out the repo locally (on MacOS) that I needed to create and mount a case-sensitive filesystem, and thus the `/ark/` namespace appears to be available. I would be willing to amend this PR (or close and open a new one?) to claim that.

I am on the ARK Alliance [technical working group](https://wiki.lyrasis.org/plugins/servlet/mobile?contentId=108757991#content/view/108757991) (linked to from https://arks.org/community/), so I could help facilitate NAAN resolution by confining PRs to updating `/ark/.htaccess`. The shared URLs would then look like `https://w3id.org/ark/ark:57802/…` to adhere to the ARK spec, which isn’t my preference, but it would work.

davidlehn:
Looks like the generic handling idea was just submitted for `arkid` in https://github.com/perma-id/w3id.org/pull/3238.  Are the two of you working together?

I think we want to avoid case conflicting ids on this service.  It's just going to cause confusion and probably break on some file systems.  So `ark` probably not going to work due to `ARK` existing.

As a side note, it does seem like with the short arks.org domain, it would be ideal to run a service there somehow.  (arks.org/ids/..., id.arks.org/..., arks.org/ark:..., etc)

dwinston:
Ah yes, @datadavev and I work together. We are also considering also using arks.org, but part of the ARK scheme’s intent is to support decentralized resolution, so we wouldn’t want to centralize on arks.org (or any other domain).

Given #3238 and your stated preference to avoid case-conflicting IDs, I’ll close this.

datadavev:
Ha, yes, but it's going to be around for a while, right? The plan is to promote arks.org as the primary location for information about ARK identifiers and resolution, etc. We would like to communicate to users that w3id.org provides another address which can be used to locate services that work with ARK identifiers. Having a stable domain for that is very helpful for a community working with resource archives.

PR #3238, opened 2023-04-11
Support for Archival Resource Key (ARK) identifiers

datadavev:
Support for Archival Resource Key (ARK) identifiers

### DIFF
--- a/arkid/.htaccess
+++ b/arkid/.htaccess
@@ -1,0 +1,23 @@
+# w3id .htaccess rules for arkid
+Options +FollowSymLinks
+RewriteEngine on
+
+# ARK specification
+RewriteRule ^spec|ark:?$ https://datatracker.ietf.org/doc/draft-kunze-ark/ [R=302,L]
+
+# Any ARK identifier resolution request (/ark:value or /ark:/value)
+# e.g. https://w3id.org/arkid/ark:12345/foo
+RewriteRule ^ark:(.+)$ https://n2t.net/ark:$1 [R=302,L]
+
+# Any ARK value resolution
+# e.g. https://w3id.org/arkid/12345/foo
+RewriteRule ^([0-9]{4,8})/(.*)$ https://n2t.net/ark:$1/$2 [R=302,L]
+
+# NAAN registry in ANVL
+RewriteRule ^naans?|registry$ https://n2t.net/e/pub/naan_registry.txt [R=302,L]
+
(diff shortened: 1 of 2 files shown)


davidlehn:
Is this related to https://github.com/perma-id/w3id.org/pull/3237?
Out of curiosity, is this using `arkid` due to `ARK` already existing?

arkid/.htaccess
```diff
@@ -0,0 +1,23 @@
...
+
+# Any ARK value resolution
+# e.g. https://w3id.org/arkid/12345/foo
+RewriteRule ^([0-9]{4,8})/(.*)$ https://n2t.net/ark:$1/$2 [R=302,L]
+
+# NAAN registry in ANVL
+RewriteRule ^naans?|registry$ https://n2t.net/e/pub/naan_registry.txt [R=302,L]
+
+# Register or update NAAN
+RewriteRule ^naans?|register$ http://n2t.net/e/naan_request [R=302,L]
```
davidlehn, line 20: The previous rule will catch `nanns?`.  Should this be just `register`?

datadavev:
> Is this related to #3237? Out of curiosity, is this using `arkid` due to `ARK` already existing?

Yes, correct. Using "ark" would be preferable, but likely confusing with the existing "ARK" entry.

arkid/.htaccess
datadavev: Yes, copy n' paste slop. I'll correct it. Thanks.

davidlehn:
> Yes, correct. Using "ark" would be preferable, but likely confusing with the existing "ARK" entry.

I'm not sure of the state of the world now, but in the past I think some file systems just overwrote files that only differed in case.  Want to try and play nice with that.

You could ask the ARK maintainers if they are still using that prefix.  I imagine some projects become obsolete over time, but I didn't look into their work.  Of course "cool URIs don't change", so taking over anything would need to be done with care.

datadavev:
> I'm not sure of the state of the world now, but in the past I think some file systems just overwrote files that only differed in case. Want to try and play nice with that.

There is one instance of this problem in w3id.org now, `airo` and `AIRO` results in a very unpleasant experience with git on my mac, which happens to be formatted case aware, but not sensitive.

davidlehn:
Oops.  Yeah, that slipped through and has an active PR.  I will try to address it.  Future plan is to have an automated check for that issue.

PR #3240, opened 2023-04-12
Update .htaccess

DelaramGlp:


davidlehn:
Hi, Having directories `AIRO` and `airo` that differ only in case can cause problems on some file systems.  It looks like you are using them the same way?  Can they be combined into just one dir?  We shouldn't have let this through in the first place, but since it's out there, if you have existing use cases that can't be corrected, we could add a [special root .htaccess rule](https://github.com/perma-id/w3id.org/blob/master/.htaccess) to redirect one to the other.  It would be nice to avoid that if one id would work for you.

See: https://github.com/perma-id/w3id.org/pull/3238#issuecomment-1505686992

DelaramGlp:
Hi, I did not know that w3id urls are case-sensitive when I first added `AIRO` and now I need to keep both https://w3id.org/AIRO and https://w3id.org/airo. I would be happy to combine the two into one (`airo`) if we could redirect `AIRO` to it. Thanks.

davidlehn:
Ok.  Ideally AIRO could be avoided altogether.  But if you are sure you need to keep it, can you please update this PR with the following:
- Remove the `AIRO` directory.
- Update `airo` to have whatever the merged content should be.
- Update the root `.htaccess` file with a change like this:
```diff
-# redirect here vs in a UN directory due to case insensitive file systems issue
-RewriteRule ^UN(.*)$ /un$1
+# special directory redirects due to case insensitive file systems issues
+RewriteRule ^AIRO(.*)$ /airo$1
+RewriteRule ^UN(.*)$ /un$1
```
Also:
- Test this locally if possible.
- If appropriate, update your project documentation to recommend using `airo`.

PR #3242, opened 2023-04-13
Update readme.md and add one rule

prapalu:


esw/readme.md
```diff
@@ -6,7 +6,7 @@ This [W3ID](https://w3id.org) provides a persistent URI namespace for Explorator
 This space is administered by:  
 
 **Gianmarco Prando**  
-*Computer Engineer *
+*Computer Engineer*
```
TallTed, line 9: Note that by ending this line with the `*` (and omitting the following two spaces I suggested), there is no line break when the Markdown is delivered to browsers as HTML; it is as if this line were 
```
*Computer Engineer*Padua, Italy  
``` 
in the Markdown, rather than as intended --
```
*Computer Engineer*  
Padua, Italy  
```
(Note that the latter line *does* end with two space chars before the newline.)

PR #3253, opened 2023-04-20
Update .htaccess

DelaramGlp:
changes applied to redirect `AIRO` to `airo`

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -3,5 +3,6 @@
 Options +FollowSymLinks
 RewriteEngine on
 
-# redirect here vs in a UN directory due to case insensitive file systems issue
+# special directory redirects due to case insensitive file systems issues
+RewriteRule ^AIRO(.*)$ /airo$1
 RewriteRule ^UN(.*)$ /un$1


davidlehn:
We are all getting confused on this! :-)

@dgarijo: @DelaramGlp specifically said AIRO redirect needs to keep working.  We want to avoid filenames that only differ in case because it causes some filesystems to have problems.  I suggested the hack here in at:
https://github.com/perma-id/w3id.org/pull/3240#issuecomment-1507770689.
It will keep both `AIRO` and `airo` working and just have one dir (`airo`) with the rules.

In general we need to avoid this happening before redirects are created.  But for compatibility, this hack works.

@DelaramGlp This would have been easier to manage in one PR instead of lots.  Now we have conflicting advice from @dgarijo and myself across too many PRs.  Please pick `AIRO` or `airo`, update the files in that one dir, delete the files from the other dir, and update the root `.htaccess` that redirects to the dir name you choose.  Please make it clear which way you are choosing to go.  Thanks.

Refs:
https://github.com/perma-id/w3id.org/pull/3240
https://github.com/perma-id/w3id.org/pull/3254
https://github.com/perma-id/w3id.org/pull/3260
https://github.com/perma-id/w3id.org/pull/3261

dgarijo:
Ok, I see. I will let you handle this one then.

PR #3254, opened 2023-04-20
Delete AIRO directory

DelaramGlp:


dgarijo:
Can you please add a redirection to /airo instead? If we mint permanent ids, we are usually reluctant on deleting them. Thanks!

DelaramGlp:
Hi Daniel! 
I have made the changes based on this comment: [https://github.com/perma-id/w3id.org/pull/3240]. 
I can redirect it to /airo as well. I just need to keep both `AIRO` and `airo`.

dgarijo:
Yeah, but I would like to avoid commits to the main htaccess, as it can potentially cause trouble. Please add your edits here instead. Thanks!

PR #3256, opened 2023-04-20
rml: add RML ontology

DylanVanAssche:
W3C Community Group on Knowledge Graph Construction will release soon the new RML ontology which uses W3ID IRIs. Add them and update the README.


DylanVanAssche:
Call to all specification maintainers to approve these redirections for the new RML ontology.

@pmaria
@anaigmo 
@dachafra
@bjdmeest
@arenas-guerrero-julian
@frmichel
@chrdebru
@DylanVanAssche 
@samiscoding
@andimou

bjdmeest:
Can we not add `RewriteRule ^(.*)/$ /$1 [L,R=301]` to skip all trailing slash duplicate lines?

rml/.htaccess
```diff
@@ -0,0 +1,106 @@
...
+RewriteRule rml/GatherMap$ https://kg-construct.github.io/rml-cc-spec/spec/docs#http://w3id.org/rml/GatherMap
+RewriteRule rml/GatherMap/$ https://kg-construct.github.io/rml-cc-spec/spec/docs#http://w3id.org/rml/GatherMap
+
+# Ontology documentation
+RewriteRule rml/cc$ https://kg-construct.github.io/rml-cc-spec/ontology/documentation/index-en.html
+RewriteRule rml/cc/$ https://kg-construct.github.io/rml-cc-spec/ontology/documentation/index-en.html
+
+# === RML FNML ===
+# Specification
+RewriteRule rml/fnml/spec$ https://kg-construct.github.io/rml-fnml-spec/spec/docs
```
bjdmeest, line 82: the spec is online at https://kg-construct.github.io/fnml-spec/

DylanVanAssche:
@bjdmeest I used https://htaccess.madewithlove.com/ to test the htaccess rules, that suggestion doesn't work there. Maybe it does in a real Apache server?

bjdmeest:
@DylanVanAssche weird, below works for me

```
Options +FollowSymLinks
RewriteEngine on

RewriteRule ^(.*)/$ /$1 [L,R=301]

# === RML Star ===
# Specification
RewriteRule rml/star/spec$ https://kg-construct.github.io/rml-star-spec/spec/docs
```

https://w3id.org/rml/star/spec/ gets redirected to https://w3id.org/rml/star/spec
https://w3id.org/rml/star/spec gets redirected to https://kg-construct.github.io/rml-star-spec/spec/docs

bjdmeest:
Shouldn't we also include a catchall, eg `RewriteRule ^(.*)$ https://kg-construct.github.io/404?page=$1 [R=302,L]` at the bottom?

Then, https://w3id.org/rml/thisdoenstexist gets redirected to https://kg-construct.github.io/404?page=rml/thisdoenstexist

davidlehn:
Without knowing anything about your project, you could vastly simplify and future proof these rules if you let the resources be scoped to the sub project names.  Use scoped `https://w3id.org/rml/scope/FooBar` vs `https://w3id.org/rml/FooBar`.  The whole file might collapse to (untested):

```
RewriteRule ^(\w+)/spec/?$ https://kg-construct.github.io/rml-$1-spec/spec/docs [R,L]
RewriteRule ^(\w+)/(\w+)/?$ https://kg-construct.github.io/rml-$1-spec/spec/docs#http://w3id.org/rml/$1/$2 [NE,R,L]
RewriteRule ^(\w+)/?$ https://kg-construct.github.io/rml-$1-spec/ontology/documentation/index-en.html [R,L]
```

Nonsense URLs would just pass through to your server and 404 or otherwise not work.  Using w3id.org for important validation is probably an anti-pattern.  It's often better handled by clients or target servers.  On the plus side, if you keep using the same pattern on your side, you never have to touch this site again when updating your resources.

But if you want the top level resources, and they have special target scopes, it gets more verbose.  Tradeoffs.

I'm assuming you all are related to the current `rml/editor/...` rules here?  The above would need to have some rule at the top to handle that properly and continue processing in the sub dir.

frmichel:
Hi all, I'm confused here, made a post on Slack but better have everyone here.
Currently, repos have various names like fnml-spec and collections-containers-spec. 
The redirections mention rml-fnml-spec and rml-cc-spec. So should we rename the repos to have a consistent naming for all of them? Or should be just update the htaccess redirections?

rml/.htaccess
DylanVanAssche: After the meeting today, the Redirections spreadsheet is finalized. Only the FNML spec needs to update its repo now to adjust its URL correctly for Github Pages.

DylanVanAssche:
> So should we rename the repos to have a consistent naming for all of them?
> Or should we just update the htaccess redirections?

For the record: As decided internally, we adjusted the repos to be consistent.

DylanVanAssche:
> Without knowing anything about your project, you could vastly simplify and future proof these rules if you let the resources be scoped to the sub project names. Use scoped https://w3id.org/rml/scope/FooBar vs https://w3id.org/rml/FooBar.

Scoping them in separate subdirectories is what was disagreed on, they want to use the same root namespace for all to have one prefix across all specifications. Using scopes would indeed make things so much easier :)

> Nonsense URLs would just pass through to your server and 404 or otherwise not work. Using w3id.org for important validation is probably an anti-pattern. It's often better handled by clients or target servers. On the plus side, if you keep using the same pattern on your side, you never have to touch this site again when updating your resources.

These IRIs are solely for the RML ontology to have stable IRIs, so I think we are fine as we agreed internally to re-use the same pattern which makes the file much smaller.

> I'm assuming you all are related to the current rml/editor/... rules here? The above would need to have some rule at the top to handle that properly and continue processing in the sub dir.

Yes, I didn't add that yet because I expected requested changes anyway :) Now that things are more finalized, I will rework this PR to include all requested changes and handle RML editor stuff as well. Thanks!

rml/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+**RML IO**
+* https://w3id.org/rml/io
+* https://w3id.org/rml/io/spec
+* https://w3id.org/rml/LogicalSource
+* https://w3id.org/rml/LogicalTarget
+* https://w3id.org/rml/Source
+* https://w3id.org/rml/Target
+
+Contacts:
```
TallTed, line 46:
```suggestion
## Contacts
```

davidlehn:
Just to check, is this ready to be merged?

PR #3262, opened 2023-04-21
Updated .htaccess to include ieee-1451.0-schema

rbrama79:
Dear Josh, I hope to have correctly updated the file in order to link to ieee-1451.0-schema repository. Hope this helps you. Best regards.

PR #3263, opened 2023-04-21
Update .htaccess

joshuagay:
Adds rule for ieee-1451.0-schema rewriterule

### DIFF
--- a/ieee/.htaccess
+++ b/ieee/.htaccess
@@ -31,7 +31,13 @@ RewriteRule ^ieee-2791-schema(/)?$ https://opensource.ieee.org/2791-object/ieee-
 ## ^ieee-1451.0-schema/
 ##
 
-RewriteRule ^ieee-1451.0-schema/((examples/)?\w+).xsd$ https://opensource.ieee.org/1451/1451.0-schema/-/blob/main/normative/$1.xsd [R=302,L]
+## redirects to xml schemas (.xsd files)
+## e.g. https://opensource.ieee.org/1451/1451.0-schema/-/raw/main/normative/CalibrationIdentificationTEDS.xsd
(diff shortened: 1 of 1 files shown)


ieee/.htaccess
```diff
@@ -27,6 +27,12 @@ RewriteRule ^ieee-2791-schema/(.+)$ https://opensource.ieee.org/2791-object/ieee
 ## link to only ^ieee-2791-schema or ^ieee-2791-schema/
 RewriteRule ^ieee-2791-schema(/)?$ https://opensource.ieee.org/2791-object/ieee-2791-schema/ [R=302,L]
 
+##
+## ^ieee-1451.0-schema/
+##
+
+RewriteRule ^ieee-1451.0-schema/((examples/)?\w+).xsd$ https://opensource.ieee.org/1451/1451.0-schema/-/blob/main/normative/$1.xsd [R=302,L]
```
davidlehn, line 34: I made an error in this simplification.  The original PR, https://github.com/perma-id/w3id.org/pull/3262, maps the examples from xsd to xml.  Two rules will address the issue:
```suggestion
RewriteRule ^ieee-1451.0-schema/(\w+).xsd$ https://opensource.ieee.org/1451/1451.0-schema/-/blob/main/normative/$1.xsd [R=302,L]
RewriteRule ^ieee-1451.0-schema/examples/(\w+).xsd$ https://opensource.ieee.org/1451/1451.0-schema/-/blob/main/normative/$1.xml [R=302,L]
```
- Should those be using `raw` instead of bringing up the `blob` UI?  I don't know the use case here to determine what you'd want.
- Should the examples be mapping `xsd` to `xml` or should that be `xml` to `xml`?

joshgay: Thank you for your help. XML -> XML and XSD -> XSD. I will update accordingly

PR #3277, opened 2023-04-26
GerPS-Ontology was added

maxraupach:
We need a permanent Identifier for our `GerPS-Ontology`. If something went wrong please contact us! Our Contacts are in the `README.md` File under the headline: `Contact`.


dgarijo:
@maxraupach have you tried this htaccess file? you are pointing to local files, so I am afraid it won't work.
See the `/example` repo for an example

maxraupach:
i will fix this Issue, but i need some time, because i co-worker has to provide me some informations =)

maxraupach:
i am afraid, i used an old identifier, can we change to: `https://w3id.org/GerPS-onto/ontology` ?

PR #3287, opened 2023-04-27
update contacts

maxraupach:


maxraupach:
i have a question, wenn i try the redirect link: `https://w3id.org/GerPS-onto/ontology`, it didnt work. Will it be unlock, or is the .htaccess broken?

dgarijo:
Your w3id is http://w3id.org/GerPS-Ontology, so that's the one that works

maxraupach:
oh ok! how can i change it?

dgarijo:
You have to rename the folder to exactly what you want your w3id to be.
You named your folder `GerPS-Ontology` and hence that's the w3id you got.
If you rename your folder to `GerPS-onto` and in the htaccess you add the redirection for `ontology`, it should work

maxraupach:
ty very much =)

PR #3291, opened 2023-04-28
New W3ID rewrite for Zotero group library 4530692

skybristol:
Please let me know if there is anything off with the suggested config. We will have several more like this. While I know I could include any number of these under a single root subdomain, I want to limit the functionality to specifically "registered" archive collections as part of our work and not any group library identifier. If you'd prefer that I move these under a single subdomain, I can do that. I just also want to keep the URLs as short as possible and "DOI-like" as they will be included in citation reference strings for print publications.


4530692/.htaccess
```diff
@@ -0,0 +1,13 @@
+# Rewrite engine setup
+RewriteEngine on
+
+# Item within library in Web UI
+RewriteRule ^(.*) https://www.zotero.org/groups/4530692/usgs_ni_43-101_reports/items/$1 [R=302,L]
```
davidlehn, line 5: This probably needs to go at the bottom.  I think it will match everything first before those cond rules are checked.

skybristol: Got it. Thanks. I made that change in the latest commit.

davidlehn:
You'll have several more random number dirs?  It is nicer to have readable names designed for long-term use.  That has been the traditional usage here.  Most use cases here try to avoid locking in permanent ids to service specific URLs and ids like a random zotero id.  If you stop using that service, or change ids there, or similar, your ids here would still include that group id.  If you're really sure that's what you want to do, we can do it.

It's also not as fun to maintain the current giant folder if it just has numbers in it! :-)

skybristol:
Like I said, I was on the fence about it. I like having a shorter path. I also wasn't sure about essentially creating a route for any Zotero library anywhere. It would be easier to manage though as we won't have to come back and commit whole new "random" number directories for you guys. Let me think about it for a minute or two and see if I can come up with an approach that will be workable.

skybristol:
Ah. I see my friends over at DataONE decided they should own /usgs/, which would have been my next choice. It was @datadavev who turned me onto this capability to begin with. Let me see about maybe coordinating something under that namespace.

skybristol:
Okay. I reworked this as a subfolder under /usgs/. I also put in a single README at the /usgs/ root describing both cases. This may help us keep track of and organize further things we want to do down the road. I think this configuration should work based on what I saw in a couple other cases with subfolders, but let me know if I need to tweak anything.

usgs/README.md
```diff
@@ -0,0 +1,43 @@
...
+### Uses
+
+Handles dereferencing logic, redirection, and SPARQL query proxying for the GNIS-LD implementation.
+
+### Links
+
+With "$1" being a unique identifier from the GNIS-LD implementation of the GNIS ontology...
+
+Links in the form <https://w3id.org/usgs/$1>
+* Redirects to http://gnis-ld.org/lod/$1
```
TallTed, line 18:
```suggestion
Links in the form `https://w3id.org/usgs/$1`
* Redirect to `http://gnis-ld.org/lod/$1`
```

usgs/README.md
```diff
@@ -0,0 +1,43 @@
...
+### Uses
+This namespace starts with the unique identifier for the Zotero Group Library containing bibliographic metadata and documents for  National Instrument (NI) 43-101 Technical Reports used as scientific references and in need of a persistent citation source.
+
+### Links
+With "$1" being a unique Zotero library identifier and "$2" being a unique item identifier within that library...
+
+Links in the form <https://w3id.org/z/$1/$2>
+* No Accept header or text/html -> <https://www.zotero.org/groups/$1/usgs_ni_43-101_reports/items/$2>
+* Using Accept header application/json -> <https://api.zotero.org/groups/$1/items/$2?format=atom>
+* Using Accept header application/atom+xml -> <https://api.zotero.org/groups/$1/items/$2?format=atom>
```
TallTed, line 36:
```suggestion
With `$1` being a unique Zotero library identifier and `$2` being a unique item identifier within that library...

Links in the form `https://w3id.org/z/$1/$2`
* No `Accept` header or `text/html` -> `https://www.zotero.org/groups/$1/usgs_ni_43-101_reports/items/$2`
* Using `Accept` header `application/json` -> `https://api.zotero.org/groups/$1/items/$2?format=atom`
* Using `Accept` header `application/atom+xml` -> `https://api.zotero.org/groups/$1/items/$2?format=atom`
```

usgs/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Contacts
+
+* Sky Bristol
+* sbristol@usgs.gov
+* Github: [https://github.com/skybristol](skybristol)
+* ORCID: [https://orcid.org/0000-0003-1682-4031](0000-0003-1682-4031)
```
TallTed, line 43:
```suggestion
  * sbristol@usgs.gov
  * Github: [https://github.com/skybristol](skybristol)
  * ORCID: [https://orcid.org/0000-0003-1682-4031](0000-0003-1682-4031)
```

davidlehn:
Some concerns here (I wasn't watching before this got merged):
- @dgarijo Did you [squash merge](https://github.com/perma-id/w3id.org/commit/a5249449836cce93bda885d23c6c8429a8db269b) this?  I just shut off squash merges, since we usually don't want to do that.
- @skybristol From testing, I see it's checking the `z` dir first.  I forgot if it checked dirs or .htaccess first.  I imagine that might confuse others too.  The `usgs/.htaccess` file does have a catch all.  In theory, if someone wanted to use the gnis-ld redirect with something that started with `z/...`, the new rules would override that now.  I don't know if that's a problem in practice.  In general, I would suggest not having setups with those edge cases in the first place.
- @blake-regalia This did jump into the redirect you setup a long while ago (from before we were insisting on some contact info).  Is the `usgs/z/...` usage going to interfere with your usage?  Is it otherwise ok to mix these two use cases here?

usgs/README.md
```diff
@@ -0,0 +1,43 @@
...
+## GeoArchive collection of NI 43-101 Technical Reports (/4530692/)
+This W3ID namespace provides persistence to scientific reference assets that we know will need to move to different technologies over time. It is one of a number of collections USGS is setting up as part of a "GeoArchive." We are currently leveraging the Zotero reference manager for metadata/file storage using the web interface as human links for reference strings in reports and articles and the API for machine access in supported formats.
+
+### Uses
+This namespace starts with the unique identifier for the Zotero Group Library containing bibliographic metadata and documents for  National Instrument (NI) 43-101 Technical Reports used as scientific references and in need of a persistent citation source.
+
+### Links
+With `$1` being a unique Zotero library identifier and `$2` being a unique item identifier within that library...
+
+Links in the form `https://w3id.org/z/$1/$2`
```
davidlehn, line 33: Left out `usgs` here:
```diff
-Links in the form `https://w3id.org/z/$1/$2`
+Links in the form `https://w3id.org/usgs/z/$1/$2`
```

PR #3303, opened 2023-05-04
Add spm redirects

joleaf:


spm/.htaccess
```diff
@@ -0,0 +1,24 @@
...
+Options +FollowSymLinks
+
+# Turn off MultiViews
+Options -MultiViews
+
+RewriteEngine On
+
+## Redirects 
+# DOCS
+RewriteRule ^(.*)docs/colpadef https://userpages.uni-koblenz.de/~process-science/docs/fg-bks-social-process-mining-collaboration-detection/main/ [R=303,L]
```
dgarijo, line 14: have you tested this rule? you are not adding the `$` at the end, which looks odd. Also, the `(.*)` before the docs redirection.

joleaf: Hey @dgarijo, thank you for the reviews. Yes, these redirects work. However, I will add the „$“ in a new commit!

Issue #2853, opened 2022-08-08
Redirecting paths with urlencoded URIs fails (AllowEncodedSlashes server config?)


awagner-mainz:
Hi all,

I have set up a server that offers a [reconciliation service](https://reconciliation-api.github.io/specs/latest/) for SKOS concept schemes. For this, it offers API endpoints that include the ConceptScheme's URI - url-encoded - in the request path. (For example, see <https://c111-064.cloud.gwdg.de/reconc/mpilhlt/https%3A%2F%2Fw3id.org%2Fmpilhlt%2Fworktime_role> or <https://c111-064.cloud.gwdg.de/reconc/mpilhlt/https%3A%2F%2Fw3id.org%2Fmpilhlt%2Fworktime_role/_preview/r2>.)

Now I would like to put this service behind a w3id.org redirection, too. My .htaccess thus has a [line](https://github.com/perma-id/w3id.org/blob/a6eca7f2af63da4b8c3efbf7936a7c8c4bf2f8e2/mpilhlt/.htaccess#L17) like this:

```htaccess
RewriteRule   ^reconcile/(.*)   https://c111-064.cloud.gwdg.de/reconc/mpilhlt/$1   [R=307,NE,L]
```

This seems not to work, however. Running curl fails as soon as the path includes an url-encoded forward slash:

```console
> curl -I https://w3id.org/mpilhlt/reconcile/https%3A
HTTP/1.1 307 Temporary Redirect
Date: Fri, 05 Aug 2022 22:31:37 GMT
Server: Apache/2.4.29 (Ubuntu)
Access-Control-Allow-Origin: *
Location: https://c111-064.cloud.gwdg.de/reconc/mpilhlt/https%3A
Content-Type: text/html; charset=iso-8859-1

> curl -I https://w3id.org/mpilhlt/reconcile/https%3A%2F
HTTP/1.1 404 Not Found
Date: Fri, 05 Aug 2022 22:31:49 GMT
Server: Apache/2.4.29 (Ubuntu)
Access-Control-Allow-Origin: *
Content-Type: text/html; charset=iso-8859-1
```

[...]

PR #3313, opened 2023-05-05
try to solve urlencoding problem

smrgeoinfo:
the rewrite rule has an httpURI in the query value and the target server requires these to be URL encode, but when the rewrite happens, all the #n encodings in the URI are ignored.  Try fix based on github  #2846.  
see also #2853 

will this work?

### DIFF
--- a/geochem/1.0/.htaccess
+++ b/geochem/1.0/.htaccess
@@ -4,6 +4,6 @@ AllowEncodedSlashes NoDecode
 
 # .htaccess rewrite rule tester (doen't do HTTP:Accept header conditions)  https://htaccess.madewithlove.be/
 
-RewriteRule    ^analyticalmethod/([A-Za-z_]+).html$     https://demo.vocabs.ardc.edu.au/repository/api/lda/earthchem-georoc/analytical-methods-for-geochemistry/0-1/resource?uri=http%3A%2F%2Fw3id.org%2Fogeochem%2Fdef%2F1%2Fanalyticalmethod%2Fmethod%2F$1  [R=303,L]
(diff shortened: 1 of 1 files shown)


davidlehn:
I don't think that's option is currently enabled.  What's an example URL and the resulting redirect you want?

smrgeoinfo:
input url:
http://w3id.org/geochem/1.0/analyticalmethod/particleinducedxrayspectrometry

rule:
RewriteRule    ^analyticalmethod/([A-Za-z_]+)$     https://demo.vocabs.ardc.edu.au/repository/api/lda/earthchem-georoc/analytical-methods-for-geochemistry/0-1/resource?uri=http%3A%2F%2Fw3id.org%2Fogeochem%2Fdef%2F1%2Fanalyticalmethod%2Fmethod%2F$1  [R=303,L]

rewrite should be:
https://demo.vocabs.ardc.edu.au/repository/api/lda/earthchem-georoc/analytical-methods-for-geochemistry/0-1/resource?uri=http%3A%2F%2Fw3id.org%2Fogeochem%2Fdef%2F1%2Fanalyticalmethod%2Fmethod%2Fparticleinducedxrayspectrometry


what is the result query uri to https://demo.vocabs.ardc.edu.au/:
httpAFFw3id.orgFogeochemFdefF1FanalyticalmethodFmethodFparticleinducedxrayspectrometry

smrgeoinfo:
add 'NE' in redirect paramters, per suggeston from Nick Car

smrgeoinfo:
this is ready to pull...

davidlehn:
This won't work.  "AllowEncodedSlashes" is not allowed in htaccess files and the server will just log errors and return a 500 error.

smrgeoinfo:
Yup, that's what happens. Back to the drawing board...

smrgeoinfo:
I tried without the allowEncodedSlashes, and added the [R=303,L,NE]  and get the same bad result-- the URL encoding gets smashed. Error is ```
That shortname is unrecognised.
unrecognised short name or literal: httpAFFw3id.orgFogeochemFdefF1FanalyticalmethodFmethodFbroadbeamxrayspectrometry
```
the 'F's should all be '%2F' and the 'A' '%3A'

PR #3317, opened 2023-05-06
Seb seis patch 3

SebSeis:


davidlehn:
What are you doing here?  Is there a use case related to hosting google verification files?  We'd rather not do that.  If you want to remove your `.htaccess`, please remove the whole directory.

SebSeis:
I want to get a better google ranking of the w3id/ocqa, so people searching for the ontology can easier find it. Do you have other recommendations?

davidlehn:
This PR is removing all your redirect rules and adding a ".html" file that is not HTML.  This service doesn't host content, valid or not.  I assume removing all your redirects is a mistake?

I'd suggest two things.  First is to spend more effort on making a good project than worrying about SEO. :-)

Second is to open a new PR that will change your last rule to a catch-all for all other paths that goes to your site.  You can then host whatever SEO config files you want there.  I have no idea if Google respects redirects for their magic files.  If not, sorry.  If so, solved.

Something like this:
```diff
-RewriteRule ^$ https://sebseis.github.io/OCQA/ [R=303,NE,L]
+RewriteRule ^(.*)$ https://sebseis.github.io/OCQA/$1 [R=303,NE,L]
```

SebSeis:
Thanks for support

PR #3334, opened 2023-05-09
ExpO: .htaccess update

mozzherina:


davidlehn:
Do the rules in this file work?   My curl testing wasn't working.  The `mod_alias` [RedirectMatch](https://httpd.apache.org/docs/2.4/mod/mod_alias.html#redirectmatch) directive used here matches against the full URL path.  It's looking for `expo` but this dir is `ExpO`.  So it will never match?  Or am I doing something wrong?

You might consider using `mod_rewrite` features, like most other rules here, which uses relative path matching.

PR #3339, opened 2023-05-10
Added emerging with redirects

Michael-Ixo:


Michael-Ixo:
@davidlehn Added github ids as requested

PR #3341, opened 2023-05-12
Adding /semanticarts to w3id.org

Jamie-SA:


semanticarts/.htaccess
```diff
@@ -0,0 +1,18 @@
+RewriteEngine on
+
+# w3id.org/semanticarts/gist
+# This is the gist namespace, we are redirecting it to a web page
+# RewriteRule ^semanticarts/gist$ https://ontologies.semanticarts.com/o/namespace.html [R=302,L]
+RewriteRule ^semanticarts/gist/$ https://ontologies.semanticarts.com/o/namespace.html [R=302,L]
```
dgarijo, line 6: you need to remove `semanticsarts/` from the rewrite rule, as you already have the folder with `semanticsarts`. Otherwise you will be creating the w3id for https://w3id.org/semanticsarts/semanticsarts/gist

dgarijo:
@Jamie-SA please see my comments and confirm if that's what you really want (it does not look like from your readme)

PR #3348, opened 2023-05-17
Update /fst/ namespace | Add a second redirect step backend

test123-all:
Hello,
We did our first successful test and want to update the  /fst/ namespace to a new backend now. That backend is a self developed second redirect webpage based on a git-data-repository structure and GitLab webpages.

If you want to test the service please let me know. I don't know how much I'm allowed to explain.

I didn't squash the commits to make it easier for you to follow along. If you are going to squash them I would suggest the pull-request name "Update /fst/ namespace | Add a second redirect step backend" as commit name.

Thank you for your help


test123-all:
Thank you for merging my pull request. 

And also thank you very much for the additional information.
I just took a look back at the main README.md about the squashing of commits. I completely overlooked the "if appropriate". I didn't know about the downsides of squashing commits either, so thank you for explaining it to me.

PR #3354, opened 2023-05-21
remove AllowEncodedSlashes in .htaccess file

smrgeoinfo:
per @davidlehn comment ""'AllowEncodedSlashes' is not allowed in htaccess files and the server will just log errors and return a 500 error.'


dgarijo:
@smrgeoinfo,
you are modifying the AIRO ontology, where you are not listed as a maintainer. Maybe this is an error?

PR #3360, opened 2023-05-23
get rid of URL encoding, ARDC RVA made fix on their side 

smrgeoinfo:
I think its set for special case with the geochem technique URIs?  rewrites work in testing...

### DIFF
--- a/geochem/1.0/.htaccess
+++ b/geochem/1.0/.htaccess
@@ -3,6 +3,6 @@ RewriteEngine on
 
 # .htaccess rewrite rule tester (doen't do HTTP:Accept header conditions)  https://htaccess.madewithlove.be/
 
-RewriteRule    ^analyticalmethod/([A-Za-z_]+).html$     "https://demo.vocabs.ardc.edu.au/repository/api/lda/earthchem-georoc/analytical-methods-for-geochemistry/0-1/resource?uri=http://w3id.org/ogeochem/def/1/analyticalmethod/method/$1"  [R=303,L,NE]
-RewriteRule    ^analyticalmethod/([A-Za-z_]+)$     "https://demo.vocabs.ardc.edu.au/repository/api/lda/earthchem-georoc/analytical-methods-for-geochemistry/0-1/resource?uri=http://w3id.org/ogeochem/def/1/analyticalmethod/method/$1"  [R=303,L,NE]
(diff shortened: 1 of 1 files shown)


geochem/1.0/.htaccess
```diff
@@ -3,6 +3,6 @@ RewriteEngine on
 
 # .htaccess rewrite rule tester (doen't do HTTP:Accept header conditions)  https://htaccess.madewithlove.be/
 
-RewriteRule    ^analyticalmethod/([A-Za-z_]+).html$     "https://demo.vocabs.ardc.edu.au/repository/api/lda/earthchem-georoc/analytical-methods-for-geochemistry/0-1/resource?uri=http%3A%2F%2Fw3id.org%2Fogeochem%2Fdef%2F1%2Fanalyticalmethod%2Fmethod%2F$1"  [R=303,L,NE]
-RewriteRule    ^analyticalmethod/([A-Za-z_]+)$     "https://demo.vocabs.ardc.edu.au/repository/api/lda/earthchem-georoc/analytical-methods-for-geochemistry/0-1/resource?uri=http%3A%2F%2Fw3id.org%2Fogeochem%2Fdef%2F1%2Fanalyticalmethod%2Fmethod%2F$1"  [R=303,L,NE]
+RewriteRule    ^analyticalmethod/([A-Za-z_]+).html$     "https://demo.vocabs.ardc.edu.au/repository/api/lda/earthchem-georoc/analytical-methods-for-geochemistry/0-1/resource?uri=http://w3id.org/ogeochem/def/1/analyticalmethod/method/$1"  [R=303,L,NE]
+RewriteRule    ^analyticalmethod/([A-Za-z_]+)$     "https://demo.vocabs.ardc.edu.au/repository/api/lda/earthchem-georoc/analytical-methods-for-geochemistry/0-1/resource?uri=http://w3id.org/ogeochem/def/1/analyticalmethod/method/$1"  [R=303,L,NE]
```
davidlehn, line 7: - `ogeochem` doesn't exist, should this be `geochem`?
- That `def/1/...` path also isn't handled by `geochem` redirects here, but that's fine if you don't need it.
- You should use `https` here if possible.  Dereferencing a `http` URL here will get redirected to `https`, so it's an extra request.  (Even though those paths won't redirect at all right now)

smrgeoinfo: I changed http to https in the URIs, also updated to the current/final URI syntax 
https://w3id.org/geochem/1.0/analyticalmethod/{methodname}. This is in a0c9819 push.

PR #3367, opened 2023-05-25
changing host for application

PietroLiuzzo:
Dear all, the application has moved server and domain as it will be maintained by the same person at a different institution. Thanks a lot!


PietroLiuzzo:
Thanks a lot @dgarijo !

PR #3396, opened 2023-06-09
Add thor-project folder

pedropaulofb:
Please add thor-project folder that includes redirections for two different repositories and related content. Please let me know If it is necessary to create two different folders so I can make the necessary modifications.


dgarijo:
There is no htaccess file. Please add one with the current redirections.

PR #3408, opened 2023-06-15
rml: implement content negotiation

DylanVanAssche:
Based on https://github.com/perma-id/w3id.org/blob/master/devops-infra/def/core/.htaccess


dgarijo:
I do not think this is the right way to set up your redirections @DylanVanAssche.
First, you have things like `RewriteRule ^StarMap/?$ https://kg-construct.github.io/rml-star/ontology/documentation/406.html#http://w3id.org/rml/StarMap [NE,L,R=406]`, which is a redirection error.

Second, you may want to add a general rule, such as:
```
RewriteRule ^$ https://kg-construct.github.io/rml-star/ontology/documentation/406.html#http://w3id.org/rml/$1 [NE,L,R=303]
```
for each term you have pointing to your vocabulary.
Have you tested this locally? It's too big of an .htaccess to maintain...

davidlehn:
- I have to agree about the too big part.
- I think this is the first case of a gigantic auto-generated .htaccess file.  It does look unmaintainable.  I assume the idea is you'd just update the input and recompile it.  So it would "work" but is not friendly to look at and us admins are unable to review it.
- I'm not sure what the effect on the service itself is.  Maybe insignificant?  Apache has to load all that stuff at least.  And it processes rules logically in order.  Maybe that's all optimized so huge rule sets are fine, I don't know.
- What have you tried to simplify this?  Everyone else uses pattern matching so they only need a handful of rules at most.  Is there some setup that would allow you to drastically reduce the rules with that technique?  If not, can you please explain why?
- I feel it's pushing the limits of what this service should offer.  In general if this sort of thing is needed I'd be more inclined to suggest a trivial top level rule here that let's you use a stable id but redirects everything to your own Apache, or other, server that can do whatever complex processing you need to redirect again.  That would avoid churn and complex rules here.

dgarijo:
@davidlehn I think the problem they have is the following:
- They have divided their ontologies in modules. 
- All modules share the same ontology URI, but each module resolves to different files.
- They want to keep the same w3id for all modules, because it's the same ontology.

The only way of doing this is to do a redirection based on terms if they don't want to merge all modules in a single one.
I would recommend maybe creating a regular expression grouping all the terms in a module, in order to simplify the redirection.
Also, you may remove the 406 links per term.

DylanVanAssche:
> First, you have things like RewriteRule ^StarMap/?$ https://kg-construct.github.io/rml-star/ontology/documentation/406.html#http://w3id.org/rml/StarMap [NE,L,R=406], which is a redirection error.

This is redirection for when none of the content negotiation things match that you get a HTTP 406.

> Second, you may want to add a general rule, such as:

> RewriteRule ^$ https://kg-construct.github.io/rml-star/ontology/documentation/406.html#http://w3id.org/rml/$1 [NE,L,R=303]

> for each term you have pointing to your vocabulary.

Will try to generalize it.

> I think this is the first case of a gigantic auto-generated .htaccess file. It does look unmaintainable. I assume the idea is you'd just update the input and recompile it. So it would "work" but is not friendly to look at and us admins are unable to review it.

That's my concern as well... However, I have no idea how to make this better.

> What have you tried to simplify this? Everyone else uses pattern matching so they only need a handful of rules at most. Is there some setup that would allow you to drastically reduce the rules with that technique? If not, can you please explain why?

As @dgarijo mentioned in the next comment, we use 1 prefix with all modules, thus all properties & resources end up in the root of the prefix. So for every resource or property, we need a redirection. I tried to make it 'pretty' and added comments per resource/property to make it more readable. Minimizing, not sure how to.

> I would recommend maybe creating a regular expression grouping all the terms in a module, in order to simplify the redirection.

That's something we could do maybe, some redirections are continuously the same. Would that solve the maintainability problem?

DylanVanAssche:
>  But this is the point: you don't do a not found per term, but per document. Think about each rule as an if. If you arrive to the final rule, it means nothing matched, and you can return the 406.

Yes, I can make this one rule, but doesn't solve the maintainability I guess because it will remain big?

dgarijo:
It does not solve the maintainability problem, but reduces it (taking out as many lines as you have terms)

dgarijo:
Another small thing: you have used `http://w3id.org/rml/` in your spec, but it should be `https`  no?

dgarijo:
@DylanVanAssche any news?

DylanVanAssche:
@dgarijo Not yet, I have some holidays now. However, we are looking into getting a server to host most of the rules there instead of here, as suggested above.

DylanVanAssche:
@dgarijo @davidlehn I addressed the concerns by rewriting the rules. They should be much more clear and smaller now. Please let me know what you think :)

PR #3409, opened 2023-06-15
PURL for DeMato Ontology 

korchiev:
Creating persistent uri for the ontology developed by our lab

### DIFF
--- a/demato/.htaccess
+++ b/demato/.htaccess
@@ -1,1 +1,1 @@
-Redirect 302 / https://github.com/korchiev/Demato-Ontology
+RewriteRule ^(.*)$  / https://github.com/korchiev/Demato-Ontology  [R=302,L]
--- a/demato/.htaccess
+++ b/demato/.htaccess
@@ -1,1 +1,18 @@
+# Turn off MultiViews
+Options -MultiViews
+
+# Directive to ensure *.rdf files served as appropriate content type,
+# if not present in main apache config
+AddType application/rdf+xml .rdf
+AddType application/rdf+xml .owl
+AddType text/turtle .ttl
+AddType application/n-triples .n3
+AddType application/ld+json .json
+
+RewriteEngine on
+
+RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+xml)
+RewriteCond %{HTTP_ACCEPT} text/html [OR]
+RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
+RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
 RewriteRule ^$  / https://github.com/korchiev/korchiev.github.io/index.html  [R=303,L]
--- a/demato/.htaccess
+++ b/demato/.htaccess
@@ -15,4 +15,9 @@ RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+
 RewriteCond %{HTTP_ACCEPT} text/html [OR]
 RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
 RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
-RewriteRule ^$  / https://github.com/korchiev/korchiev.github.io/index.html  [R=303,L]
+RewriteRule ^$ https://github.com/korchiev/korchiev.github.io/  [R=303,L]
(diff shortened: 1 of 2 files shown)


dgarijo:
No `.htaccess` file with the redirection is uploaded. Maybe you forgot?

korchiev:
Oh yes I did. Thank you for pointing out. Will add it shortly

demato/.htaccess
```diff
@@ -0,0 +1 @@
+Redirect 302 / https://github.com/korchiev/Demato-Ontology
```
davidlehn, line 1: You need to use a full path if you are using mod_alias.  mod_redirect uses relative paths.
https://httpd.apache.org/docs/2.4/mod/mod_alias.html#redirect

Nkorchi: I am fairly new to this. I have created github.io page and I want this repo to redirect there? What command should I use?

Thank you so much for your help!

korchiev: I am fairly new to this. I have created github.io page and I want this repo to redirect there? What command should I use?

Thank you so much for your help!

davidlehn:
Please coordinate with https://github.com/perma-id/w3id.org/pull/3410 so there is only one similar name.

demato/.htaccess
davidlehn: There's a slightly complex example here: https://github.com/perma-id/w3id.org/tree/master/example/.  As needed you may be able to strip out RewriteCond/RewriteRules lines you don't need and end up with something like the last "default" line in there.  There are many other ids you could look at for inspiration too.

korchiev: I have followed an example. Could you please verify that?

My logic was if someone accesses https://w3id.org/demato with empty URL path it will redirect to github.io page. 

Is that right?

demato/.htaccess
```diff
@@ -0,0 +1,23 @@
...
+
+RewriteEngine on
+
+RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+xml)
+RewriteCond %{HTTP_ACCEPT} text/html [OR]
+RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
+RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
+RewriteRule ^$ https://github.com/korchiev/korchiev.github.io/  [R=303,L]
+RewriteCond %{HTTP_ACCEPT} ^.*application/rdf\+xml.* 
+RewriteRule ^$ https://github.com/korchiev/korchiev.github.io/ontology.owl [R=303,NE,L]
```
dgarijo, line 20: This link is a 404

demato/.htaccess
```diff
@@ -0,0 +1,23 @@
...
+AddType application/n-triples .n3
+AddType application/ld+json .json
+
+RewriteEngine on
+
+RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+xml)
+RewriteCond %{HTTP_ACCEPT} text/html [OR]
+RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
+RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
+RewriteRule ^$ https://github.com/korchiev/korchiev.github.io/  [R=303,L]
```
dgarijo, line 18: This is not serving the HTML representation of your ontology. Please consider serving the documentation through github pages or similar

demato/.htaccess
```diff
@@ -0,0 +1,23 @@
...
+RewriteCond %{HTTP_ACCEPT} ^.*application/rdf\+xml.* 
+RewriteRule ^$ https://github.com/korchiev/korchiev.github.io/ontology.owl [R=303,NE,L]
+RewriteCond %{HTTP_ACCEPT} ^.*text/turtle.* 
+RewriteRule ^$ https://github.com/korchiev/korchiev.github.io/ontology.ttl [R=303,NE,L]
```
dgarijo, line 22: Link is a 404

dgarijo (changes requested):

demato/.htaccess
korchiev: Thank you! I have fixed it now (I think)

demato/.htaccess
korchiev: fixed

dgarijo (approved):

dgarijo:
FYI, you can activate github pages on any repo

korchiev:
Thank you for guidance! Later I will activate on another page!

PR #3413, opened 2023-06-17
update rewrite for published location

smrgeoinfo:
vocab going public


smrgeoinfo:
the airo files are still causing weird problems

dgarijo:
@smrgeoinfo what is the problem you face? Can't you pull the latest update and build from there?
I do not think the solution should be forcing everyone to ignore a vocabulary that you do not own. If the authors want to add changes to that folder it may be problematic.

smrgeoinfo:
I sync'd my fork with master, and they didn't go away. I'll try a clean pull.

dgarijo:
Thanks!

PR #3446, opened 2023-07-06
update htaccess

junjun7613:
Updated .htaccess because so far, server errors occurred during redirection..

### DIFF
--- a/HIMIKO/.htaccess
+++ b/HIMIKO/.htaccess
@@ -1,4 +1,4 @@
 Options +FollowSymLinks
 RewriteEngine on
 
-RewriteRule ^himiko.owl$ https://junjun7613.github.io/MicroKnowledge/himiko.owl [R=302, L]
+RewriteRule ^(.*)$ https://junjun7613.github.io/MicroKnowledge/$1 [R=302,L]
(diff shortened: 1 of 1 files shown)


HIMIKO/.htaccess
```diff
@@ -1,4 +1,4 @@
 Options +FollowSymLinks
 RewriteEngine on
 
-RewriteRule ^(.*)$ https://junjun7613.github.io/MicroKnowledge/$1 [R=302, L]
\ No newline at end of file
+RewriteRule ^himiko.owl$ https://junjun7613.github.io/MicroKnowledge/himiko.owl [R=302, L]
```
dgarijo, line 4: I do not think you want to add the `hiimiko.owl` in the first part of the rewrite rule. Otherwise your w3id will be:
`https://w3id.org/HIMIKO/himiko.owl` only

davidlehn, line 4: The original is fine except you need to remove the space in the options.
```suggestion
RewriteRule ^(.*)$ https://junjun7613.github.io/MicroKnowledge/$1 [R=302,L]
```

junjun7613: Thank you for your comment! I did remove the space from the original.

junjun7613: I re-uploaded the original but removing the space in the options.

junjun7613: Hope this will work well..

PR #3447, opened 2023-07-06
add rewrite rule to geno

elespdn:


geno/.htaccess
```diff
@@ -4,4 +4,6 @@ RewriteEngine on
 
 RewriteRule ^$ https://gen-o.github.io/ [R=302,L]
 
-RewriteRule (.*) https://gen-o.github.io/$1 [R=302,L,QSA]
\ No newline at end of file
+RewriteRule (.*) https://gen-o.github.io/$1 [R=302,L,QSA]
+
+RewriteRule ^CommentEnMer https://comment-enmer.github.io [R=302,L]
```
davidlehn, line 9: Rules are processed in order.  This needs to be above the catch all one above.  If doing so, you may consider `^CommentEnMer$` if you don't want to also match other trailing text.

elespdn: Thanks! I changed the order of the rules.

PR #3453, opened 2023-07-09
fixing directoryslash

athalhammer:
sorry, should have checked the syntax before


athalhammer:
I think this should work now

athalhammer:
Tested this now on another apache server

PR #3471, opened 2023-07-13
moved the folder for the dsv-ontology and added a README file

ritamargherita:
I have moved the folder for the dsv-ontology from under the odyssey folder, to its own individual folder. Also, added a READ me file


ritamargherita:
Added GitHub page as requested

PR #3508, opened 2023-07-28
Add new rewrite rule for `biopragmatics`

cthoyt:
This PR adds a rewrite rule to expose files in https://github.com/biopragmatics/biomappings (this is a repository I own) through the `biopragmatics` namespace (which I also own)

For example, we want https://w3id.org/biopragmatics/biomappings/biomappings.sssom.tsv to redirect to https://raw.githubusercontent.com/biopragmatics/biomappings/master/docs/_data/sssom/biomappings.sssom.tsv

PR #3510, opened 2023-07-29
Add new rewrite rule for biopragmatics

cthoyt:
Redo of #3508

This PR adds a rewrite rule to expose files in https://github.com/biopragmatics/biomappings (this is a repository I own) through the biopragmatics namespace (which I also own)

For example, we want https://w3id.org/biopragmatics/biomappings/biomappings.sssom.tsv to redirect to https://raw.githubusercontent.com/biopragmatics/biomappings/master/docs/_data/sssom/biomappings.sssom.tsv


cthoyt:
thanks @davidlehn

PR #3531, opened 2023-08-11
replace generic rule with specific ones

elespdn:


geno/.htaccess
```diff
@@ -4,7 +4,9 @@ RewriteEngine on
 
 RewriteRule ^CommentEnMer https://comment-enmer.github.io [R=302,L]
 
-RewriteRule ^$ https://gen-o.github.io/ [R=302,L]
+RewriteRule ^$ https://gen-o.github.io [R=302,L]
 
-RewriteRule (.*) https://gen-o.github.io/$1 [R=302,L,QSA]
+RewriteRule 2023-07-01 https://gen-o.github.io [R=302,L]
+
+RewriteRule website https://gen-o.github.io/website [R=302,L]
```
dgarijo, line 11: Are you sure this is ok? I think you need a "$" to indicate the end. Like `website$`

elespdn: Yes, better, thanks a lot

PR #3544, opened 2023-08-20
add pmd subpath for glass-ontology

SimonStier:
approved by @joergwa (see https://github.com/materialdigital/w3id/pull/1 )


dgarijo:
I suggest you add yourself as a maintainer so you don't need to seek approval constantly

PR #3549, opened 2023-08-24
w3Id for Coscine

mpolitze:
Please add a w3id for the research platform Coscine LDP endpoint and static RDF graphs. 

@LukasCBossert: It would be nice if you could give a short acknowledgement that you are ok to be a maintainer for this w3id.


LukasCBossert:
I am very much ok, being a maintainer for this w3id @mpolitze

PR #3553, opened 2023-08-26
Update data-integrity security URL redirects.

msporny:
This PR updates the data-integrity security URL redirects to match the latest values that are changing in the spec.

### DIFF
--- a/security/.htaccess
+++ b/security/.htaccess
@@ -13,18 +13,24 @@ RewriteRule ^v3-unstable$ https://w3c-ccg.github.io/security-vocab/contexts/secu
 # https://w3id.org/security/data-integrity
 
 RewriteRule ^data-integrity$ https://www.w3.org/TR/vc-data-integrity/ [R=302,L]
-RewriteRule ^data-integrity/v1$ https://www.w3.org/ns/data-integrity/v1 [R=302,L]
+RewriteRule ^data-integrity/v1$ https://w3c.github.io/vc-data-integrity/contexts/data-integrity/v1 [R=302,L]
+RewriteRule ^data-integrity/v2$ https://w3c.github.io/vc-data-integrity/contexts/data-integrity/v2 [R=302,L]
 
 # https://w3id.org/security/multikey
 
-RewriteRule ^multikey$ https://w3c.github.io/vc-data-integrity/ [R=302,L]
-RewriteRule ^multikey/v1$ https://digitalbazaar.github.io/multikey-context/contexts/multikey-v1.jsonld [R=302,L]
+RewriteRule ^multikey$ https:/w3id.org/security/ [R=302,L]
+RewriteRule ^multikey/v1$ https://w3c.github.io/vc-data-integrity/contexts/multikey/v1 [R=302,L]
+
+# https://w3id.org/security/jwk
+
+RewriteRule ^jwk$ https:/w3id.org/security/ [R=302,L]
+RewriteRule ^jwk/v1$ https://w3c.github.io/vc-data-integrity/contexts/jwk/v1 [R=302,L]
 
 # https://w3id.org/security/suites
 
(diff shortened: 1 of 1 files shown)


dlongley (approved):

security/.htaccess
```diff
@@ -13,18 +13,24 @@ RewriteRule ^v3-unstable$ https://w3c-ccg.github.io/security-vocab/contexts/secu
 # https://w3id.org/security/data-integrity
 
 RewriteRule ^data-integrity$ https://www.w3.org/TR/vc-data-integrity/ [R=302,L]
-RewriteRule ^data-integrity/v1$ https://www.w3.org/ns/data-integrity/v1 [R=302,L]
+RewriteRule ^data-integrity/v1$ https://w3c.github.io/vc-data-integrity/contexts/data-integrity/v1 [R=302,L]
```
davidlehn, line 16: These are going to come through as `application/octet-stream`.  Would be nice if the source repos used `.jsonld` for the contexts so github serves them as `application/ld+json`.  Similar to the other contexts in this file.

security/.htaccess
```diff
@@ -13,18 +13,24 @@ RewriteRule ^v3-unstable$ https://w3c-ccg.github.io/security-vocab/contexts/secu
...
+
+RewriteRule ^jwk$ https:/w3id.org/security/ [R=302,L]
+RewriteRule ^jwk/v1$ https://w3c.github.io/vc-data-integrity/contexts/jwk/v1 [R=302,L]
 
 # https://w3id.org/security/suites
 
 # TODO: update registry to reflect this file.
 
-RewriteRule ^suites$ https://w3c-ccg.github.io/ld-cryptosuite-registry [R=302,L]
+RewriteRule ^suites$ https://w3c.github.io/vc-specs-dir/ [R=302,L]
```
davidlehn, line 33: Is the theory more suites will be listed at this page?  It doesn't list all the things in the previous ld-cryptosuite-registry, or all the suites below.

davidlehn (commented):
What's the timeline for when this should be merged?  Other than perhaps addressing my comments, should it be merged now?

security/.htaccess
msporny: hrm, yes, good point. Does Github pages allow you to set media types on files?

security/.htaccess
msporny: Yes, it will list more suites in time. The old ld-cryptosuite-registry is abandoned at this point, long live vc-specs-dir!

msporny:
> What's the timeline for when this should be merged? Other than perhaps addressing my comments, should it be merged now?

Sooner the better... we can clean up/fix in other PRs, especially since your suggestions don't seem like they're going to be implemented soon (due to having to get WG buy in to change/duplicate some of the content... need more time to think about it. Merging now will fix at least half of the problem (files not resolving at all / resolving to the wrong thing).

security/.htaccess
davidlehn: That github pages feature hasn't existed in the past.  I suspect it still does not.  Target file extensions fix the issue, people are probably not linking direct to those files, and it helps text editors know what syntax to use.

davidlehn: I think pages handles symlinks too, but that makes the target dirs kind of a mess.

davidlehn:
One reason to handle target URL changes first is to avoid tricky issues with downtime of live files.  One could publish the new targets, then update links here, then remove old targets.  But I suspect it would be more of a one shot rename and coordinated merge in both places to get near zero downtime.

msporny:
@davidlehn wrote:
>> What's the timeline for when this should be merged? Other than perhaps addressing my comments, should it be merged now?

@msporny wrote:
> Sooner the better... we can clean up/fix in other PRs

Ping @davidlehn -- can we get this merged, please?

PR #3564, opened 2023-09-05
DOLCE ontology

wterkaj:


wterkaj:
I've added the list of maintainers. Thanks.

Issue #1643, opened 2020-02-17
Improving examples.


davidlehn:
Some ideas for helping new folks figure out how to use this service.  I've had these ideas rattling around in my head for years, might as well note them here.

@dgarijo started with an ontology example in https://github.com/perma-id/w3id.org/pull/1639.  (That has been needed for a long time.  Thanks!)

- Long ago I had thought of having two example locations.  `example/` with a really basic example, more like a template for just having a `.htaccess` and `README.md` to get started and redirect something to https://example.com/.  Then have a secondary `examples/` that has a handful of subdirs for various use cases.  In this case the new `example` dir would be under something like `examples/ontology/`.
- Refer people to appropriate Apache docs (also noted in https://github.com/perma-id/w3id.org/issues/1530)
- Add other examples based on patterns we've seen other people using over the years.
  - Simple redirects.
  - Pattern matching and replacement.
  - All in one file cover subdirs vs multiple directories.
  - CORS compatability.
  - It hasn't been an issue before, but show how to avoid serving the local README.md if you don't have a wildcard redirect that covers that.
- Figure out which `.htaccess` options are commonly needed and document them and why they are there.
- The root README should point people at the examples.
- The main homepage should point people at the examples.
- Figure out if info and maintainer info in `.htaccess` is ok vs putting in `README.md`. Recently it was suggested to just use README.  I had originally thought it would have been the other way.
- Add github PR magic info file with link to or an inline checklist for common issues we've seen (no maintainer info, etc).

I'm sure there's more... if anyone has ideas, add a comment.

dgarijo:
Thanks, @davidlehn,
I created the example also because I am writing a longer tutorial on best practices for ontology publication (I am using https://w3id.org/example#). Would you rather have the URI located under example/ns? 

Right now it does not prevent other people from doing their own examples, but just in case you had something else in mind.

davidlehn:
@dgarijo I'm not sure what the layout should be.  Open to discussion.  I'm just guessing that since no one likes to read docs, they will immediately look to `example(s)/` type dirs and copy & modify those.  So those dirs should have some examples of current best practices for various use cases.  Ontology publication is certainly a good use case!  Unsure how to organize so people can see how to do simpler things too.

dgarijo:
Ok, my proposal then is as follows:
Let's create an "examples" directory. Inside, we can have "ontology", which will redirect at the "example"directory I already did (plus a few explanations on a readme file)
[...]

Issue #1530, opened 2019-10-27
a simple tutorial for common redirecting examples?


abubelinha:
I would appreciate a simple guide, or a list of .htaccess files, showing examples of how to do some quite common redirections.

w3id.org/myname  --->   my-current.domain/my-site-path/my-file.php
w3id.org/myname?a=1&b=4  ---> my-current.domain/my-site-path/my-file.php?a=1&b=4
w3id.org/myname/a_1/b_4  ---> my-current.domain/my-site-path/my-file.php?a=1&b=4
w3id.org/myname/site1  ---> my-current.domain/my-site-path/site1
w3id.org/myname/anothersite  ---> my-current.domain/my-site-path/anothersite

... and so on

Perhaps this is such a basic question, but there are no links in w3id.org to any good tutorial for beginners.

davidlehn:
Yes, this would be good.  It's been on my todo list since forever.  Lots of things have been copied around that are probably not optimal.  I think the examples should be live working ones and in the an "example" directory.

abubelinha:
Thanks @davidlehn 

Some other related thoughts:

- Nothing is said about "what is an .htaccess file"

I understand it stands for an Apache web server special file, used to handle redirections from urls named after the folder where the .htaccess file is placed. 

But I might be wrong: nothing is said about Apache, anywhere.  

- Should we look at any particular Apache web server version?
I don't know if the directives used inside .htaccess files could be different depending on the web server version.

bact:
- [x] w3id.org/myname ---> my-current.domain/my-site-path/my-file.php [Covered]
- [x] w3id.org/myname?a=1&b=4 ---> my-current.domain/my-site-path/my-file.php?a=1&b=4 [Covered; not by individual parameters, but the entire query string]
- [ ] w3id.org/myname/a_1/b_4 ---> my-current.domain/my-site-path/my-file.php?a=1&b=4
- [x] w3id.org/myname/site1 ---> my-current.domain/my-site-path/site1 [Covered]
- [x] w3id.org/myname/anothersite ---> my-current.domain/my-site-path/anothersite [Covered]

PR #3599, opened 2023-09-28
 examples: Examples of .htaccess and README.md 

bact:
* Move current `example/` to `examples/ontology/` - as proposed in #1643
* Add new `example/` that redirects to `https://example.com/` - as proposed in #1643
* Add new `examples/README.md` with examples of .htaccess and links to examples of README.md and examples/ontology/ - as proposed in #1530 and #1643

### DIFF
--- a/example/.htaccess
+++ b/example/.htaccess
@@ -1,74 +1,13 @@
-# Turn off MultiViews
-Options -MultiViews
-
-# Directive to ensure *.rdf files served as appropriate content type,
-# if not present in main apache config
-AddType application/rdf+xml .rdf
-AddType application/rdf+xml .owl
-AddType text/turtle .ttl
-AddType application/n-triples .n3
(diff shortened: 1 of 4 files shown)


dgarijo:
@bact,
@davidlehn is not the admin for `example`, I am. See [https://github.com/perma-id/w3id.org/tree/master/example](https://github.com/perma-id/w3id.org/tree/master/example)
I do not want to change the w3id for the example ontology, as I already used the ontology URI in different publications. The point of w3ids, in my opinion, is that they should not change. Therefore I am closing this pull request.
You are welcome to create an `examples` directory in a separate pull request.

bact:
Thank you @dgarijo. I will create a separated PR for that. Thank you for the suggestions.

bact:
Please use PR #3600 instead.

PR #3600, opened 2023-09-28
examples: Examples of .htaccess and README.md

bact:
* Add `examples/README.md` with examples of .htaccess and links to examples of README.md and examples/ontology/ - as proposed in #1530 and #1643
* Add `examples/ontology` that redirects to `https://w3id.org/example` - as proposed in #1643

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -188,4 +188,4 @@ An example of a good README file: [w3id.org/dggs/README.md](https://github.com/p
 ## Publish vocabularies with W3ID
 
 If you plan to publish a vocabulary/ontology with W3ID,
-see https://w3id.org/examples/ontology.
+see https://w3id.org/example/.
--- a/examples/README.md
+++ b/examples/README.md
@@ -36,10 +36,10 @@ You are encouraged to put a breif ID info and a maintainer info in the comment
 ```ApacheConf
 # Example
 #
-# https://w3id.org/example redirects to https://example.com/
+# https://w3id.org/examples/simple/ redirects to https://example.com/
 #
 # ## Contact
-# This space is administered by:  
+# This space is administered by:
 #
 # Firstname Secondname
 # email@example.com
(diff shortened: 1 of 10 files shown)


examples/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+### Put ID info and maintainer info inside .htaccess
+
+You are encouraged to put a breif ID info and a maintainer info in the comment
+(lines staring with `#` character) of a `.htaccess` file.
+
+```ApacheConf
+# Example
+#
+# https://w3id.org/example redirects to https://example.com/
```
dgarijo, line 39: This is not correct.

examples/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+GitHub will automatically display the content of a `README.md` to repository visitors.
+
+An example of a good README file: [w3id.org/dggs/README.md](https://github.com/perma-id/w3id.org/blob/master/dggs/README.md)
+
+
+## Publish vocabularies with W3ID
+
+If you plan to publish a vocabulary/ontology with W3ID,
+see https://w3id.org/examples/ontology.
```
dgarijo, line 191: Wouldn't it be better to link to /example/ directly? The one in /examples is a redirect.

dgarijo:
@bact 
I added small comments. Also, aren't you missing the `.htaccess` file redirecting to example.org in the /examples?

examples/README.md
bact: Thank you. That's better. I have changed that accordingly.

examples/README.md
bact: Thank you. I have fixed all the instances that mentioned https://w3id.org/example wrongly.

In that example, I replace it with w3id.org/examples/simple instead.

For .htaccess in examples/, I am not sure where to redirect it to.

bact:
> @bact I added small comments. Also, aren't you missing the `.htaccess` file redirecting to example.org in the /examples?

For .htaccess in examples/, I am not sure where to redirect it to.

bact:
I'm also not sure about what is the best practice on HTTP status, which one is recommended for W3ID use case: 302, 303, or 307? Or it really depends on each ID use case?

dgarijo:
I think it depends on the use case. 303 is the recommended for semantic resources at least.

bact:
> > @bact I added small comments. Also, aren't you missing the `.htaccess` file redirecting to example.org in the /examples?
> 
> For .htaccess in examples/, I am not sure where to redirect it to.

decided to redirect it to https://github.com/perma-id/w3id.org/tree/master/examples - see #3601

PR #3623, opened 2023-10-09
Update .htaccess

firmao:


firmao:
It was a mistake with an extra "/" on the URL. PLease, do the merge with the master branch as soon as possible. Thanks in advance!

PR #3628, opened 2023-10-10
minting new identifier: FCB085

proccaserra:


dgarijo:
This does not seem to be redirecting to anything. Is there something wrong?

proccaserra:
many thx @dgarijo

PR #3636, opened 2023-10-11
Introduce istat/ path under italia

RRRoby:


RRRoby:
Introduce istat/ path under italia

RRRoby:
removed the file extension from the htaccess files

dgarijo:
@RRRoby, you added this PR on a namespace you don't control. I missed this because you were adding a folder, and did not realize the parent folder already existed. But if it's not vetted by the original maintainers,  as raised in #3799.

Please provide an alternate parent folder by Dec 27th, or I will move this to /istat (outside the parent folder)

PR #3654, opened 2023-10-18
renamed project folder

lisestork:


lisestork:
I agree, but in this case deleting it is fine: we are preparing the camera ready so the namespace will be updated everywhere before publication.

PR #3657, opened 2023-10-18
Inrupt: Add opentelemetry vocab

davidboweninrupt:
This is used by our tooling so we want to make it public. It's not endorsed by https://opentelemetry.io/


davidboweninrupt:
Too many commits.

PR #3658, opened 2023-10-18
Remove documentation of link checking

davidboweninrupt:
The checking [was disabled some time ago](https://github.com/perma-id/w3id.org/commit/7636d95dff670b178ef2b4ac17196af64c2f4bf1) so it's probably best to remove references to that in the documentation.


davidlehn:
Thanks.  Should have removed those docs long ago.  I had idea on how to better do that functionality, but the service seems to get by without it.

PR #3661, opened 2023-10-20
Update .htaccess

nanoyan:
Thank you for your suggestion!


iqb/.htaccess
```diff
@@ -1,4 +1,7 @@
 RewriteEngine on
+
+RewriteRule ^profile-registry https://raw.githubusercontent.com/iqb-vocabs/profile-registry/master/registry.csv [R=302,L]
+RewriteRule ^v([0-9]+)/(.*) https://skohub.io/iqb-vocabs/v$1/heads/master/w3id.org/iqb/v$1/$2 [R=302,L]
 RewriteRule ^v25/(.*) https://skohub.io/iqb-vocabs/v25/heads/master/w3id.org/iqb/v25/$1 [R=302,L]
```
davidlehn, line 5: All these explicit v# rules should be removed since they will never match due to the new wildcard rule.

nanoyan: Now, I erased the explicit rules. Is it ok? Thank you

PR #3685, opened 2023-11-02
Chalk Research Group namespace creation

stuchalk:
Create initial redirections for the Chalk Research Group resources.

### DIFF
--- a/chalkgrp/.htaccess
+++ b/chalkgrp/.htaccess
@@ -1,0 +1,39 @@
+RewriteEngine On
+
+# The cookbook itself
+# -------------------
+# The identifier https://w3id.org/chalkgrp is for the cookbook itself:
+#
+RewriteRule ^$   https://stuchalk.domains.unf.edu/ [R=303,L]
+#
+# Catch https://w3id.org/chalkgrp/ and remove the slash; the slashed version shall not have meaning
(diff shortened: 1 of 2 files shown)


chalkgrp/.htaccess
```diff
@@ -0,0 +1,39 @@
+RewriteEngine On
+
+# The cookbook itself
+# -------------------
+# The identifier https://w3id.org/chalkgrp is for the cookbook itself:
+#
+RewriteRule ^$   https://stuchalk.domains.unf.edu/ [R=303,L]
+#
+# Catch https://w3id.org/chalkgrp/ and remove the slash; the slashed version shall not have meaning
+RewriteRule ^/$   https://w3id.org/chalkgrp [R=301,L]
```
davidlehn, line 10: Can you say more about what this is for?  Without doing some testing, I would have thought this never gets used due to default processing of trailing slashes.

stuchalk: I actually got this from the faircookbook .htaccess file.  Reading up on the pros and cons of adding a trailing slash it just makes the request faster as the slash indicates a directory and thus time is spent checking for the directory.  Looking across all the files in this repo there are a couple of others that use this approach.  I can remove if its a problem...

davidlehn: Not a problem.  I haven't noticed that pattern before and though it looked odd.

PR #3691, opened 2023-11-06
Update .htaccess (#2)

r0sek:
Updated with new Rewrite Rule

### DIFF
--- a/CMECS/.htaccess
+++ b/CMECS/.htaccess
@@ -7,5 +7,4 @@
 # GitHub username: r0sek
 
 RewriteEngine on
-RewriteRule ^ https://mmisw.org/ont/~katerose/CMECS [R=302,L]
-RewriteRule ^(.*)$ https://mmisw.org/ont?iri=$1 [R=302,L]
+RewriteRule ^(.*)$ https://mmisw.org/ont?iri=https://w3id.org/$1 [R=302,L]


CMECS/.htaccess
```diff
@@ -7,5 +7,4 @@
 # GitHub username: r0sek
 
 RewriteEngine on
-RewriteRule ^ https://mmisw.org/ont/~katerose/CMECS [R=302,L]
-RewriteRule ^(.*)$ https://mmisw.org/ont?iri=$1 [R=302,L]
+RewriteRule ^(.*)$ https://mmisw.org/ont?iri=https://w3id.org/$1 [R=302,L]
```
davidlehn, line 10: The matched pattern is relative to the current directory.  You'll want to add your prefix into the target URL:
```suggestion
RewriteRule ^(.*)$ https://mmisw.org/ont?iri=https://w3id.org/CMECS/$1 [R=302,L]
```

r0sek: I did try what you suggested, but when I test ex. https://w3id.org/CMECS/CMECS_00000001 in https://htaccess.madewithlove.com/ with your rule `RewriteRule ^(.*)$ https://mmisw.org/ont?iri=https://w3id.org/CMECS/$1 [R=302,L]` it returns https://mmisw.org/ont?iri=https://w3id.org/CMECS/CMECS/CMECS_00000001, which returns "IRI not found"

So I tested the same with `RewriteRule ^(.*)$ https://mmisw.org/ont?iri=https://w3id.org/$1 [R=302,L]` and it returns https://mmisw.org/ont?iri=https://w3id.org/CMECS/CMECS_00000001, which resolves correctly. Per the mmisw admin, it has something to do with the way they handle re-hosted resources.

dgarijo: I think @davidlehn is right, this won't work. I will approve if you want and see the results

davidlehn: @r0sek I think that tool is testing as if the rules are at the top level of the server.  Which is useful to get close, but you do need to adjust due to the rules being in a subdir.  You'll need to make another PR to fix this.
@dgarijo Probably didn't need to merge this until it was fixed, but yeah, it is not doing the right thing now that it's live.

r0sek: It's working for me.

PR #3693, opened 2023-11-07
Update .htaccess

r0sek:
Updating Rewrite Rule with directory prefix


r0sek:
Yes, this is working now. Thanks so much for all your help!!

PR #3729, opened 2023-11-27
Renaming SPEECH to PODIO

anaigmo:


dgarijo:
You may want to leave the old w3id and a redirect, since w3ids are supposed to be persistent

anaigmo:
In this case the previous w3id is not going to be used anywhere and is not published, so it is fine, thanks!

PR #3731, opened 2023-11-27
Update .htaccess

dwinston:
pass-through to dynamic api.microbiomedata.org implementation for resolution by type, rather than attempt a hairy and update-needing regex for entity types.

use 307 redirects

remove commented-out lines

/cc @turbomam @cmungall


dwinston:
thank you!

PR #3739, opened 2023-11-29
Creation of two new redirects.

TheRazorace:
Creation of two new redirects for upcoming publication submission. They will updated with more information after submission, and official links after submission.

### DIFF
--- a/mlseakg/README.MD
+++ b/mlseakg/README.MD
@@ -1,0 +1,10 @@
+MLSO: Machine Learning Sailor Ontology
+
+Website:
+https://w3id.org/mlso
+
+Contacts:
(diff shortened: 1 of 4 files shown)


dgarijo:
@TheRazorace 
please add a readme file on each of the redirects following the instructions in our readme file (i.e., small description and who the maintainers for that w3id will be). Thanks!

mlseakg/README.MD
```diff
@@ -0,0 +1,10 @@
+MLSO: Machine Learning Sailor Ontology
```
dgarijo, line 1: This does not seem to be the right readme. It says mlso, but the w3id is mlseakg

PR #3740, opened 2023-11-29
Request for two new redirects for upcoming publications. 

TheRazorace:
Request for two new redirects for upcoming publications. Added simple README.md, thanks for the feedback. More information and redirect links will be added in the near future. Thank you!


dgarijo:
Target redirects do not work. Are you aware of this?

TheRazorace:
Hello! what do you mean?

dgarijo:
That when I go to the target URL, I get redirected to https://gitlab.kuleuven.be/users/sign_in. Not sure if that's what you are intending, especially since this is for a publication

Issue #1629, opened 2020-02-10
Travis tests are disabled?


namedgraph:
It looks like `travis.yml` is all commented out and not testing anything?
https://travis-ci.org/perma-id/w3id.org/jobs/648154777/config?utm_medium=notification&utm_source=github_status

Our PR passed but we got non-working links anyway.

dgarijo:
Can you please specify which links are not working? Travis has not been working for a while, and should not be a reference. For example https://w3id.org/atomgraph/linkeddatahub/domain seems to work fine

davidlehn:
Yes, I disabled that long ago because it was causing too much trouble.  For every PR it was scanning for all w3id.org URLs in all READMEs.  Peoples redirects rot over time so there are always a bunch of failing URLs.  That caused everyone to get confused.  Also the link checker itself had some odd TLS network problems on the travis infrastructure, and I failed to debug it.  If someone would like to update the script to just check the files in a PR, and figure out why the link checker is sometimes broken, that would be great, and we'll re-enable PR testing.

davidboweninrupt:
I've created #3656 so people don't think the linkchecker is running.

davidlehn:
Link checking docs have been removed.  The link checker won't likely be enabled like it was before.  At least not anytime soon.  Closing.

PR #3786, opened 2023-12-15
add redirect test workflow / github action

simontaurus:
Currently there are no or only [limited](https://htaccess.madewithlove.com/) pre-deployment testing options (see also https://github.com/perma-id/w3id.org/issues/1629).

This pullrequest adds a workflow that can be run in the main repo, any fork or on local checkouts if docker is available.

Test specifications are currently extracted from .htaccess file comments, e. g. 
```
##TESTv1 '/mypath --header "Accept: text/html"' "https://my-target-domain.com/test.html"
```
Example: https://github.com/perma-id/w3id.org/commit/4892875ccbf88e246ce629d1461c876943e1b1a8

Contributors can run the github action in their forks selecting only the subpath of interest:
grafik
Example: https://github.com/OpenSemanticWorld/w3id.org/actions/workflows/redirect_tests.yml

Run will fail if at least one test is not successful, log will show errors e.g. 
```
Processing file /usr/local/apache2/htdocs/emmo/.htaccess
TEST '/emmo/domain-test --header "Accept: text/html"' => 'https://emmo-repo.github.io/domain-test/test.html'
ERROR  : Result is 'https://raw.githubusercontent.com/emmo-repo/EMMO/domain-test.ttl'
TEST '/emmo/domain-test --header "Accept: application/rdf\+xml"' => 'https://emmo-repo.github.io/domain-test/test.ttl'
ERROR  : Result is 'https://raw.githubusercontent.com/emmo-repo/EMMO/domain-test.ttl'
Error: Process completed with exit code 1.
```
Example: https://github.com/OpenSemanticWorld/w3id.org/actions/runs/7224639650/job/19686349546

More details can be found here: https://github.com/OpenSemanticWorld/w3id.org/blob/master/.test/README.md

Let me know if that makes sense to you and if I should stash my commits


.test/README.md
```diff
@@ -8,6 +8,14 @@ Test specifications are currently extracted from .htaccess file comments, e. g.
 ##TESTv1 '/mypath --header "Accept: text/html"' "https://my-target-domain.com/test.html"
 ```
 
+## Run the test as Github Action
+1. Navigate to `https://github.com/<your fork>/w3id.org/actions/workflows/redirect_tests.yml` (example see [here](https://github.com/OpenSemanticWorld/w3id.org/actions/workflows/redirect_tests.yml))
+1. Select 'Run workflow'
+1. Keep the master branch or select your custom branch
+1. Change the subpath to your subpath, e.g. `/mypath`
```
TallTed, line 15:
```suggestion
1. Accept the master branch, or select your custom branch
1. Change the subpath to your subpath, e.g., `/mypath`
```

.test/README.md
```diff
@@ -0,0 +1,32 @@
+# Redirect test
+
+Tests running directly on the source .htaccess files. It does only test the redirect, not the existance of the resource redirected to.
+This is done by mounting the w3id.org repo in an apache docker and test the redirects with curl.
+
+Test specifications are currently extracted from .htaccess file comments, e. g.
```
jesper-friis, line 6:
```suggestion
Test specifications are currently extracted from .htaccess file comments, e.g. 
```

davidlehn:
This is an interesting testing approach.  Thanks.  We'll probably need to think this over a bit.

- Probably could slim down that apache config since it's mostly comments.  I can check our production config to line things up so it's doing realistic testing.
- I didn't know workflows could have inputs like that.
- Based on the old travis-ci linkchecker experiment, we don't want to default to running all the tests in every dir.  As this service grows, so do bitrotten configs and links.  Need to avoid other peoples broken setups causing false positive failures for a PR.
- For a long while, I've had ideas for other checks that should also be run.  For that I wanted to do something to look at the PR changed files and only run tests in those dirs (or whatever makes sense).  So if you you edit files in `/foo` and `/bar` in a commit, it would only check those dirs.  Seems like something similar could be applied here.  I'm not sure the implementation details of that.
- Ideally we'd run this sort of thing by default.  To avoid wasting cycles, probably would want a pre-action that quickly checks what tests should even run.  No reason to spin up docker and apache if no tests exist.
- For my other project ideas, I was considering a `.w3id.json` (or .yaml, .etcetc) config.  Do you think it would be better to have test pragmas inline in .htaccess?  Or would a more easily parsable file with test info be ok?  I can see pros/cons there.

.test/conf/httpd.conf
```diff
@@ -0,0 +1,551 @@
...
+#LoadModule dav_lock_module modules/mod_dav_lock.so
+#LoadModule vhost_alias_module modules/mod_vhost_alias.so
+#LoadModule negotiation_module modules/mod_negotiation.so
+LoadModule dir_module modules/mod_dir.so
+#LoadModule imagemap_module modules/mod_imagemap.so
+#LoadModule actions_module modules/mod_actions.so
+#LoadModule speling_module modules/mod_speling.so
+#LoadModule userdir_module modules/mod_userdir.so
+LoadModule alias_module modules/mod_alias.so
+LoadModule rewrite_module modules/mod_rewrite.so
```
simontaurus, line 199: Diffs to the default config: Rewrite modul must be enabled

.test/conf/httpd.conf
```diff
@@ -0,0 +1,551 @@
...
+    # for more information.
+    #
+    Options Indexes FollowSymLinks
+
+    #
+    # AllowOverride controls what directives may be placed in .htaccess files.
+    # It can be "All", "None", or any combination of the keywords:
+    #   AllowOverride FileInfo AuthConfig Limit
+    #
+    AllowOverride All
```
simontaurus, line 286: Diffs to the default config: AllowOverride must be enabled in the target directory

simontaurus:
> This is an interesting testing approach. Thanks. We'll probably need to think this over a bit.
> ...
* I agree it would be best to use the actual config. The current file is the default apache config with two changes applied (see comments)
* have to admit I also just found out =)
* I was keeping that in mind with the search path option
* the workflow now checks only changed dirs on pull requests
* good point. I've added a precheck before the docker steps
* Maybe we can support both (need to agree on the formats of course). My idea was that it's most convenient to write the tests directly above the rules that should cover the cases. This also provides additional documentation since the tests show what we actually want to achieve.

PR #3791, opened 2023-12-18
Update of w3id.org/pmd directory structure

MattTJung:
I divided the different "sub-namespaces" into directories to make reuse of the .htaccess files a little easier (I think it also increases readability a lot). Also, I added the PMD ontology publication template which incorporates version resolving.

@joergwa could you please review this


joergwa:
Thank you!  I agree with the changes. 👍

Issue #3799, opened 2023-12-21
Remove directory under /italia


bfabio:
Hello, the directory created with #3636 should be removed, as it was not vetted by the [contacts listed in the readme.md](https://github.com/perma-id/w3id.org/blob/master/italia/readme.md) where the new directory was placed.

dgarijo:
You are right, I have given them until 27/12 to provide an alternate parent folder or I will move them to the root parent folder.

bfabio:
@dgarijo thank you :raised_hands: I appreciate your work

I'm thinking of a GitHub workflow / action to check anomalies like those, which can maybe help highlight them right away in the PR. Would that be helpful and would you be willing to include it?

dgarijo:
It's tricky since in some cases people don't add their github ids, or there are other people doing PRs on behalf of other people. In any case, up to @davidlehn to assess. From my end, it would be great to have some extra help

bfabio:
The idea is to check just `italia/` for now as we know we'll always add the maintainers' usernames in the readme files for each subdirectory, if they're supposed to be managed by a different entity.

I don't expect it to fully replace your valuable review of course, but at least it can be a layer of protection.

If that works out well, owners of other paths might be interested in using the system as well and adapting their readme. We could even come up with a machine readable microformat to specify the username.

Let's do a PoC and see how it goes.

About this directory, maybe the best thing to do here is to revert that commit: they don't need or want a top-level dir and they will provide a new fixed PR to place a different directory under `italia` anyway

bfabio:
Closing this one as it was resolved by #3805. Thanks @dgarijo

PR #3801, opened 2023-12-22
Add CI checks for italia/ hierarchical maintainership.

bfabio:
When opening a PR changing something under the italia/ identifier, check whether the author is listed as a maintainer in one of the READMEs, starting from the top directory (italia/) and and progressively narrowing down to the specific directory of the change.

Then create a comment with the findings, tagging the relevant users.

This is a PoC for https://github.com/perma-id/w3id.org/issues/3799#issuecomment-1866075273

### DIFF
--- a/.github/workflows/italia.yml
+++ b/.github/workflows/italia.yml
@@ -1,0 +1,118 @@
+name: Review PR permissions for italia/ directory
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+    paths:
+      - 'italia/**'
+
+permissions:
+  pull-requests: write
+
+jobs:
+  check-pr:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check for directory maintainership in README files
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const ROOT_DIR = "italia";
+            const pull_number = context.issue.number;
+            const { owner, repo } = context.repo;
+            const author = context.payload.pull_request.user.login.toLowerCase();
+            const authorUrlPattern = new RegExp(`https://github\\.com/${author}(?![\\w-])`, 'i');
+
+            async function getReadmeContent(directory, readmeFilename) {
+              try {
+                const { data } = await github.rest.repos.getContent({
+                  owner,
+                  repo,
+                  path: `${directory}/${readmeFilename}`,
+                });
+
+                return Buffer.from(data.content, 'base64').toString();
+              } catch (error) {
+                if (error.status === 404) {
+                  // README not found, return null to check parent directory
+                  return null;
+                }
+
+                throw error;
+              }
+            }
+
+            function extractUsernames(readmeContent) {
+              const urlPattern = /https:\/\/github\.com\/([a-zA-Z0-9-]+)/g;
+              const matches = readmeContent.matchAll(urlPattern);
+
+              return Array.from(matches, m => m[1]);
+            }
+
+            async function checkReadme(directory) {
+              const pathParts = directory.split('/').filter(part => part !== '');
+
+              let currentDir = "";
+              let usernames = null;
+              for (const part of pathParts) {
+                currentDir += `/${part}`;
+                const readmeFilenames = ['README.md', 'readme.md'];
+
+                for (const filename of readmeFilenames) {
+                  const content = await getReadmeContent(currentDir, filename);
+                  if (content !== null) {
+                    const isAuthorListed = authorUrlPattern.test(content);
+                    usernames = extractUsernames(content);
+
+                    // Deduplicate users, in case there are multiple mentions of
+                    // the same user in the README
+                    usernames = Array.from(new Set(extractUsernames(content)));
+
+                    if (isAuthorListed) {
+                      return { author_found: true, directory: currentDir, usernames };
+                    }
+                  }
+                }
+              }
+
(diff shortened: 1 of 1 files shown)


davidlehn:
Nice!  Automating maintainer checks has been on the todo list forever.  As mentioned, this will be more of a guideline than a strict rule in every case.

There's another check action here: https://github.com/perma-id/w3id.org/pull/3786

As said in that PR, I was thinking of a config file to setup this sort of metadata.  A structured data file would be far easier to parse, validate, and use than README parsing and heuristics.  Adding maintainer info would then be straightforward.  But there are concerns with duplication and syncing of info, since people seem to like to use READMEs.  Some design experimentation is likely needed.

I'll have to look at this script more.  I'm not that familiar with their API but this looks easier to write than I imagined.  

Does the `github.rest.repos.getContent` call for the README get the PR readme or the target branch readme?  Need to handle the use case of maintainers being added and removed in a good way.  At a minimum handle the case where PR author adds themselves which would need approval.  Probably good to make notifications for any changes to important data like that.

I do think we'd want this to be more generalized before being added.  But it seems a great start!

bfabio:
> Nice! Automating maintainer checks has been on the todo list forever. As mentioned, this will be more of a guideline than a strict rule in every case.
> 
> There's another check action here: #3786

@davidlehn great, I'm glad we're all on the same page about automation!

> As said in that PR, I was thinking of a config file to setup this sort of metadata. A structured data file would be far easier to parse, validate, and use than README parsing and heuristics. Adding maintainer info would then be straightforward. But there are concerns with duplication and syncing of info, since people seem to like to use READMEs. Some design experimentation is likely needed.
> 
> I'll have to look at this script more. I'm not that familiar with their API but this looks easier to write than I imagined.

The rationale for using READMEs was avoiding duplication, as you mentioned, and keep things simple. I agree that for this PoC the regexp approach is a little rough, but I was thinking about a microformat (not in the sense of microformats.org, which AFAIU are tied to HTML, and I'd rather not add an HTML parser into this) to have both something more solid to parse and keep it readable in the READMEs.

Something like, for example:
* `maintainer:https://github.com/abc123` or
* `maintainer:@abc123`
* or something equally simple

We can force it to be one per line, wdyt?

Also, these are examples of the output of this check:

* [PR from a "maintainer"](https://github.com/bfabio/w3id.org/pull/7#issuecomment-1867745805) 
* [PR from non-maintainer](https://github.com/bfabio/w3id.org/pull/8#issuecomment-1868286201)
 
It checks each directory recursively, starting from the top-level one and going down to the more nested ones, to see if the maintainer is listed in any of the READMEs.

Note that in the second PR, the `@` tags are not there so as not to notify users for this test, but they'll be included in this PR's version.

> Does the `github.rest.repos.getContent` call for the README get the PR readme or the target branch readme?

It *should* be the target branch, but good catch. This is crucial for the whole process and we should double check it does what we mean.

>Need to handle the use case of maintainers being added and removed in a good way. At a minimum handle the case where PR author adds themselves which would need approval. Probably good to make notifications for any changes to important data like that.

The action doesn't fail, but instead comments on the PR with a warning, as you can see from above, giving hints to the reviewer. In case a PR author adds themselves it would warn about them not being in one of the READMEs and tag the appropriate people, who can then review the PR.

> I do think we'd want this to be more generalized before being added. But it seems a great start!

Nice to hear! The idea is that we can maybe operate on an opt-in basis, [adding the toplevel dir](https://github.com/perma-id/w3id.org/pull/3801/files#diff-d9ac58060f3a6f15a3f7e4744d195da1ecf3d3352fcefb89312357a980ddaebcR7) of projects that want to do this check.

After that, the only place to generalize is https://github.com/perma-id/w3id.org/pull/3801/files#diff-d9ac58060f3a6f15a3f7e4744d195da1ecf3d3352fcefb89312357a980ddaebcR21

.github/workflows/italia.yml
```diff
@@ -0,0 +1,118 @@
...
+              for (const directory of Array.from(directories)) {
+                const { author_found: isAuthorFound, directory: dir, usernames } = await checkReadme(directory);
+
+                const icon = isAuthorFound ? "✅" : "⚠️";
+
+                body += `\n* ${icon} \`${directory}/\`\n`
+                body += isAuthorFound
+                  ? `Author @${author} is listed as maintainer of \`${directory}\` (via [\`${dir}/README.md\`](../blob/master/${dir}/README.md))`
+                  : [
+                    `Author @${author} is not listed in any README.md within the \`italia/\` directory`,
```
bfabio, line 105:
```suggestion
                    `Author @${author} is not listed in any README.md within the \`${ROOT_DIR}/\` directory`,
```

.github/workflows/italia.yml
```diff
@@ -0,0 +1,118 @@
+name: Review PR permissions for italia/ directory
+
+on:
+  pull_request_target:
```
bfabio, line 4: This is one of the most annoying missing features in GitHub actions: workflows from forks use a `GITHUB_TOKEN` with no permission to comment on the PR and every solution is either convoluted, a foot-gun or both.

`pull_target_request` is the simplest foot-gun and is not a security issue **as long as**:
* we don't checkout of the repository's code
* there are no secrets (https://github.com/perma-id/w3id.org/settings/secrets/actions)
* changes to this file are carefully reviewed, making sure the above items don't change over time or in the proposed patch

PR #3814, opened 2024-01-04
Corrected EMMO redirections and added tests

jesper-friis:
Checked and corrected the redirections by manually using https://htaccess.madewithlove.com/

Added a test suite intended to be run with CI from the EMMO repo. The tests will not pass before this commit has been merged to master.


davidlehn:
I'm not quite sure what to do here.  In general we've avoided checking in extra files.  There are some other repo issues discussing how to proceed with a more general testing setup.  I hesitate to start adding support files for different testing methods that won't be run here.  I know it's directly related to the redirects, but no one but the maintainers will ever use or look at those.  It might be better to hold off on that part until we figure out what test data format we'll use then those checks could be run here with common tooling.

I imagine the redirect updates themselves are fine.  Any chance you could do a PR for only those while we work on the other testing issues?

jesper-friis:
> I'm not quite sure what to do here. In general we've avoided checking in extra files. There are some other repo issues discussing how to proceed with a more general testing setup. I hesitate to start adding support files for different testing methods that won't be run here. I know it's directly related to the redirects, but no one but the maintainers will ever use or look at those. It might be better to hold off on that part until we figure out what test data format we'll use then those checks could be run here with common tooling.
> 
> I imagine the redirect updates themselves are fine. Any chance you could do a PR for only those while we work on the other testing issues?

No problem. I have now removed the redirection tests from the PR.

davidlehn:
Any chance you could rebase that file out of the history?

jesper-friis:
> Any chance you could rebase that file out of the history?

I had to refresh how to git rebase works, but I think that the test file should be removed from history now.

davidlehn:
The commits with that file add/remove are still there.  Perhaps I'm asking for too much git wizardry?  Maybe a squash of commits would be easier?  I could handle it on command line or in github, but then you'd have to resync your branch.  Thoughts?

jesper-friis:
> The commits with that file add/remove are still there. Perhaps I'm asking for too much git wizardry? Maybe a squash of commits would be easier? I could handle it on command line or in github, but then you'd have to resync your branch. Thoughts?

Please go ahead. I can resync my branch

davidlehn:
Committed in https://github.com/perma-id/w3id.org/pull/3822.  Please check and make another PR if needed.

jesper-friis:
Thanks for the fast response and doing the rebase!

PR #3815, opened 2024-01-04
Fix /vc/status-list destination.

BigBlueHat:
The /vc/status-list/2021 path had been fixed, but references from other specs (VCDM and Bitstring Status List itself) do not use the 2021 path component, but only /vc/status-list

This solves for those references.

### DIFF
--- a/vc/status-list/.htaccess
+++ b/vc/status-list/.htaccess
@@ -2,4 +2,4 @@ Header set Access-Control-Allow-Origin *
 Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
 Options +FollowSymLinks
 RewriteEngine on
-RewriteRule ^$ https://w3c-ccg.github.io/vc-status-list-2021/ [R=302,L]
+RewriteRule ^$ https://w3c.github.io/vc-bitstring-status-list/ [R=302,L]
--- a/vc/status-list/README.md
+++ b/vc/status-list/README.md
@@ -1,6 +1,6 @@
-# VC Status List Vocabulary
+# VC Bitstring Status List Vocabulary
 
-- https://w3c-ccg.github.io/vc-status-list-2021/
+- https://w3c.github.io/vc-bitstring-status-list/
 
 ## Maintainers
 


davidlehn:
@msporny @dlongley I'm not up-to-date on all the renaming, versioning/unversioning, moving around, and legacy requiresmetns of these status list resources and URLs.  The URLs here are broken, but is this the best fix?  Or should these point elsewhere?  Please advise.

https://github.com/w3c/vc-bitstring-status-list/pull/109
https://github.com/w3c/vc-data-model/pull/1406
https://github.com/w3c/vc-data-model/pull/1409

vc/status-list/.htaccess
```diff
@@ -2,4 +2,4 @@ Header set Access-Control-Allow-Origin *
 Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
 Options +FollowSymLinks
 RewriteEngine on
-RewriteRule ^$ https://w3c-ccg.github.io/vc-status-list-2021/ [R=302,L]
+RewriteRule ^$ https://w3c.github.io/vc-bitstring-status-list/ [R=302,L]
```
msporny, line 5:
```suggestion
RewriteRule ^$ https://www.w3.org/TR/vc-bitstring-status-list/ [R=302,L]
```

vc/status-list/README.md
```diff
@@ -1,6 +1,6 @@
-# VC Status List Vocabulary
+# VC Bitstring Status List Vocabulary
 
-- https://w3c-ccg.github.io/vc-status-list-2021/
+- https://w3c.github.io/vc-bitstring-status-list/
```
msporny, line 3:
```suggestion
- https://www.w3.org/ns/credentials/status/
```

vc/status-list/.htaccess
msporny: We should probably also have a redirect to the v1 context:

https://www.w3.org/ns/credentials/status/v1

Leaving that to @BigBlueHat and @davidlehn on the proper form for that these days.

msporny:
@BigBlueHat @davidlehn provided some guidance in the PR via change requests and one remaining task for one of you (how to add the `v1` context in a future-compatible rewrite rule... we have them elsewhere, I just can't remember the syntax right now).

vc/status-list/.htaccess
BigBlueHat: @davidlehn do we typically do content negotiation for that? or a dedicated URL?

dgarijo:
@BigBlueHat last ping before closing due to inactivity.

kmk142789 (approved):

PR #3842, opened 2024-01-26
Add /arp namespace for use by the ARP project

beepsoft:
This W3ID provides a persistent URI namespace for the Hungarian national repository service "Adatrepozitórium Platform" ([ARP](https://science-research-data.hu/en)).


beepsoft:
Thank you!

PR #3850, opened 2024-01-30
Moved the redirect from University GitLab to a public Github

kirnerlukas:
Hi dear folks at w3id,
I was able to finally move the Internet of Construction (IoC) Process Ontology to its new home.
Therefore changed the URLs.
Greetings,
Lukas


davidlehn:
Looks like the README needs an update too.  I'm guessing you changed from `kirner-ip` to `kirnerlukas`?

kirnerlukas:
yes, I forgot about this. Thanks for pointing it out :)

PR #3854, opened 2024-01-31
Use HTTPS for ILIAD JF property redirects

avillar:
Use `https://defs-dev.opengis.net/...` URIs instead of `http://...` ones.

@rapw3k, as the official maintainer for `/iliad`, are you ok with this change?


rob-metalinkage:
This is because there are limitations for https web clients accessing http resources after the redirect. It seems we should make all URIs https: now while we can?

rapw3k:
hi @avillar yes it makes sense of course. Where should we make changes?

davidlehn:
@rapw3k Is the proposed patch here enough?  It looks like the site handles https and http.

avillar:
@rapw3k, do we want to do the full switch to https now, not just the two last redirects to Vocprez (which is what I was proposing)? If so, I can amend the PR so that the JF properties all use https.

avillar:
@rapw3k , I've updated the PR for the JF properties to use HTTPS URIs.

dgarijo:
@rapw3k please approve and we'll merge

rapw3k:
yes, thanks @avillar

PR #3857, opened 2024-01-31
Added redirections for the EMMO-based BTO, CHAMEO, HPO and MAEO ontologies.

Leienad:


Leienad:
I have added the missing README files. These four ontologies are based on EMMO, but follow a different naming scheme with respect to the EMMO itself and are meant to be separate from the EMMO and its configuration files. Thank you.

dgarijo:
@Leienad please add the github ids of the maintainers, as I cannot map who among them you are (maybe Daniele).
We need this information in case future updates to the w3ids are needed.
Also, are all authors maintainers?

Leienad:
I have added the maintainer info. As specified, I am the only maintainer. Thank you.

PR #3878, opened 2024-02-12
Podio ontology specific version redirect

Ibaii99:
Redirection to specific version


dgarijo:
@Ibaii99 for supporting redirects to any version, have a look at https://github.com/perma-id/w3id.org/blob/master/example/.htaccess

PR #3889, opened 2024-02-15
Update redirection rules for better consistency

jesper-friis:
# Updated redirection rules for better consistency
    
For consistency with `https://w3id.org/emmo` that cant be distinguished from `https://w3id.org/emmo/`, URLs with and without trailing slash are now equivalent.
    
This is probably also what people would expect, but makes it slightly less convenient to refer to the source files in the github repo.

Also:
- added specialized rules for collecting modules in EMMO, like tlo, merocausality, disciplines, etc...
- updated comments


dgarijo:
There are conflicts in this file. Can you please address them?

jesper-friis:
> There are conflicts in this file. Can you please address them?

The conflicts are resolved now.

PR #3908, opened 2024-02-21
[AD4GD] Rules for sensors

avillar:
Add AD4GD rules for sensors -> hosted RAINBOW instance

@rapw3k, as the `ad4gd` maintainer, are you ok with this?


rapw3k (approved):
thanks @avillar

PR #3911, opened 2024-02-23
change to adoc

nicholascar:


davidlehn:
Need some updates to fix named link style.  `pandoc` probably could do it automatically.  Out of curiosity, why asciidoc?

PR #3924, opened 2024-02-28
update every subontology in ecsel-dr

bohong0528:
We've created the documentation for every lobe/subontology in previous project ecsel-dr, so we need to redirect those namespaces to their own documentation.

Thank you!

### DIFF
--- a/ecsel-dr-AH/.htaccess
+++ b/ecsel-dr-AH/.htaccess
@@ -1,0 +1,11 @@
+Options +FollowSymLinks
+# Turn off MultiViews
+Options -MultiViews
+
+# Directive to ensure *.rdf files served as appropriate content type,
+# if not present in main apache config
+
+# Rewrite engine setup
+RewriteEngine On
+
+RewriteRule ^$ https://ifx-dr.github.io/subontologies/Arrowhead/index-en.html [R=301,NE,L]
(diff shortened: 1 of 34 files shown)


davidlehn:
- The files mention `ifx-dr`, not your github id.  Who will be making updates to these?
- The directive comment in every `.htaccess` file seems extraneous and left over from copy & paste.  Same comment is in `ecsel-dr/.htaccess`.
- Is there another way this could be setup?  17 new dirs with README/.htaccess files and each for a single redirect to a related HTML file is pushing the limits of the intent and scalability of this service.  It's far more difficult for both you and us to maintain with so many files.
- It looks like `ecsel-dr` only uses the root redirect.  Why not add all these rules into that file?  Something like:

```htaccess
// ...
RewriteRule ^$ http://ontology.tib.eu/dr/ [R=301,NE,L]
RewriteRule ^AH$ https://ifx-dr.github.io/subontologies/Arrowhead/index-en.html [R=301,NE,L]
RewriteRule ^AT$ https://ifx-dr.github.io/subontologies/Automation%20Technologies/index-en.html [R=301,NE,L]
// ... same for all the others
```

Then URLs would be of the form:
- https://w3id.org/ecsel-dr
- https://w3id.org/ecsel-dr/AH

If `ecsel-dr` isn't appropriate, then maybe put all these in another top level `ecsel-dr-docs` or similar?

PR #3928, opened 2024-03-04
Redirects for SPARQL-CDT spec

hartig:
In the Amazon Neptune team at AWS we are working on a spec and corresponding artifacts (e.g., test suites) for an approach to capture and to query composite datatypes (lists and maps) as RDF literals. Our Github repo for this work is https://github.com/awslabs/SPARQL-CDTs

With this PR we want to introduce permanent identifiers for the new datatype IRIs and for the IRIs of the SPARQL functions defined in our spec, as well as for the spec document itself.

We are planning to publicly announce our proposal within the next couple of weeks. Until then, we would appreciate if pointers to it are not yet shared on mailing lists, etc.

### DIFF
--- a/awslabs/neptune/SPARQL-CDTs/.htaccess
+++ b/awslabs/neptune/SPARQL-CDTs/.htaccess
@@ -1,0 +1,34 @@
+# This space is administered by:
+#
+# Olaf Hartig
+# ohartig@amazon.com
+# GitHub username: hartig
+
+Options +FollowSymLinks
+RewriteEngine on
+
+# URIs of the datatypes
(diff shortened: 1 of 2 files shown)


hartig:
Thanks Daniel! That was fast!!

dgarijo:
no problem. I usually serve html through github pages though. It may be helpful in the future.

hartig:
Usually I would have done the same, but in this case it was not possible as we have no control over the awslabs Github pages.

dgarijo:
You can enable github pages on any repo, including https://github.com/awslabs/SPARQL-CDTs

hartig:
Ah, now it worked and I have the github pages setup for the repo. I will update the redirects and create another PR.

PR #3930, opened 2024-03-04
Fixes encoding of hash URLs in redirects for SPARQL-CDTs

hartig:
Daniel, I am terribly sorry that I have to bother you again. I just noticed that the target URLs that contain a hash character (`#`) are URL encoded, which turns them into the from URLs. The change in this PR should fix this issue (I hope!)


dgarijo:
no prob. it looks like you would save quite a few redirections by doing `#func_$1`, or similar,  instead of having all the terms in the rule file

hartig:
Good point! Next time ;-)

PR #3941, opened 2024-03-09
Rule #1a and Rule #3a have been removed since they make rule #2 and #4 ineffective.

Leienad:
Rule #1a and Rule #3a have been removed since they make rule #2 and #4 ineffective, with the consequence of making it impossible to retrieve the ttl file when using the "https://w3id.org/emmo/domain/{DOMAIN}/" url or the https://w3id.org/emmo/domain/{DOMAIN}{VERSION}/ url. For accessing the ontology's html documentation, a different mechanism needs to be established. This needs to be taken care of also for the rules concerning the whole EMMO, which suffer from the same issues.

### DIFF
--- a/CA-ODE4HERA/BiblioOnto/.htaccess
+++ b/CA-ODE4HERA/BiblioOnto/.htaccess
@@ -1,0 +1,36 @@
+# Turn off MultiViews
+Options -MultiViews
+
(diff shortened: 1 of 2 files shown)


dgarijo:
@Leienad this PR has conflicting changes, so I can't merge it. Please make sure you are working with the latest version of your repo

PR #3946, opened 2024-03-12
Add resolution for gensc/terms endpoint.

MattBlissett:
Identifiers of the form https://w3id.org/gensc/terms/MIXS:0000001 were proposed and used for the [Darwin Core extension](https://rs.gbif.org/extension/gbif/1.0/dna_derived_data_2022-02-23.xml) incorporating MIXS terms. These do not currently resolve.

This PR:

1. Replaces the `<!-- XML comment -->` with `# Apache comment`, which is the cause of the HTTP 500 errors for https://w3id.org/gensc/
2. Removes the spurious `$1` from the existing https://w3id.org/gensc/mixs 
3. Adds support for identifiers in the form https://w3id.org/gensc/terms/MIXS:0000001, redirecting them to https://genomicsstandardsconsortium.github.io/mixs/0000001/

Since I'm not sure of the status of the `gensc/terms` route I haven't added YAML/JSON/etc redirects, which exist for the [mixs route](https://github.com/perma-id/w3id.org/blob/master/mixs/.htaccess).

This is a draft PR; I am not part of the Genomics Standards Consortium.  It needs review and approval from @ramonawalls, @turbomam or @sujaypatil96.

### DIFF
--- a/gensc/.htaccess
+++ b/gensc/.htaccess
@@ -1,10 +1,12 @@
 Options +FollowSymLinks
 RewriteEngine on
 
-<!-- gensc/mixs ednpoint route -->
-RewriteRule ^mixs$ https://genomicsstandardsconsortium.github.io/mixs/$1 [R=302,L]
+# gensc/mixs endpoint route
+RewriteRule ^mixs$ https://genomicsstandardsconsortium.github.io/mixs/ [R=302,L]
 
-<!-- TODO: gensc/terms endpoint -->
+# gensc/terms endpoint route
+RewriteRule ^terms$ https://genomicsstandardsconsortium.github.io/mixs/ [R=302,L]
+RewriteRule ^terms/MIXS:(.*)$ https://genomicsstandardsconsortium.github.io/mixs/$1 [R=302,L]
 
-<!-- Rewrite Base URL -->
+# Rewrite Base URL
 RewriteRule ^(.*)$ https://gensc.org/$1 [R=302,L]


turbomam:
Thanks @MattBlissett 

I will be meeting with some of the GSC people in half an hour. I agree with your syntax fixes, but I am concerned that this may have been left in an incomplete state because we were working on some other resolver. If that is the case, I think we should help you use the preferred resolver. But I could be mistaken.

 I'll write back in about 2 hours.

turbomam:
This PR pertains to `gensc/.htaccess`

turbomam:
I believe that our maintained resolver is https://github.com/perma-id/w3id.org/blob/master/mixs/.htaccess

MattBlissett:
Thanks.

I've seen https://github.com/perma-id/w3id.org/blob/master/mixs/.htaccess, but the Darwin Core extension was produced using https://w3id.org/gensc/terms/MIXS:0000001 around 2021, predating the new resolver.  We can investigate changing to the new one for a later edition of the extension. I don't know offhand how smooth such a change is for the software.

A very similar form was used here, though it may only be a demo: https://github.com/GenomicsStandardsConsortium/gensc.github.io/blob/main/_data/terms.json#L12

turbomam:
I misspoke. The meeting's not for another 2 hours.

turbomam:
Discussed, but a decision probably won't be made for another week

turbomam:
We will probably redirect from this `gensc` namespace to the `mixs` w3id namespace as a step toward deprecating `gensc`

ramonawalls:
Yes, please wait just a bit longer. We have a technical next week where we will try to resolve this PR.

turbomam:
@sujaypatil96 and I are testing now.

The original request was illustrated with a request that  https://w3id.org/gensc/terms/MIXS:0000001 resolve. I don't believe that we have a term #1 anymore, so I would like to test with https://w3id.org/gensc/terms/MIXS:0000002

https://genomicsstandardsconsortium.github.io/mixs/0000002/ does resolve (`samp_collect_device`). That doesn't use a colon as a delimiter.

https://genomicsstandardsconsortium.github.io/mixs/0000002/ is equivalent to https://w3id.org/mixs/0000002

**I propose that we redirect from https://w3id.org/gensc/terms/MIXS:0000002 to https://w3id.org/mixs/0000002 which already has successful redirection to https://genomicsstandardsconsortium.github.io/mixs/0000002/**

turbomam:
This PR would have it redirect to https://gensc.org/gensc/terms/MIXS:0000002 which is and probably will continue to be 404. We don't intend to host term definition pages through that site any more.

turbomam:
I agree with the spirit of this PR but would like to merge this instead
- https://github.com/perma-id/w3id.org/pull/4003

dgarijo:
So shall this PR be accepted or closed?

MattBlissett:
I don't have time to review the new PR just now, but it will supersede this one.

sformel-usgs:
Just some extra context that the DwC extension now uses the current resolver per @MattBlissett 

> I think this was already done — it made sense to implement it before we enabled verbatim extension downloads in DWCA downloads on GBIF.org.  See http://rs.gbif.org/terms/1.0/DNADerivedData which has "Qualname" values like https://w3id.org/mixs/0001107

PR #3947, opened 2024-03-12
Add new 'policy' redirection inside existing catenax namespace

matgnt:
This is used for the definition of the Catena-X ODRL Profile

@HFocken can you please approve?

### DIFF
--- a/catenax/.htaccess
+++ b/catenax/.htaccess
@@ -87,5 +87,9 @@ RewriteRule ^next/taxonomy(/|\.ttl)?$ https://raw.githubusercontent.com/catenax-
 # Rewrite rule to serve complete taxonomy
 RewriteRule ^taxonomy(/|\.ttl)?$ https://raw.githubusercontent.com/catenax-ng/product-ontology/v23.12/taxonomy.ttl [R=302,L]
 
+# ODRL Profile definition
+RewriteRule ^policy/$ https://raw.githubusercontent.com/catenax-eV/cx-odrl-profile/main/profile.ttl [R=302,L]
+
+
 # Rewrite rule to serve the TTL content from the vocabulary URI by default
 RewriteRule ^(.*)$ https://catenax-ng.github.io/product-ontology [R=303,L]


HFocken (approved):

catenax/.htaccess
```diff
@@ -87,5 +87,9 @@ RewriteRule ^next/taxonomy(/|\.ttl)?$ https://raw.githubusercontent.com/catenax-
 # Rewrite rule to serve complete taxonomy
 RewriteRule ^taxonomy(/|\.ttl)?$ https://raw.githubusercontent.com/catenax-ng/product-ontology/v23.12/taxonomy.ttl [R=302,L]
 
+# ODRL Profile definition
+RewriteRule ^policy/$ https://raw.githubusercontent.com/catenax-eV/cx-odrl-profile/main/profile.ttl [R=302,L]
```
davidlehn, line 91: Did you want the similar optional trailing slash or .ttl extension like the other rules here?
```suggestion
RewriteRule ^policy(/|\.ttl)?$ https://raw.githubusercontent.com/catenax-eV/cx-odrl-profile/main/profile.ttl [R=302,L]
```

matgnt: Thanks for the hint @davidlehn ! I have refined the rule to also allow without / and everything behind the / that someone who clicks on one of our identifers like:
https://w3id.org/catenax/policy/FrameworkAgreement
is also forwarded to the ttl file.

We might further refine this in the future to have dedicated html files for those, but right now, all should end up in the ttl file.

dgarijo:
I don't think so, since the change is minimal

catenax/.htaccess
```diff
@@ -87,5 +87,9 @@ RewriteRule ^next/taxonomy(/|\.ttl)?$ https://raw.githubusercontent.com/catenax-
 # Rewrite rule to serve complete taxonomy
 RewriteRule ^taxonomy(/|\.ttl)?$ https://raw.githubusercontent.com/catenax-ng/product-ontology/v23.12/taxonomy.ttl [R=302,L]
 
+# ODRL Profile definition
+RewriteRule ^policy(/|.*)$ https://raw.githubusercontent.com/catenax-eV/cx-odrl-profile/main/profile.ttl [R=302,L]
```
davidlehn, line 91: Note that when written like this the `/` part is not needed since `.*` will catch everything.  Which also means `policyfoobar` will match.  The expected forms will match too so likely no real harm.  Fun with regexes.

matgnt: ohh, right. good point! It's not relevant right now, but with future changes I might make this clear. Thanks for the hint!

PR #3949, opened 2024-03-13
Clean Aviation ODE4HERA project (2024-2026): updated URL

ftorr29:
Updated URL of ontology and source repository


dgarijo:
@ftorr29, you'll need to approve this PR from `FrancescoTorrigiani` account, since that one is listed as the main maintainer.

FrancescoTorrigiani:
@dgarijo do I need to make a new pull request with this account? Or we can merge the current PR?

dgarijo:
No need for a new pull request, but it would help if you log in with the other account and add a +1 to this PR. The thing is that I don't know if you two are the same person. I suspect you are, but I would like to keep everything transparent. And please add both maintainers ids if  you plan changes from any of the two accounts.

FrancescoTorrigiani:
Yes, it is the same person for both account, apologies for the mess. However, I would only have [ftorr29](https://github.com/ftorr29) as maintainer, since I will dismiss this account in the future.

PR #3956, opened 2024-03-17
CA-ODE4HERA Project: updated directory structure and htaccess file

ftorr29:
- restructured directory
- updated bibliography ontology URL
- added glossary ontology URL


dgarijo:
Please note you are removing a w3id. I suggest leaving a redirect if you have used it in a publication/report

PR #3965, opened 2024-03-19
Introduce w3id.org/dgassist

stefanbischof:


dgassist/README.md
```diff
@@ -0,0 +1,5 @@
+# /sri/
```
davidlehn, line 1: Typo.

TallTed:
```suggestion
# /dgassist/
```

stefanbischof: You are right, thanks for point this out. I'll fix this in the next PR, there might be some changes next week anyway.

PR #3971, opened 2024-03-21
Default response for modules

SebSeis:
I missed the default response: html


dgarijo:
@SebSeis you are not adding content negotiation, so it's THE response, not a default response :)

PR #3976, opened 2024-03-24
MoOnt: adjust redirect

ElisEck:
Changes in my respository structure made an adjustment of the corresponding redirect necessary


dgarijo:
@ElisEck you are changing a comment, so the w3id will not be effectively changed

ElisEck:
thanks for the hint. I updated the file accordingly and fixed the redirects for the other content types as well

PR #4013, opened 2024-04-03
Add two additional rewrite rules for Biopragmatics

cthoyt:
This adds two new rewrite rules for the Biopragmatics namespace. These are specifically for artifacts in the Bioregistry project, which is a part of the Biopragmatics organization on GitHub. I'm the maintainer of both the projects and this redirect configuration on w3id


cthoyt:
thank you @dgarijo!

PR #4015, opened 2024-04-03
Add flags to rules in /np/l/

tkuhn:
Sorry for bothering you again, Daniel. No hurry for this one. And thanks a million for your terrific work in keeping the w3id service up!! :)


tkuhn:
Yes, sorry!

I promise that in the future I will test them even when they look so simple that there is no way that there could be anything wrong with them! and then there is...

:)

PR #4039, opened 2024-04-12
IFC4X3_ADD2 & Built Element Ontology (BEO)

cramonell:


dgarijo:
@cramonell you are trying to add to the `ifc` path. I see no readme file, but @pipauwel was the last one to commit to that repo. Hence I need his approval to go through with this PR.
Also, please add a readme file for the ifc w3id you are creating. I would recommend separating your w3ids if you prefer that in the meantime we approve the other one

pipauwel:
@dgarijo : Carlos @cramonell is temporarily working here with me in the office. I asked him to make these updates. So hereby you have my approval. With his input, we may be better able to keep this repository updated.

dgarijo:
@pipauwel please add a README in the root /ifc w3id stating who the maintainer for the family of w3ids is and a minimal description. We require this for all w3ids. If @cramonell is going to become a maintainer, even temporarily, then please have him add that file in another commit in this PR

PR #4050, opened 2024-04-16
Update redirect

diskettmann:
Sorry, I needed to do  minor fixes.

### DIFF
--- a/riskman/README.md
+++ b/riskman/README.md
@@ -2,7 +2,7 @@
 
 The Risk Management Ontology defines various notions related to risk management for medical devices. It builds on terms from ISO 14971.
 
-- [https://w3id.org/riskman/docs/](https://w3id.org/riskman/docs/) leads to the documentation.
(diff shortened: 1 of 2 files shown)


riskman/README.md
```diff
@@ -2,7 +2,7 @@
 
 The Risk Management Ontology defines various notions related to risk management for medical devices. It builds on terms from ISO 14971.
 
-- [https://w3id.org/riskman/docs/](https://w3id.org/riskman/docs/) leads to the documentation.
+- [https://w3id.org/riskman/docs/](https://w3id.org/riskman/) leads to the documentation.
```
davidlehn, line 5: I think you meant to change the text too?  You could also use plain URLs in the text and markdown will auto-link them.
```suggestion
- [https://w3id.org/riskman/](https://w3id.org/riskman/) leads to the documentation.
```

diskettmann: That's right, I overlooked that. Thank you very much!

PR #4051, opened 2024-04-16
permanent url for semantic-tool-hub resources

fekaputra:


toolhub/readme.md
```diff
@@ -0,0 +1,11 @@
+The Semantic Toolhub
+===
+
+The aim of this url to provide permanent location for resources related to semantic-toolhub.
```
TallTed, line 4: The w3id-based URI does not provide a location; rather it provides a redirecting address/identifier which can be re-targeted to locations that may change over time.
```suggestion
The aim of this url to provide a permanent address for resources related to semantic-toolhub.
```

fekaputra:
thank you for the suggestion - I accepted the changes

PR #4054, opened 2024-04-18
Update .htaccess for digibatt

jsimonclark:
This pull request modifies the .htaccess file for digibatt to remove the redundant "digibatt" string from the PURL, such that the following redirection is valid:

`https://w3id.org/digibatt` --> `https://digibatt.github.io/digibatt/index.html`

### DIFF
--- a/digibatt/.htaccess
+++ b/digibatt/.htaccess
@@ -1,7 +1,7 @@
 RewriteEngine On
 
 # Rule 1: point to index
-RewriteRule ^digibatt$ https://digibatt.github.io/digibatt/index.html [R=301,L]
+RewriteRule ^$ https://digibatt.github.io/digibatt/index.html [R=301,L]
 
 # Rules 2-10: point to respective pages
-RewriteRule ^digibatt/(organization|deliverable|milestone|workpackage|task|publication|data|event|software)(|/public|#(.*))$ https://digibatt.github.io/digibatt/$1.html$3 [R=301,L]
+RewriteRule ^/(organization|deliverable|milestone|workpackage|task|publication|data|event|software)(|/public|#(.*))$ https://digibatt.github.io/digibatt/$1.html$3 [R=301,L]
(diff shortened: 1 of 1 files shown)


digibatt/.htaccess
```diff
@@ -1,7 +1,7 @@
 RewriteEngine On
 
 # Rule 1: point to index
-RewriteRule ^digibatt$ https://digibatt.github.io/digibatt/index.html [R=301,L]
+RewriteRule ^$ https://digibatt.github.io/digibatt/index.html [R=301,L]
 
 # Rules 2-10: point to respective pages
-RewriteRule ^digibatt/(organization|deliverable|milestone|workpackage|task|publication|data|event|software)(|/public|#(.*))$ https://digibatt.github.io/digibatt/$1.html$3 [R=301,L]
+RewriteRule ^(organization|deliverable|milestone|workpackage|task|publication|data|event|software)(|/public|#(.*))$ https://digibatt.github.io/digibatt/$1.html$3 [R=301,L]
```
davidlehn, line 7: I don't think this is going to work like you expect.  User Agents are not going to send the hash fragment to the server.   So the `#(.*)` and `$3` part won't be used.

jsimonclark: @davidlehn , thank you very much for your review! 

You're absolutely correct that server-side redirects typically don't interact with hash fragments. In this case, the hash fragment serves as a reference to a specific anchor on a page, which the client-side browser handles to jump to human-readable documentation for the term. A similar example from a related ontology we manage can be found here: [https://w3id.org/emmo/domain/battery#battery_68ed592a_7924_45d0_a108_94d6275d57f0](https://w3id.org/emmo/domain/battery#battery_68ed592a_7924_45d0_a108_94d6275d57f0)

I tested the .htaccess file using [htaccess.madewithlove.com](https://htaccess.madewithlove.com/) and observed the desired behavior. Considering the planned usage, I believe it should work effectively. Do you agree?

dgarijo: If you tested it, I will merge

PR #4055, opened 2024-04-18
add subpath

telsaleh:
add subpath "ontology"


sedimark/ontology/README.md
```diff
@@ -0,0 +1 @@
+ontology
```
davidlehn, line 1: Was this supposed to have more content?

telsaleh: The intention is to have this new subpath to point to the ontology.
The original one will be kept as placeholder for other assets for the project. I can remove the redirect rules for this now if needed, and then added once we have a online target asset.
Please let me know if there is any concerns or restrictions with this approach.

davidlehn: - The single word "ontology" in a README doesn't add much value considering the directory has that name too.
- For this service we do want the maintainers github ids listed in .htaccess or README.  The parent /sedimark/ dir has no maintainer info.  It's only in the marketplace/ dir.  For future edits here, it becomes difficult to understand who the maintainers are.
- Could you copy & edit the marketplace/ README into this one?  Or add the similar info in the parent dir that is shared by sub directories?

telsaleh (commented):
Thanks for the feedback. I have copied over the maintainer details in the parent folder, and have added a description for the "ontology" subpath README.
Please let me know if anything else is missing,

PR #4070, opened 2024-04-22
fix: back to starting htaccess with working html redirect; serializat…

sebastianbarzaghi:
…ion still not served correctly


davidlehn:
If you need some help, please ask.

PR #4099, opened 2024-05-07
Update .htaccess

luizbonino:
Add rule for the FAIR Academics Ontology (fa-ont)


FAIR-academic/.htaccess
```diff
@@ -2,3 +2,4 @@ RewriteEngine on
 
 # FAIR4FOIS guidelines
 RewriteRule ^fair4fois/?(.*)$ https://www.utwente.nl/en/eemcs/fois2024/fair-4-fois/$1 [R=307,L]
+RewriteRule ^fa-ont#?(.*)$ https://github.com/luizbonino/fair4fois/blob/main/vocabulary/vocabulary.owl#$1 [R=307,L]
```
davidlehn, line 5: User agents likely won't send through the hash and fragment.  Also with the optional hash, this will match any trailing data, like `.../fa-ontfoobar`.

PR #4134, opened 2024-05-24
added the pghdprovo readme and htaccess files

abdullahikawu:


dgarijo:
@abdullahikawu please rename `htaccess.txt` to `.htaccess` so it works correctly

PR #4155, opened 2024-05-31
chore: upgrade to Catena-X 24.05 (and refer to intermediate 24.03)

drcgjung:
- also includes domain taxonomy rules.
- tested with a suite of url patterns against a local apache equipped with the branch.

[test.csv](https://github.com/user-attachments/files/15514276/test.csv)
[test_local.csv](https://github.com/user-attachments/files/15514278/test_local.csv)

```console

#!/bin/bash

while IFS=";" read -r url1 url2; do
    ret=$(curl -L -s -o /dev/null "$url1" -w "%{url_effective}\n")
    if [ "$ret" != "$url2" ]; then
      echo "Expected redirect from $url1 to $url2 but got $ret";
      exit 1
    fi
    retr=$(curl -s -L -o /dev/null -w "%{http_code}\n" "$ret")
    if [ "$retr" != "200" ]; then
      echo "Expected success status 200 from $ret, but got $retr";
      exit 1
    fi
done < test_local.csv

```

### DIFF
--- a/catenax/.htaccess
+++ b/catenax/.htaccess
@@ -16,6 +16,12 @@ RewriteEngine On
 RewriteBase /
 
 # Rewrite rule to serve versioned domain ontology links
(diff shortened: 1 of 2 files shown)


obalandi (approved):
LGTM

PR #4168, opened 2024-06-06
Redirect to .jsonld variant of JSON-LD context

stain:
... to ensure the Content-Type from GitHub Pages matches content negotiation

Also avoid file extension (which should not be in PIDs) for https://w3id.org/ro/terms/workflow-run/context


simleo:
> Also avoid file extension (which should not be in PIDs) for https://w3id.org/ro/terms/workflow-run/context

Is https://w3id.org/ro/terms/workflow-run/context supposed to resolve to something? I get a 404.

dgarijo:
oops, it should be (.*) instead of (.+). 
https://w3id.org/ro/terms/workflow-run/context/ works

dgarijo:
fixed

PR #4175, opened 2024-06-07
Added .htaccess file

Ibrahim-teyim:


davidlehn:
You need to create a sub directory with your redirects instead of editing the root .htaccess file.  See the instructions on the home page or look at other directories for examples: https://w3id.org/

davidlehn:
No response.  Closing.

PR #4176, opened 2024-06-09
FEP: support term definitions as ontology/schema

trwnh:
Update the .htaccess to support JSON-LD, RDF/XML, and Turtle term definitions.

### DIFF
--- a/fep/.htaccess
+++ b/fep/.htaccess
@@ -1,11 +1,30 @@
 RewriteEngine on
 
-# catch FEP-specific context requests
-RewriteCond %{HTTP_ACCEPT} application/ld\+json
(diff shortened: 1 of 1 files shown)


fep/.htaccess
```diff
@@ -1,11 +1,30 @@
 RewriteEngine on
 
+# catch FEP-specific term definitions
+
+## JSON-LD ontology/schema
+RewriteCond %{HTTP_ACCEPT} application/ld\+json
+RewriteRule ^([A-Za-z0-9]+)\/(.*?)$ https://raw.codeberg.page/fediverse/fep/fep/$1/$2/$2.jsonld [R=302,L]
```
gobengo, line 7: nice 🙏

PR #4180, opened 2024-06-11
html rewrite in .htaccess

ManueleVeggi:
Add rewrite rule for html lode documentation of the ontology

### DIFF
--- a/cnr-ispc/.htaccess
+++ b/cnr-ispc/.htaccess
@@ -3,5 +3,9 @@
 Header set Access-Control-Allow-Origin *
 Options +FollowSymLinks
 RewriteEngine on
- 
+
+# Rewrite rule to serve HTML content from the vocabulary URI if requested
+RewriteCond %{HTTP_ACCEPT} html
+RewriteRule ^ontology/branco$ https://dhilab-fi.github.io/branco/current/branco.html [R=303,L]
+
 RewriteRule ^ontology/branco$ https://dhilab-fi.github.io/branco/current/branco.owl [R=303,L]
--- a/cnr-ispc/.htaccess
+++ b/cnr-ispc/.htaccess
@@ -5,7 +5,7 @@ Options +FollowSymLinks
 RewriteEngine on
 
 # Rewrite rule to serve HTML content from the vocabulary URI if requested
-RewriteCond %{HTTP_ACCEPT} html
+RewriteCond %{HTTP_ACCEPT} text/html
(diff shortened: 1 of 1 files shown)


ManueleVeggi:
HTAccess has been modified to include rewrite rule for html

cnr-ispc/.htaccess
```diff
@@ -3,5 +3,9 @@
...
 RewriteEngine on
- 
-RewriteRule ^ontology/branco$ https://dhilab-fi.github.io/branco/current/branco.owl [R=303,L]
\ No newline at end of file
+
+# Rewrite rule to serve HTML content from the vocabulary URI if requested
+RewriteCond %{HTTP_ACCEPT} html
+RewriteRule ^ontology/branco$ https://dhilab-fi.github.io/branco/current/branco.html [R=303,L]
+
+RewriteRule ^ontology/branco$ https://dhilab-fi.github.io/branco/current/branco.owl [R=303,L]
```
dgarijo, line 11: This will be a default response, but you won't return the right type in the response. Have a look at https://github.com/perma-id/w3id.org/blob/master/example/.htaccess for a better implementation

cnr-ispc/.htaccess
```diff
@@ -3,5 +3,9 @@
 Header set Access-Control-Allow-Origin *
 Options +FollowSymLinks
 RewriteEngine on
- 
-RewriteRule ^ontology/branco$ https://dhilab-fi.github.io/branco/current/branco.owl [R=303,L]
\ No newline at end of file
+
+# Rewrite rule to serve HTML content from the vocabulary URI if requested
+RewriteCond %{HTTP_ACCEPT} html
```
dgarijo, line 8: this is not a correct type, it should be text/html:
`RewriteCond %{HTTP_ACCEPT} text/html `

dgarijo:
Please have a look at my comments

cnr-ispc/.htaccess
davidlehn: It probably should be that.  But the `CondPattern` is a regex (+magic), so matching any `html` string is ok if that's what is desired.

cnr-ispc/.htaccess
davidlehn: Are you suggesting the `AddType` setup?  Those are in *lots* of files for some reason.  But for all redirects this server is returning simple HTML and the target server needs to setup file type mappings.  (Unless I've been confused this entire time!)

dgarijo: it's also adding rdf/xml. It may be useful to return that in the type as well.

PR #4200, opened 2024-06-20
Add redirect for SCS

claudenirmf:


dgarijo:
I have merged this PR, but have you tested it out? It has an uncommon format.

davidlehn:
The `<!-- -->` comment style will likely cause the whole file to fail.

claudenirmf:
Thank you for the feedback. I used [this file](https://github.com/perma-id/w3id.org/blob/master/ontouml-models/.htaccess) as a reference, but I will upload a fix using tested RewriteRules.

PR #4208, opened 2024-06-24
Update SCS redirect rules

claudenirmf:
Update tested .htaccess file using rewrite rules

### DIFF
--- a/scs/.htaccess
+++ b/scs/.htaccess
@@ -2,19 +2,14 @@ Options -MultiViews
 RewriteEngine on
 AddType text/turtle .ttl
 
-# GENERAL RULES ---------------------------------------------------------------
+# Redirects URIs to SCS FAIR Data Point
+RewriteRule ^/scs/fdp/(.+)/?$ https://fois-fdp.eemcs.utwente.nl/$1 [R=302,L]
 
-## Organization on GitHub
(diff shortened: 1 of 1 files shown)


scs/.htaccess
```diff
@@ -2,19 +2,14 @@ Options -MultiViews
...
-### FOIS 2024 specific rule
-RedirectMatch 302 ^/scs/fdp/fois-2024/(.+)/?$ https://fois-fdp.eemcs.utwente.nl/$1
-
-### General FDP rule
-RedirectMatch 302 ^/scs/fdp/(.+)/?$ https://fois-fdp.eemcs.utwente.nl/$1
-
-<!-- ### Catalog item page -->
-<!-- RedirectMatch 302 ^/ontouml-models/?$ https://scs-ontouml.eemcs.utwente.nl/catalog/b663ca18-8085-44a7-bcfe-2c2b5ba1faa8 -->
+# Basic redirect to SCS's website
+RewriteRule ^/scs https://www.utwente.nl/en/eemcs/scs/ [R=303,L]
```
davidlehn, line 15: This probably isn't what you want for all these rules.  RedirectMatch matches on the full url path, but RewriteRules are relative to the current dir.  So you should strip the leading scs part.

PR #4228, opened 2024-06-27
Backward compatibility with PURL and additional managed and working identifiers

kaerez:


dgarijo:
Can you please provide a small explanation in the README for the w3ids?
The redirections are all except one to https://bitit.app, which redirects to google. I am a bit hesitant, to be honest.

davidlehn:
What is the purpose and use case of the purl redir?  It doesn't seem useful to use https://w3id.org/purl/purl/example to go to https://purl.archive.org/purl/example which will go to whatever target is setup.  Why not use their service directly?

I have same concerns as @dgarijo too.

dgarijo:
Ping @kaerez

dgarijo:
No reaction for over a year, closing PR. Please reopen if appropriate

PR #4242, opened 2024-07-01
Add redirect for citizenship/uscis/v1.

wes-smith:


dgarijo:
@wes-smith you are not listed among the maintainers of this repo.
Please have: 
- @davidlehn 
- @dlongley
- @mattcollier
- @msporny
Approve this PR. Also, having a readme would help!

davidlehn:
Noting that this and the previous added URL here is not an ideal pattern.  It should be scoped differently with a top level directory for a specific project.  In these cases, the URLs got into code and data that is difficult to change at this time and people are aware of the issue.  We'll just deal with the issue for now.

davidlehn:
@dgarijo I find .htaccess comments just as easy to use for short docs and looking up user ids.

PR #4254, opened 2024-07-07
add redirection for mlao ontology

friendlynihilist:
Creating a new identifier for Multi-Level Annotation Ontology (MLAO) and add redirection for MLAO ontology.

### DIFF
--- a/mlao/.htaccess
+++ b/mlao/.htaccess
@@ -1,0 +1,4 @@
+Options +FollowSymLinks
+RewriteEngine On
+RewriteRule ^$ https://friendlynihilist.github.io/mlao/ [R=302,L]
+RewriteRule (.*) https://friendlynihilist.github.io/mlao/$1 [R=302,L]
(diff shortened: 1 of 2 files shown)


friendlynihilist:
> Worth noting, "MLAO ontology" is redundant, given that the expansion of the acronym is "Multi-Level Annotation Ontology ontology". That is likely to lead to various trouble down the road. I suggest avoiding saying "MLAO ontology" as much as possible, and rather saying either "MLA ontology", where `MLA` has [~80 common expansions](https://www.acronymfinder.com/MLA.html), or just "[MLAO](https://www.acronymfinder.com/MLAO.html)", which currently has [1 relatively common expansion](https://www.acronymfinder.com/MLAO.html) and [1 archaic](https://www.acronymattic.com/Maritime-Legal-Affairs-Office-(MLAO).html). I also recommend adding your meaning of [MLAO](https://w1.acronymfinder.com/~/app/add.aspx?acronym=MLAO) and/or [MLA](https://w1.acronymfinder.com/~/app/add.aspx?acronym=MLA) to their list.

Absolutely, thank you for the suggestion!

PR #4276, opened 2024-07-15
update htaccess, testing content negotiation

sreyfe:


iberz/.htaccess
```diff
@@ -1,6 +1,28 @@
-AddType text/turtle .ttl
 
+# Do not remove this line, otherwise mod_rewrite rules will stop working
+RewriteBase /
```
davidlehn, line 3: Why would the rules stop working?  Rewrite base is used to reference local files and only remote redirects are used here.

iberz/.htaccess
```diff
@@ -1,6 +1,28 @@
...
+
+# Rewrite engine setup<br/>
```
davidlehn, line 5:
```suggestion
# Rewrite engine setup
```

iberz/.htaccess
```diff
@@ -1,6 +1,28 @@
...
 
-RewriteRule ^(.*)$ https://iberz.org/$1 [R=302,L]
+
+AddType application/rdf+xml .owl
+
+AddType application/rdf+xml .rdf .owl
+
+AddType text/turtle .ttl
+
+# Rewrite rule to serve HTML content<br/>
```
davidlehn, line 15:
```suggestion
# Rewrite rule to serve HTML content
```

iberz/.htaccess
sreyfe: just copied the code, I can delete the comment

iberz/.htaccess
```diff
@@ -1,6 +1,28 @@
-AddType text/turtle .ttl
 
+# Do not remove this line, otherwise mod_rewrite rules will stop working
+RewriteBase /
+
```
davidlehn, line 4: Remove unneeded RewriteBase.
```suggestion
```

iberz/.htaccess
davidlehn: Ok.  Added a suggestion. Feel free to test it out and make sure it does what you want.

sreyfe: looks good!

PR #4280, opened 2024-07-17
Adding Persistent URIs related to the Cultural Thematic Path (CTP) Ontology

riccardoAlbertoni:
Please consider merging this contribution.
It adds persistent URIs related to the Cultural Thematic Path (CTP) Ontology.

Regards, 
Riccardo


ctp/README.MD
```diff
@@ -0,0 +1,21 @@
+Persistent URIs related to the Cultural Thematic Path (CTP) Ontology, an ontology for the creation and structuring of cultural thematic paths, aimed at supporting the linked data publishing and managing for the fruition of cultural heritage.
+
+
+The project is developed under the Ph.D. Research Project in Digital Humanities, supported by PON "Ricerca e Innovazione" 2014-2020, Asse IV "Istruzione e ricerca per il recupero", Azione IV.5 "Dottorati su tematiche Green" DM 1061/2021, and is driven by Italian National Research Council - [Istituto di matematica applicata e tecnologie informatiche "Enrico Magenes" (IMATI)](https://www.cnr.it/en/istituto/050/)
+
+
+===================
+
+We intend to use **https://w3id.org/ctp** for  persistent URIs in referring to the ontology and its complementary documentation
```
davidlehn, line 9: It would be better to phrase these paragraphs as a description of what this is and what's here.  It looks like it's describing a request and future plans now.  For instance "We intend to use foo for" can be "foo is used for".  And "We initially plan to create" can be "The following are available" or whatever makes sense.

riccardoAlbertoni: @davidlehn Thanks for the comment.
I have considered it in my new commit, also I have provided further details on the URIs we are minting.

dgarijo:
Instead of https://htmlpreview.github.io/?https://raw.githubusercontent.com/TizianaPasciuto/CTP_Ontology/master/documentation/index-en.html, you may consider using GitHub pages

PR #4323, opened 2024-08-02
fix: temporarily migrate catenax ontologies 

drcgjung:
The Catena-X consortium ended at 2024-07-31
We are currently migrating the ontologies to the Catena-X association and its Open Source Organisation (Eclipse Tractus-X)
Until that migration is over, we temporarily introduce an intermediate location (at big-data-spaces orga) to avoid downtimes in any referring systems.

### DIFF
--- a/catenax/.htaccess
+++ b/catenax/.htaccess
@@ -16,106 +16,106 @@ RewriteEngine On
 RewriteBase /
 
 # Rewrite rule to serve versioned domain ontology links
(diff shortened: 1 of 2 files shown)


obalandi:
This is a good interim solution until the ontology repository is moved to Eclipse Tractus-X (https://github.com/eclipse-tractusx/sldt-ontology-model/pull/2). Thx!

PR #4331, opened 2024-08-05
Update GH repo name for all redirections in .htaccess

turbomam:
repo name changed

### DIFF
--- a/aio/.htaccess
+++ b/aio/.htaccess
@@ -4,4 +4,4 @@ RewriteEngine on
 # Rewrite Base URL
 # will add $1 suffix later
 # static redirect for now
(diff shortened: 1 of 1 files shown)


davidlehn:
Did you want to change the README links too?  GitHub will auto-redirect, but if the name changed, might as well match it.  

If doing that, it would help to add github ids for maintainers so it's easier to match who's making PRs.  Thanks.

cmungall:
Hi @turbomam I think this PR can be closed

- #4356 - I updated the README in response to @davidlehn's comments
- #4355 - This PR incorporates your change but also sets up paths for individual terms and downloads (which we should have done originally...)

PR #4340, opened 2024-08-11
Add ontogen vocabularies

marcelotto:


dgarijo:
You are allowing for w3id.org/ontogen/ontogen, not sure if intended

PR #4344, opened 2024-08-13
updates for new iSamples vocabulary URI paths

smrgeoinfo:


davidlehn:
Merging.  A bit complex to review.  Are some of those parent rules matching what looks like versions overriding the child dirs?

PR #4380, opened 2024-08-29
Creation of perma-id for the Thermodynamics of Alloys Ontology

kgubaev:
Link to github:
https://github.com/kgubaev/thermodynamic-alloy-ontology

Please consider my request.

Best regards,
Konstantin


davidlehn:
@kgubaev: There's a problem with this addition.  See https://github.com/perma-id/w3id.org/issues/4417.  There was an existing `tao` directory.  Case insensitive file systems, like on MacOS, have problems handling the git checkouts in situations like this and it causes confusion and PR problems.  The serving system itself has no problems with this, but to avoid the issues we have a policy of avoiding files and directories that only differ in case.  This one slipped through.  Is there an alternate name to `TAO` you could use?  I know that's annoying, but one of these needs to change.

dgarijo:
@kgubaev ping again. This is causing problems (see https://github.com/perma-id/w3id.org/pull/4424). 
What about ThAO?

davidlehn:
Changed in https://github.com/perma-id/w3id.org/pull/4427.

PR #4385, opened 2024-08-31
idn rd/ -> resources/

nicholascar:


davidlehn:
The commit and PR messages are confusing here.  I assume the content is what you intended?

PR #4414, opened 2024-09-17
Update .htaccess

ifx-dr:
Update link from "http://ontology.tib.eu/dr/" to "https://ifx-dr.github.io/DigitalReference/"

### DIFF
--- a/ecsel-dr/.htaccess
+++ b/ecsel-dr/.htaccess
@@ -8,7 +8,7 @@ Options -MultiViews
 # Rewrite engine setup
 RewriteEngine On
 
-RewriteRule ^$ http://ontology.tib.eu/dr/ [R=301,NE,L]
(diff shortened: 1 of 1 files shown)


ifx-dr:
Hello good afternoon, 
with this pull request I wanted to change the page that redirects to when you enter the link: "[w3id.org/ecsel-dr](https://w3id.org/ecsel-dr)". 
But after making the change and seeing that the merge has been applied, the page still redirects to the old link: "http://ontology.tib.eu/dr/" and it should be the new link: "https://ifx-dr.github.io/DigitalReference/". 
Which is the change I proposed in the commit.

Any idea why it keeps redirecting to the old link?

davidlehn:
Maybe a caching issue? Looks like it's working:
```
$ curl -sI https://w3id.org/ecsel-dr/ | grep Location
Location: https://ifx-dr.github.io/DigitalReference/
```

PR #4423, opened 2024-09-22
Adding Green's Dictionary of Slang and Gay/Lesbian/Bisexual Television Characters.

Superraptor:
Adding redirects from two sources that utilized hashes (`#`) in their identifiers. The reason for doing this is so that the external identifiers datatype property utilized in instances of Wikibase can access these identifiers. These redirects have been tested using the online tool (https://htaccess.madewithlove.com/).

### DIFF
--- a/GayLesbianBisexualTVCharacters/.htaccess
+++ b/GayLesbianBisexualTVCharacters/.htaccess
@@ -1,0 +1,16 @@
+# Example
+#
+# https://w3id.org/GayLesbianBisexualTVCharacters/1980s/insi redirects to https://home.cc.umanitoba.ca/~wyatt/tv-char1980s.html#insi
+#
+# ## Contact
+# This space is administered by:
+#
+# Clair Kronk
+# Clair.Kronk@mountsinai.org
+# GitHub username: Superraptor
+
+RewriteEngine On
+RewriteBase /
+
+# Redirect URLs of the form /GayLesbianBisexualTVCharacters/$1/$2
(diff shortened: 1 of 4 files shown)


dgarijo:
@Superraptor it would be great if you also added your contact name in the readme file (apart from htaccess). Thanks in advance!

GayLesbianBisexualTVCharacters/.htaccess
```diff
@@ -0,0 +1,25 @@
...
+# Clair Kronk
+# Clair.Kronk@mountsinai.org
+# GitHub username: Superraptor
+
+# Examples tested using:
+# https://htaccess.madewithlove.com/
+# (22 September 2024)
+
+RewriteEngine On
+RewriteBase /
```
davidlehn, line 19: This is used for local path redirections.  Not needed for remote URLs.  Same in other .htaccess.

GayLesbianBisexualTVCharacters/.htaccess
```diff
@@ -0,0 +1,25 @@
...
+
+# Redirect URLs of the form /GayLesbianBisexualTVCharacters/$1
+RewriteRule ^GayLesbianBisexualTVCharacters/([a-z]{1,10})$ https://home.cc.umanitoba.ca/~wyatt/tv-characters.html#$1 [R=301,NE,L]
```
davidlehn, line 22: RewriteRule patterns are relative to the current directory.  Remove the prefix in all the rules for both these .htaccess files.

Superraptor: Sincere apologies-- fixed here: https://github.com/perma-id/w3id.org/pull/4428.

PR #4433, opened 2024-09-26
Switch Inrupt administrator to be David Bowen

davidboweninrupt:
Pat is no longer administering the persistent resources.


dgarijo:
Approving since I see the thumbs up from Nicolas A. Seydoux

davidboweninrupt:
Thanks for merging this so quickly @dgarijo

PR #4437, opened 2024-09-27
Create .htaccess

essepuntato:


essepuntato:
HI @dgarijo, it seems that the redirect is not working so far. Is that normal? Do we need to wait a bit longer?

Have a nice day :-)

dgarijo:
@essepuntato it's usually immediate. Did you test the file locally?

essepuntato:
@dgarijo now it works!

PR #4453, opened 2024-10-03
Create a permanent identifier for the dob ontology and instance data

garyedwards:
Dear maintainer,

Could you please create a permanent identifier for the dob ontology?
I hope I provided you with the correct documents via this way.

Project: Open Built Environment Data
Prefix: dob
Namespace: https://w3id.org/dob

Best, Gary
gary@xri.online

### DIFF
--- a/dob/.htaccess
+++ b/dob/.htaccess
@@ -1,0 +1,17 @@
+# # DOB: Open Built Environment Data
+#
+# DOB is an open data initiative to share built environment data in a structured way.
+#
+# https://w3id.org/dob/ redirects to https://github.com/abc-rp/dob/
+#
+# ## Contact
+# This space is administered by:  
+#
+# Gary Edwards
+# GitHub username: garyedwards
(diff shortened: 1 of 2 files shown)


davidlehn:
- Did you want voc and id paths to resolve to something or are they plain opaque ids?
- The project page says "W3C’s w3id.org".  w3id.org is not directly affiliated with the W3C.

garyedwards:
Hi @davidlehn 

Thanks for your comments.  I have removed the reference to w3c on the project page.

At the moment we are going to publish the ontologies and example data as ttl files in the project repo for comment.  I have added some place holder ttl files for now.  These will be populated shortly.

When we the project is more stable we will have a more complete hosting solution with http + ttl representations backed off a triple store.  At this point we will update the .htaccess file accordingly.  We have no intention of having opaque ids in the longer term.

Please let me know if you have any other comments.

Gary

PR #4455, opened 2024-10-03
2 redirect rules for pointing /data/$id and /assessments/$id to the fair-checker web site 

albangaignard:
Many thanks for merging !


albangaignard:
I tested the .htaccess locally and everything should be ok.

PR #4461, opened 2024-10-09
feat: enable redirection for Compliance Document criteria

egavard:
Enables a redirect from specs/criteria/cd24.06/criteria/XXX to either the JSON criteria or its HTML documentation depending on the Accept header.


egavard:
Thanks for the quick merge. Very appreciated !

PR #4465, opened 2024-10-10
Permanent URIs for the Horizon project SmartEdge

jpcik:
URIs for the ontologies of SmartEdge, including swarm ontology, and recipe vocabulary

https://www.smart-edge.eu/


jpcik:
Ok added

PR #4478, opened 2024-10-15
virtual reality ontology

italojsoliveira:
I am creating this for research work with industrial implications.


dgarijo:
have you tried the redirect? I think it may not be correct

italojsoliveira:
It seems to be working. Is there any problem yet?

PR #4479, opened 2024-10-17
updated iiot entry redirection

gatemezing:


iiot/readme.md
```diff
@@ -0,0 +1,8 @@
+https://w3id.org/IIoT is a permanent URI 
+
+for the IIoT ontology for Industrial Internet of Things: 
+
+https://ontoit.github.io/IIoT/IIoT.owl
+
+Maintainer: Liviana Tudor
```
TallTed, line 7: @dgarijo — You did not request inclusion of Liviana Tudor's email address nor github ID?

davidlehn:
I'm not sure what's going on here, but I'm going to revert this now.  These `iiot` contents are a duplicate of `IIoT`.  And we have a policy to avoid case similar ids due to confusion in some file systems.

iiot/readme.md
gatemezing: tudor.liviana@gmail.com - @OntoIT

dgarijo:
@gatemezing look at the folder names you are uploading. You have to keep consistent with the case sensitive systems.

gatemezing:
Probably I’m using a different OS.
My understanding was to have only one “IIoT” with the new redirections.
Is it something possible ?



"*Love all, trust a few, do wrong to none*" (W. Shakespeare)
Web: http://atemezing.org



Le ven. 18 oct. 2024 à 08:56, Daniel Garijo ***@***.***> a
écrit :

dgarijo:
This PR adds a new namespace `iiot`. It does not rename anything.

dgarijo:
In order to simplify things, would it be ok if we only kept `iiot` (non-capital letters)?

gatemezing:
> In order to simplify things, would it be ok if we only kept `iiot` (non-capital letters)?
This is more for you @OntoIT - could you answer?

PR #4480, opened 2024-10-17
Turn several redirects case-insensitive

Muffexx:
Hello everyone,

in this commit we have made our sub-redirects case-**in**sensitive.

However, we have also made the main-redirects for CEIR, ColActDOnt and SocDOnt ( which are all **our** projects) case-**in**sensitive. Unfortunately, this is necessary because **lower-case** links are expected in some publications and misunderstandings have already occurred.

Kind regards

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -5,7 +5,5 @@ RewriteEngine on
 
 # special directory redirects due to case insensitive file systems issues
 RewriteRule ^AIRO(.*)$ /airo$1
-RewriteRule ^CEIR(.*)$ /CEIR$1 [NC]
-RewriteRule ^ColActDOnt(.*)$ /ColActDOnt$1 [NC]
-RewriteRule ^SocDOnt(.*)$ /SocDOnt$1 [NC]
+RewriteRule ^ceir(.*)$ /CEIR$1
 RewriteRule ^UN(.*)$ /un$1
--- a/.htaccess
+++ b/.htaccess
@@ -5,4 +5,7 @@ RewriteEngine on
 
 # special directory redirects due to case insensitive file systems issues
 RewriteRule ^AIRO(.*)$ /airo$1
+RewriteRule ^CEIR(.*)$ /CEIR$1 [NC]
+RewriteRule ^ColActDOnt(.*)$ /ColActDOnt$1 [NC]
+RewriteRule ^SocDOnt(.*)$ /SocDOnt$1 [NC]
 RewriteRule ^UN(.*)$ /un$1
--- a/CEIR/.htaccess
+++ b/CEIR/.htaccess
@@ -11,9 +11,9 @@ Options -MultiViews
 RewriteEngine On
 
 # Redirects
-RewriteRule ^(.*)CWC$ https://github.com/ceir-koblenz/CWC [R=303,L]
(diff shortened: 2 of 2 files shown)


davidlehn:
- We have tried to avoid these case issues.  As you see, two slipped through as special cases.
- I'm not sure what to do here.  We *really* don't want everyone to request these special cases.
- At a minimum, the root redirects should be specific strings, not `[NC]` to allow ArbITrAry nonsense to get through.
- Are all those ids having this issue or is it only one?
- Are publications really expecting lower case paths?  That sounds absurd since URLs don't, and never have, worked like that.
- I think handling multiple cases for paths is more of a last resort due to errors that made it to production and are difficult or impossible to fix.  In general, URLs should use a specific spelling and stick with that.  Otherwise you can have numerous problems with equivalency and comparison at least.

dgarijo:
We can state that from now on, only non-capital letters are accepted as part of w3ids. Old redirects are kept.
I agree that adding exceptions for every single thing with capital letters does not make a lot of sense.

Muffexx:
Hello again.

First of all, thank you very much. I'm a big fan of w3id's permanent identifiers and I think it's great that this service is provided free of charge.

Issues regarding case-sensitivity will probably occur more often. In my opinion, it is not particularly intuitive that URLs have to be case-sensitive. Of course, the acronyms for our ontologies (SocDOnt & ColActDOnt) make it particularly easy to inject a typo.

Why aren't all redirects set to `[NC]` at the top level? Should it really be possible to have `Abc`, `ABc`, `ABC`, `aBc`, `aBC ` and so on as separate redirects? I suspect that the desired resources have not been reached several times. What would be the disadvantage of “ArbITrAry nonsense” if you still reach the resource? (This links does work fine: [HTtpS://gITHub.coM/perMA-iD/w3Id.oRg](HTtpS://gITHub.coM/perMA-iD/w3Id.oRg))

Having only one format/redirect certainly facilitates the processing and comparison of URLs. However, in view of the fact that w3id is a project that is intended to help make content accessible for years to come, I see it as an advantage if a upper/lower case character that got lost in translation (or elsewhere) does not cause any problems.

Currently there is only an acute difficulty for our “**CEIR**” redirect, which was referenced as “ceir”. The corresponding exception is _really important_ to me here.

How about allowing the _native spelling_ and (only) _lowercase_? This would limit the alternatives per project to a maximum of two.

These are just my (naive) ideas. After all, I can't make any decisions for w3id. Thank you very much for the feedback so far!

Regards

TallTed:
@dgarijo — I strongly advise that propositions like "We can state that from now on, only non-capital letters are accepted as part of w3ids." be taken up in their own issues, rather as part of a thread attached to a specific w3id pull request, where people who are quite interested in the overall w3id.org project have no expected reason to pay attention.

Muffexx:
Hello everyone

On the assumption that the topic is not a particularly high priority, I have updated the PR. Now there are only changes within the subfolders and **one exception** (ceir->CEIR).

Possible fundamental changes can be decided in due course at a more appropriate place.

Kind regards

dgarijo:
Thanks @Muffexx.
@davidlehn in order to avoid having this PR blocked, can we proceed with the current change, and discuss how to proceed in general in another issue?

davidlehn:
Merged.  Not too thrilled about adding special cases though.

PR #4482, opened 2024-10-17
Revert "updated iiot entry redirection".

davidlehn:
Reverts perma-id/w3id.org#4479.

Reverting due to maintainer confusion and the `iiot` contents added here duplicate `IIoT` and we want to avoid case similar directories.


dgarijo:
apologies, I did not check for duplicates

PR #4502, opened 2024-10-28
Security: Forbid browsing of .git directory

jkbzh:
Hi, although we don't manage your site, w3.org got a report related to your site for  a potential security issue. Here's the fix for it.
> 
> Hi team
> 
> Description
> .git repository exposed at  https://w3id.org/.git/
> 
> Step
> 1. In your terminal type   wget -mirror -I .git https://w3id.org/.git/
> 2. go to the downloaded file directory
> 3. git status
> 4. Deleted files will appear
> 5. git restore (any file)
> 6. check git log also
> 
> Impact
> Attacker can restore deleted files and any of file consist of db user and
> db password
> 
> Best regards
> Manjot Singh

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -3,6 +3,13 @@
 Options +FollowSymLinks
 RewriteEngine on
 
+# If you use version control systems in your document root, you should
+# probably deny access to their directories. 
+#
+<DirectoryMatch "/.git">
+   Require all denied
+</DirectoryMatch
+
 # special directory redirects due to case insensitive file systems issues
 RewriteRule ^AIRO(.*)$ /airo$1
 RewriteRule ^ceir(.*)$ /CEIR$1


davidlehn:
- Thanks.  I don't think it was any security issue, since all the git data is available here on github anyway.  But it's cleaner to not allow that.
- I put a block in the server config for now.  When we eventually get https://github.com/perma-id/w3id.org/pull/3264 done, the .git dir would be outside of the web root dir making this patch not needed.
- Leaving this open for a bit.  But probably will close unless someone thinks block should be in the root .htaccess too.

jkbzh:
Hi Agreed it's not a security issue. 

You could also just add that restriction directly in the server's configuration file. The vanilla apache on debian proposes such a rule in conf-available/security.conf. You just need to remove the comment. Maybe this would be a big compromise to keep you from modifying your .htaccess file.

You may close this MR when you want.

KUDOS

davidlehn:
I did add the restriction to the site config when this was created.  That security line is shorter though.  Either way, it's blocked now.  Closing.

PR #4512, opened 2024-11-04
Fixed schema rewrite rule

mntlra:


dgarijo:
Please try all your redirections locally before doing a PR.

PR #4513, opened 2024-11-04
Add extra sub-domain for PharmAccess' MomCARE FHIR distribution

hbcesar:


davidlehn:
The existing wildcard will do the same redirection for sub-paths.  Was the target supposed to be different?  "Add extra sub-domain" in the PR title is why I'm asking.

hbcesar:
We're using the same target temporarily during development. We'll update it later.

PR #4518, opened 2024-11-07
Fix redirections to the SSSOM documentation.

gouttegd:
The documentation for the SSSOM specification was recently overhauled and several parts of it have moved, so some redirections currently point to 404 errors.

This commit fixes those redirections to get them to point to the appropriate new documents.

CC @matentzn


matentzn (approved):
Thank you @gouttegd

PR #4521, opened 2024-11-10
Update NSDRA rules

ammar257ammar:


dgarijo:
I think you can generalize the rule, but I will merge

PR #4538, opened 2024-11-17
UniOnto Ontology Creation

gassis:


davidlehn:
- The file name has a typo, it needs to be `.htaccess`.
- Please add a list of github ids of the maintainers of this directory.
- This file has all the rules commented out, and the final fallback is trying to use a local error file (which we don't support with this service).  Please enable and test the rules you want to use before we merge this.

UniOnto/.htaccess
```diff
@@ -0,0 +1,3 @@
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^$ https://https://gassis.github.io/UniOnto/ [R=302,L]
```
davidlehn, line 3: typo

PR #4542, opened 2024-11-19
LCA ontology

JulKaltenegger:


bmp/.htaccess
```diff
@@ -37,7 +37,7 @@ RewriteRule ^$ https://julkaltenegger.github.io/bmp/ontology.nt [R=303,L]
 RewriteCond %{HTTP_ACCEPT} text/turtle [OR]
 RewriteCond %{HTTP_ACCEPT} text/\* [OR]
 RewriteCond %{HTTP_ACCEPT} \*/turtle 
-RewriteRule ^$ https://julkaltenegger.github.io/bmp/ontology.ttl [R=303,L]
+RewriteRule ^$ https://julkaltenegger.github.io/bmp/ontology.html [R=303,L]
```
davidlehn, line 40: Looks like this got in by mistake?

JulKaltenegger:
Yes. I wanted to add the lca ontology only.
Working from two laptops (transition), hence the overlap.
Did you receive the lca pull request?

dgarijo:
Yes, but you are also modifying the other file (in `bmp`). Please amend that edit if it was not intended

JulKaltenegger:
i think it should be fine now?

PR #4552, opened 2024-11-25
KG for the Murten Panorama Digital Scholarly Edition

chauraph:
This is the initial commit of the project's PURL prefix.


dgarijo:
@chauraph can you please add your github id in the readme so we know it's you?

PR #4560, opened 2024-11-30
Update .htaccess

DelaramGlp:


aiup/.htaccess
```diff
@@ -1,3 +1,19 @@
...
 RewriteEngine on
+
+RewriteCond %{HTTP_ACCEPT} text/turtle
+RewriteRule ^$ https://delaramglp.github.io/aiup/aiup.ttl [R=302,L]
+
+# https://w3id.org/aiup/1.0
 RewriteRule ^$ https://delaramglp.github.io/aiup [R=302,L]
+
+# Default
+RewriteRule ^$ https://delaramglp.github.io/aiup [R=302,L]
```
davidlehn, line 18: This is the same as the above rule.

PR #4575, opened 2024-12-09
feature: setup up to Catena-X 2412

drcgjung:
- documents the catena-x namespace releases
- links to the new markdown documentation base (without github pages)
- introduces rewrite rules for intermediate versions and links current and next version to the right tags.
- tested on a local installation of w3id.org


drcgjung:
would be nice if @obalandi could confirm the changes.

obalandi (approved):
I confirm the changes!

PR #4581, opened 2024-12-11
adding ancmed for ancmed.ulb.be

CodingCerberus:


ancmed/.htaccess
```diff
@@ -0,0 +1,8 @@
+RewriteEngine on
+
+# Urls ending in numbers redirect to the relevant php page for the entry
+RewriteRule ^(.*)\/([0-9]+)$ https://ancmed.ulb.be/details.php?row=$2 [R=302,L]
+RewriteRule ^(.*)\/([0-9]+)/$ https://ancmed.ulb.be/details.php?row=$2 [R=302,L]
```
davidlehn, line 5: FYI, You can simplify to one line with an optional trailing slash:
```diff
-RewriteRule ^(.*)\/([0-9]+)$ https://ancmed.ulb.be/details.php?row=$2 [R=302,L]
-RewriteRule ^(.*)\/([0-9]+)/$ https://ancmed.ulb.be/details.php?row=$2 [R=302,L]
+RewriteRule ^(.*)\/([0-9]+)/?$ https://ancmed.ulb.be/details.php?row=$2 [R=302,L]
```

CodingCerberus: Thanks David! I'm not the best with regex but that makes a lot of sense!

Could you give me some advice on why the redirect isn't working as intended? I used a htaccess checker before committing, but w3id.org/ancmed/1 doesn't redirect to https://ancmed.ulb.be/details.php?row=1, just to ancmed.ulb.be

I don't know what's wrong

PR #4616, opened 2024-12-20
Add dacbarbos w3id

orgadm:
Create new W3ID/workspace for DACBARBOS Brand.

### DIFF
--- a/dacbarbos/.htaccess
+++ b/dacbarbos/.htaccess
@@ -2,4 +2,4 @@
 
 Options +FollowSymLinks
 RewriteEngine on
-RewriteRule ^(.+)$ https://github.com/$1 [R=302,L]
+RewriteRule ^(.*)$ https://github.com/dacbarbos/$1 [R=302,L]
--- a/dacbarbos/.htaccess
+++ b/dacbarbos/.htaccess
@@ -1,0 +1,5 @@
+# w3id .htaccess rules for dacbarbos
+
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^(.+)$ https://github.com/$1 [R=302,L]
--- a/dacbarbos/README.md
+++ b/dacbarbos/README.md
@@ -1,0 +1,7 @@
+# W3ID for /dacbarbos/
+
+## Workspace: DACBARBOS Brand's permanent ID
+
+## Contacts: DACBARBOS Brand's Organization Administrator (admin@dacbarbos.eu)
+
+### GitHub username: @orgadm


dacbarbos/.htaccess
```diff
@@ -0,0 +1,5 @@
+# w3id .htaccess rules for dacbarbos
+
+Options +FollowSymLinks
+RewriteEngine on
+RewriteRule ^(.+)$ https://github.com/$1 [R=302,L]
```
davidlehn, line 5: Did you mean to have a github path in there like `https://github.com/dacbarbox/$1`?  I'm hesitant to enable redirect to the entirety of github.

orgcontrib: That's my GH org, thus w3id/org/repo would go to gh/org/repo. Better answered by Gemini [here](https://bit.ly/3VNVLdu).

orgcontrib: I'd like to start with something casual until I think to a better use of it.

orgadm: @orgcontrib is my regular org-user profile while @orgadm is the manager org-profile

davidlehn: The concern here is that your rule would allow a redirect to anything on github.  For instance you could redirect to this PR with `https://w3id.org/dacbarbos/perma-id/w3id.org/pull/4616`.  That's a bit too "casual" and could be open for intentional or unintentional abuse of this service.  What is your use case?  What resources are you trying to redirect to specifically?

orgadm: As I understand it, that would go to `https://github.com/dacbarbos/perma-id/w3id.org/pull/4616` and thus will return a 404 page. I'm not sure I get your concern.

davidlehn: The matched [patterns](https://httpd.apache.org/docs/2.4/mod/mod_rewrite.html#rewriterule) are relative to the current directory.

So this sequence is roughly:
- Request URL of `https://w3id.org/dacbarbos/example/path`.
- Check rules in `/dacbarbos`.
- Matches `^(.+)$` pattern and sets `$1` to `example/path`.  (Does not have the `dacbarbos` prefix!)
- Substitute match to result in `https://github.com/example/path`.
- Return the redirect.

orgadm: I tried this in [htaccess tester](https://htaccess.madewithlove.com/).
Please refer to the attached pdf.
[htaccess tester - dacbarbos.pdf](https://github.com/user-attachments/files/18216163/htaccess.tester.-.dacbarbos.pdf)
Is this tool unreliable/wrong?

davidlehn: That tool tests as if the `.htaccess` is in the root dir.  It's not wrong, but it's not testing what you are trying to test.

orgadm: Oh, I see. What about this proposed A/B refactoring?

```
# A
Options +SymLinksIfOwnerMatch
RewriteBase /dacbarbos/
RewriteCond %{REQUEST_URI} ^/dacbarbos/(.*)$ [NC]
RewriteRule ^dacbarbos(.*) https://github.com/dacbarbos$1
# B
Options +SymLinksIfOwnerMatch
RewriteBase /dacbarbos/
RewriteCond %{REQUEST_URI} ^/dacbarbos/(.*)$ [NC]
RewriteRule ^dacbarbos/(.*)$ https://github.com/dacbarbos/$1 [R=302,L]
```

davidlehn: - `RewriteBase` is for local path redirects.
- Patterns are relative to current dir, so `^dacbarbos/(.*)$` in `/dacbarbos/.htaccess` will match `https://w3id.org/dacbarbos/dacbarbos/...`.
- I'm not sure what the `Options` or `RewriteCond` are for.

Keep it simple, use this:
```suggestion
RewriteRule ^(.*)$ https://github.com/dacbarbos/$1 [R=302,L]
```

With all this confusion, I'm curious why you are setting this up?  The common use case, and reason we started this service, was for projects where very long term stable URLs are required and the target may change over time.  There are other varied use cases too.  The simple redirect to the github UI is of limited use.  Most people redirect specific patterns to rendered html docs or raw data files like json, ttl, xml, etc.

orgadm: You're absolutely right. I spent some time browsing thru the existing IDs/folders prior to my PR. I even tried some and noticed broken targets. I will try to fix it over the weekend. I am tired now and need to rest. Thanks for your patience and expertise sharing. It's much appreciated.

dacbarbos/README.md
```diff
@@ -0,0 +1,7 @@
+# W3ID for /dacbarbos/
+
+## Workspace: DACBARBOS Brand's permanent ID
+
+## Contacts: DACBARBOS Brand's Organization Administrator (admin@dacbarbos.eu)
+
+### GitHub username: @orgadm
```
TallTed, line 7:
```suggestion
## Workspace

DACBARBOS Brand's permanent ID

## Contacts

DACBARBOS Brand's Organization Administrator (admin@dacbarbos.eu)
GitHub username: @orgadm
```

orgadm: PR #4621 is a follow-up to apply the suggested change.

PR #4620, opened 2024-12-23
Push coursesonto

naiayti:
New ontology

### DIFF
--- a/courses/.htaccess
+++ b/courses/.htaccess
@@ -1,0 +1,5 @@
+Options +FollowSymLinks
+RewriteEngine on
+
+# Match ontology
+RewriteRule ^ontology https://github.com/naiayti/TeachingKG/blob/main/ontology/ontology.owl [R=303,L]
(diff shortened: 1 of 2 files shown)


courses/.htaccess
```diff
@@ -0,0 +1,5 @@
+Options +FollowSymLinks
+RewriteEngine on
+
+# Match ontology
+RewriteRule ^coursesonto(.*)$ https://github.com/naiayti/TeachingKG/blob/main/ontology/ontology.owl [R=303,L]
```
davidlehn, line 5: Can you confirm that you want urls like `https://w3id.org/courses/coursesonto...`?  Asking since people sometimes duplicate the prefix.  It's different here, but thought I'd ask.

naiayti: Happy new year and thanks for the comment!

The right one is 
https://w3id.org/coursesonto

dgarijo: Then please change `RewriteRule ^coursesonto(.*)$` to be `RewriteRule ^(.*)$`

davidlehn: The directory name needs to change too.

naiayti: Thanks for the feedback, I am making a new push request

PR #4640, opened 2025-01-07
fixed regex

micheldumontier:


dgarijo:
I do not know, sorry. @davidlehn ?

davidlehn:
I'm not sure what "mod_lc" is, and no, it's not enabled.  I think the normal RewriteMap tricks don't work in per-directory .htaccess files.  I'm not sure what other solutions might be.  We do try to avoid extra complexities in this service.

micheldumontier:
no worries, i got around it :) thanks for your help!

PR #4670, opened 2025-01-22
Create new gerps ID

BMI24:
We decided to deprecate the GerPS-onto PURLs (multiple reasons, they were created without regards to our self-defined conventions) and instead use the lower-case gerps PURL


BMI24:
If I need to squash the merge commit with the actual change, please ping me

gerps/ontology/.htaccess
```diff
@@ -0,0 +1,7 @@
+# Turn off MultiViews
+Options -MultiViews
+
+# Rewrite engine setup
+RewriteEngine On
+
+RedirectMatch 301 ^/gerps/ontology/?$ https://fusion-jena.github.io/gerps-onto-landingpage/#
```
dgarijo, line 7: I don't think you want this, since you would be going to `/gerps/ontology/gerps/ontology`
In this case, it would suffice to do ^$

GerPS-onto/.htaccess
```diff
@@ -0,0 +1,4 @@
+# Turn off MultiViews
+Options -MultiViews
+
+Redirect 301 /GerPS-onto /gerps/ontology/process
```
dgarijo, line 4: This would go to the same id twice: `w3id.org/GerPS-onto//GerPS-onto`. Please review

dgarijo:
@BMI24 please have a look at the comments

gerps/ontology/.htaccess
davidlehn: This is probably ok since it's `RedirectMatch`.  Consider aligning with the rest of these files and use `RewriteRule` with ^$ instead.

GerPS-onto/.htaccess
davidlehn: I think this is ok since it's `Redirect`.

gerps/ontology/.htaccess
BMI24: Replaced it with `RewriteRule`, thanks! (but yeah, I tested it locally beforehand)

PR #4674, opened 2025-01-23
pull request for dsanno

chauraph:


dsanno/. htaccess
davidlehn, line 1: Looks like there is a spec in the file name.

dsanno/. htaccess
```diff
@@ -0,0 +1,11 @@
...
+# This space is administered by:
+#
+# Raphael Tsz-Kin Chau
+# GitHub username: chauraph
+
+RewriteEngine On
+
+# Redirect platform root to GitHub repository
+RewriteCond %{REQUEST_URI} ^/dsanno/platform/*.*$
+RewriteRule ^ https://github.com/chauraph/researchspace [R=302,L]
```
davidlehn, line 11: This looks strange with a condition with subpath, but rule with no path.  If it's tested, that's fine.

chauraph: Thanks for the review. It's tested. That's fine.

PR #4682, opened 2025-01-28
Add CoatO - an ontology for thin films

MGlauer:
CoatO is an ontology of thin film sputtering processes.


coato/.htaccess
```diff
@@ -0,0 +1,25 @@
+# # /gptto/
```
davidlehn, line 1:
```suggestion
# # /coato/
```

PR #4696, opened 2025-01-31
Fixed dseco redirect rule for ontology/.

GenEars:


GenEars:
Thanks @dgarijo, everything's fine.

PR #4697, opened 2025-01-31
Adding the spec-ontology to w3id.org.

RiX012:
Creating persistant URI's for the Specifications Ontology: https://docs.crow.nl/spec-ontology/


spec-ontology/README.md
```diff
@@ -0,0 +1,50 @@
+# /sml/
```
davidlehn, line 1: typo

RiX012: thanks. I will fix it

PR #4698, opened 2025-01-31
Fixed: rewrite rule for escaping `#`

RiX012:
My apologies, the rewrite rule did not escape the `#` character


RiX012:
Much obliged

PR #4706, opened 2025-02-04
Add ICS 500-13 and 500-15

IC-Standards:


davidlehn:
That target host has an untrusted cert and the URLs redirect to some other page.  (And they strangely don't allow two tabs?)  Still want this merged?

IC-Standards:
Yes the host is one that requires a login so that's the redirect. If you were logged in it would go straight to the file.

PR #4721, opened 2025-02-13
Updated .htaccess with right paths for owl file

karanthpallavi:


davidlehn:
The existing URL seems to work.  You want to redirect to the github UI instead?

rosettastmtont/.htaccess
```diff
@@ -2,8 +2,8 @@ Options +FollowSymLinks
 RewriteEngine on
 
 # Redirect the base URL to the latest version
-RewriteRule ^$ https://raw.githubusercontent.com/karanthpallavi/rosetta_statement_ontology/main/rosettastmtont-full.owl [R=303,L]
+RewriteRule ^$ https://github.com/karanthpallavi/rosetta_statement_ontology/blob/main/rosettastmtont-full.owl [R=303,L]
```
dgarijo, line 5: this will not serve the ontology, but the GitHub page where the ontology is.

karanthpallavi: Sorry, did not get this, the redirect URL was no tworking and hence changed redirect URL in .htaccess

davidlehn: As setup now, it's only redirecting the main directory:
```sh
$ curl -sI https://w3id.org/rosettastmtont/ | grep Location
Location: https://raw.githubusercontent.com/karanthpallavi/rosetta_statement_ontology/main/rosettastmtont-full.owl
```
There are various ways to write redirects to handle a full path as well, or both, if that's what you were looking for.

karanthpallavi: Thanks much! I see the redirection works fine!

PR #4722, opened 2025-02-13
Adding information for Gay Passport website.

Superraptor:
Added and tested for all example URLs.

### DIFF
--- a/GayPassport/.htaccess
+++ b/GayPassport/.htaccess
@@ -1,0 +1,61 @@
+# Example
+#
+# https://w3id.org/GayPassport/belfast redirects to https://belfast.gaypassport.com/
+# https://w3id.org/GayPassport/belfast/en redirects to https://belfast.gaypassport.com/en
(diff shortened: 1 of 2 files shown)


davidlehn:
Seems complex to remap an entire sites URL structure this way rather than use the original URLs directly.  More to maintain here, adds request latency, etc.  I'm curious what the use case is if you're willing to share.  If you have your reasons, that's fine.  This service was more targeted at long lived URLs for semantic web use where the targets may move.  This sort of general URL remapping is much less common.

Superraptor:
Use case is for external identifiers used on a Wikibase (https://lgbtdb.wikibase.cloud/wiki/Main_Page). Unfortunately they prepend the location and then append the other information. So for example, if I want to link to the Europa Hotel (https://belfast.gaypassport.com/en/outing/venue/europa-hotel) I have to create a separate Belfast Identifier to refer to it as an external identifier (since Wikibase only allows a single replacement for $1), then I'd have to create another one for London (https://london.gaypassport.com/en/outing/venue/strand-palace) and so on and so forth. GayCities does the same thing, which is why I added that one as well.

PR #4726, opened 2025-02-14
Semantic Time Series Ontology (SemTS)

alexgFIT:
Initial pull request for the Semantic Time Series Ontology (SemTS)

https://semts-ontology.github.io/SemTS/

Description:
SemTS is an ontology to semantically structure insights gained from multivariate time series analyses combined with domain-specific information.


SemTS/.htaccess
```diff
@@ -0,0 +1,48 @@
...
+AddType text/turtle .ttl
+AddType application/n-triples .n3
+AddType application/ld+json .jsonld
+# Rewrite engine setup
+RewriteEngine On
+
+# Rewrite rule to serve JSON-LD content from the vocabulary URI if requested
+# Placed before HTML to support serving JSON-LD from a browser (e.g., JSON Playground)
+RewriteCond %{HTTP_ACCEPT} application/ld\+json
+RewriteRule ^(ontology)(/|#)*$ ^$ https://semts-ontology.github.io/SemTS/ontology.jsonld [R=303,L]
```
davidlehn, line 18: - these patterns are strange. did you test them?
- they are capturing matches, but not using them.
- the `#` is (usually) a client side only thing.
- is the last pattern going to appear 0 or more times?

You might want patterns to be more like the following to simplify and allow optional trailing slash (please test):
```suggestion
RewriteRule ^ontology/?$ https://semts-ontology.github.io/SemTS/ontology.jsonld [R=303,L]
```

davidlehn (commented):
Review comments and please squash commits if you can.

SemTS/.htaccess
alexgFIT: Hi,
first of all, thanks a lot for helping out!
- Patterns were changed according to your suggestion (Thanks)
- semts changed to SemTS in the path
- Added the github tag to the README
- Removed incorrect/empty ontotsk files

I hope the patterns are correct now. Tried to test the rules locally, but for some reason the redirect didn't work and resulted in a 404 (despite common adjustments in the httpd.conf, including directory access and loading the rewrite module). Squashing commits also led to problems. If both is mandatory I will have another look at the problems and try to resolve them.

PR #4738, opened 2025-02-20
Update .htaccess for Ecoportal

r0sek:
Please replace existing rewrite rule for CMECS with this one. Thank you.

### DIFF
--- a/CMECS/.htaccess
+++ b/CMECS/.htaccess
@@ -8,7 +8,9 @@
 
 RewriteEngine on
 
-#redirect terms to mmisw.org
+#redirect terms to https://ecoportal.lifewatch.eu/ontologies
 #test using https://w3id.org/CMECS/CMECS_00000389 (Geologic Substrate)
 
-RewriteRule ^(.*)$ https://mmisw.org/ont?iri=https://w3id.org/CMECS/$1 [R=302,L]
+RewriteRule ^CMECS/?$ http://ecoportal.lifewatch.eu/ontologies/CMECS [R=301,L]
+RewriteCond %{REQUEST_URI} ^.*CMECS/.*$
+RewriteRule ^.*/([^/#]+)/?$ http://ecoportal.lifewatch.eu/ontologies/CMECS/$1 [R=301,L]


CMECS/.htaccess
```diff
@@ -8,7 +8,9 @@
...
 RewriteEngine on
 
-#redirect terms to mmisw.org
+#redirect terms to https://ecoportal.lifewatch.eu/ontologies
 #test using https://w3id.org/CMECS/CMECS_00000389 (Geologic Substrate)
 
-RewriteRule ^(.*)$ https://mmisw.org/ont?iri=https://w3id.org/CMECS/$1 [R=302,L]
+RewriteRule ^CMECS/?$ http://ecoportal.lifewatch.eu/ontologies/CMECS [R=301,L]
+RewriteCond %{REQUEST_URI} ^.*CMECS/.*$
+RewriteRule ^.*/([^/#]+)/?$ http://ecoportal.lifewatch.eu/ontologies/CMECS/$1 [R=301,L]
```
davidlehn, line 16: What is the intention here?  That first rule will match https://w3id.org/CMECS/CMECS, which is usually not what people want.  The second part looks complex.  Reminder that rules are relative to the current directory.  Maybe that can be simplified to something like:
```suggestion
RewriteRule ^$ http://ecoportal.lifewatch.eu/ontologies/CMECS [R=301,L]
RewriteRule ^(.*)$ http://ecoportal.lifewatch.eu/ontologies/CMECS/$1 [R=301,L]
```
Or even send every subpath to your endpoint:
```suggestion
RewriteRule ^(.*)$ http://ecoportal.lifewatch.eu/ontologies/CMECS/$1 [R=301,L]
```

r0sek: Hello, and thank you!

Each Class in the [CMECS ontology ](https://ecoportal.lifewatch.eu/ontologies/CMECS?p=summary) has its own URL e.g., `https://w3id.org/CMECS/CMECS_00000389`. The intention is to redirect those URLs to resolve the Class page on the Ecoportal site e.g., [Geological Substrate](http://ecoportal.lifewatch.eu/ontologies/CMECS/CMECS_00000389).

I had copied and pasted autogenerated rewrite rules from the Ecoportal site but I see now that the first rule `RewriteRule ^CMECS/?$ http://ecoportal.lifewatch.eu/ontologies/CMECS [R=301,L]` probably should not be included. 

I [tested here](https://htaccess.madewithlove.com?share=6e24fd85-b977-44ca-93c1-c81bd6865469) with just
`RewriteCond %{REQUEST_URI} ^.*CMECS/.*$`
`RewriteRule ^.*/([^/#]+)/?$ http://ecoportal.lifewatch.eu/ontologies/CMECS/$1 [R=301,L]`
and the output URL does seem to correctly resolve to the expected ontology Class page.

If that seems ok to you, I will make that change and resubmit the PR. 

I also [tested your suggestion here](https://htaccess.madewithlove.com?share=298e5358-52fd-46dd-977f-6e851027b699) but that only resolves to the ontology Summary page. I had the same result with your second suggestion as well.

davidlehn: I don't have time to test at the moment.  That test site tests rules in the root dir.  This site has them in directories.  Some adjustment is needed to account for the relative nature of per-dir rules.  Let us know what you want to do. 
 Can always fix later.

r0sek: Ok thx. I now remember the same issue with the htaccess tester from an earlier change.

Would you please go ahead and merge the commit as it is, then I can go back and check with the Ecoportal admins when/if it fails. Thanks!

PR #4750, opened 2025-02-25
sco - revisions for permanent links

gioUbbiali:


dgarijo:
@gioUbbiali you are changing the case sensitivity. Is this intended?
We usually recommend avoiding capital letters in w3ids, as it messes things up in certain systems.

gioUbbiali:
Hi @dgarijo,
I just noticed I wrote in capital letters. So, for the reasons you mentioned, I am modifying the repo.
Thanks.

PR #4769, opened 2025-03-07
codemeta: properly rewrite versioned requests for json-ld

proycon:
Fixes: https://github.com/codemeta/codemeta/issues/376


proycon:
Note: this is untested so please review and test, I think it'll fix the referenced issue.

dgarijo:
Since this may make things clearer, I am approving it

PR #4778, opened 2025-03-11
Request for permanent URI: https://w3id.org/my-ontology

Umi0529:
I would like to request a permanent URI for my ontology.

The new URI: https://w3id.org/my-ontology
It redirects to: https://umi0529.github.io/my-ontology/

The `.htaccess` file has been added to handle content negotiation, allowing different formats to be accessed.

Thank you for your time and consideration!

### DIFF
--- a/my-ontology/.htaccess
+++ b/my-ontology/.htaccess
@@ -1,0 +1,18 @@
+Options +FollowSymLinks
+RewriteEngine on
+
+# HTML 文件（人類可讀）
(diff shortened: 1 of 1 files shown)


davidlehn:
The ttl link is the only one that exists now.  Are the other ones going to be setup soon?

davidlehn:
Can you please also add the maintainer(s) github ids either in .htaccess comments or in a README.md file.

PR #4779, opened 2025-03-11
Geomapi ontology

JelleKUL:
Added the Geomapi ontology permalink from the Geomatics research group at KU Leuven, Belgium

### DIFF
--- a/geomapi/.htaccess
+++ b/geomapi/.htaccess
@@ -10,6 +10,8 @@
 #
 # Jelle Vermandere
 # GitHub username: JelleKUL
+# Maarten Bassier
+# Github username: Saiga1105
 
 RewriteEngine on
-RewriteRule ^ https://raw.githubusercontent.com/KU-Leuven-Geomatics/geomapi/refs/heads/main/geomapi/ontology/ [R=302,L]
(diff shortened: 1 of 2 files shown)


geomapi/.htaccess
```diff
@@ -0,0 +1,27 @@
...
+# Jelle Vermandere
+# GitHub username: JelleKUL
+# Maarten Bassier
+# Github username: Saiga1105
+
+Options +FollowSymLinks
+RewriteEngine on
+
+# Redirect requests for geomapi ontology to the raw GitHub URL
+RewriteRule ^geomapi$ https://raw.githubusercontent.com/KU-Leuven-Geomatics/geomapi/refs/heads/main/geomapi/ontology/geomapi_ontology.ttl [R=303,L]
```
davidlehn, line 16: Rules are relative to the current dir.  Did you want URLs to be `https://w3id.org/geomapi/geomapi`? You might mean to use `^$`.

JelleKUL: I want  `https://w3id.org/geomapi/geomapi` to point to the file indeed. I will update the RewriteRule, thanks

PR #4783, opened 2025-03-12
Geomapi Permalink

JelleKUL:
I would like to add the Geomapi permalink to this repo for our defined ontology Geomapi.

### DIFF
--- a/geomapi/.htaccess
+++ b/geomapi/.htaccess
@@ -10,6 +10,8 @@
 #
 # Jelle Vermandere
 # GitHub username: JelleKUL
+# Maarten Bassier
+# Github username: Saiga1105
 
 RewriteEngine on
-RewriteRule ^ https://raw.githubusercontent.com/KU-Leuven-Geomatics/geomapi/refs/heads/main/geomapi/ontology/ [R=302,L]
+RewriteRule ^ https://raw.githubusercontent.com/KU-Leuven-Geomatics/geomapi/refs/heads/main/geomapi/ontology/geomapi_ontology.ttl [R=302,L]
--- a/geomapi/.htaccess
+++ b/geomapi/.htaccess
@@ -1,0 +1,15 @@
+# # /examples/
+#
+# A collection of URL redirection examples.
+#
+# https://w3id.org/geomapi/ redirects to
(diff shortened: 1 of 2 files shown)


geomapi/.htaccess
```diff
@@ -0,0 +1,27 @@
...
+
+Options +FollowSymLinks
+RewriteEngine on
+
+# Redirect requests for geomapi ontology to the raw GitHub URL
+RewriteRule ^$ https://raw.githubusercontent.com/KU-Leuven-Geomatics/geomapi/refs/heads/main/geomapi/ontology/geomapi_ontology.ttl [R=303,L]
+
+# Content negotiation for different RDF formats
+RewriteCond %{HTTP_ACCEPT} application/rdf\+xml
+RewriteRule ^$ https://raw.githubusercontent.com/KU-Leuven-Geomatics/geomapi/refs/heads/main/geomapi/ontology/geomapi_ontology.ttl [R=303,L]
```
davidlehn, line 20: Should this point at a .xml file?

geomapi/.htaccess
```diff
@@ -0,0 +1,27 @@
...
+
+RewriteCond %{HTTP_ACCEPT} text/turtle
+RewriteRule ^$ https://raw.githubusercontent.com/KU-Leuven-Geomatics/geomapi/refs/heads/main/geomapi/ontology/geomapi_ontology.ttl [R=303,L]
+
+RewriteCond %{HTTP_ACCEPT} application/ld\+json
+RewriteRule ^$ https://raw.githubusercontent.com/KU-Leuven-Geomatics/geomapi/refs/heads/main/geomapi/ontology/geomapi_ontology.ttl [R=303,L]
```
davidlehn, line 26: Should this point at a .json or .jsonld file?

geomapi/.htaccess
JelleKUL: No, I only need the link to the ttl file

geomapi/.htaccess
JelleKUL: No, I only need the link to the ttl file

dgarijo: But that is wrong.  If  want to return only ttl, why don't you only return the TTL with a condition?

PR #4784, opened 2025-03-12
Update .htaccess

pedropaulofb:
Adding redirections and content type negotiation in laderr's htaccess file.

### DIFF
--- a/laderr/.htaccess
+++ b/laderr/.htaccess
@@ -5,8 +5,11 @@ AddType text/turtle .ttl
 # RESOURCES' VERSIONING ----------------------------------------------------------------------------------------
 ## Latest release
 RedirectMatch 302 ^/laderr/(latest|current)/?$  https://github.com/pedropaulofb/laderr/releases/latest
+RedirectMatch 302 ^/laderr/engine/(latest|current)/?$  https://github.com/pedropaulofb/laderr-engine/releases/latest
+
 ## All releases
 RedirectMatch 302 ^/laderr/releases/?$  https://github.com/pedropaulofb/laderr/releases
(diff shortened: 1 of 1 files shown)


laderr/.htaccess
```diff
@@ -1,4 +1,39 @@
...
+RedirectMatch 302 ^/laderr/engine/?$  https://pedropaulofb.github.io/laderr-engine
+RedirectMatch 302 ^/laderr/engine/(git|repo)/?$  https://github.com/pedropaulofb/laderr-engine
+
+# CONTENT NEGOTIATION -----------------------------------------------------------------------------------------
+
+## Redirect to HTML if no specific format is requested
+RewriteCond %{HTTP_ACCEPT} !text/turtle
+RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml
+RewriteCond %{HTTP_ACCEPT} !application/ld\+json
+RewriteRule ^laderr/?$ https://pedropaulofb.github.io/laderr/ [R=303,L]
```
davidlehn, line 30: RedirectMatch regex is on the whole URL, RewriteRule is relative to the current dir so you don't want the leading `laderr` prefix.  Also rules are processed in order, so I think the above RedirecMatch is going to take effect before these content negotiation ones.  You could move them before the similar RedirectMatch pattern to fix.  I'd also recommend testing this.

pedropaulofb: Thank you very much for your review, @davidlehn. I’ve made the following updates:  

- Moved content negotiation rules to before `RedirectMatch` to ensure they are processed first.  
- Removed `/laderr/` prefix from `RewriteRule` patterns.  
- Tested the configuration (considering some limitations) at [htaccess.madewithlove.com](https://htaccess.madewithlove.com/).  

Let me know if any further adjustments are needed! 

Thank you again for your support! 

Kind regards, 
Pedro.

PR #4791, opened 2025-03-14
Revisions

gioUbbiali:
We revised the SCO htaccess file.

### DIFF
--- a/sco/.htaccess
+++ b/sco/.htaccess
@@ -5,9 +5,7 @@ AddType text/turtle .ttl
 # CONTENT NEGOTIATION ----------------------------------------------------------------------------------------
 
 ## Serve TTL content when RDF formats are explicitly requested
-RewriteCond %{HTTP_ACCEPT} text/turtle [OR]
-RewriteCond %{HTTP_ACCEPT} application/rdf\+xml [OR]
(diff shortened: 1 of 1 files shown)


sco/.htaccess
```diff
@@ -8,13 +8,13 @@ AddType text/turtle .ttl
 RewriteCond %{HTTP_ACCEPT} text/turtle [OR]
 RewriteCond %{HTTP_ACCEPT} application/rdf\+xml [OR]
 RewriteCond %{HTTP_ACCEPT} application/ld\+json
-RewriteRule ^sco/?$ https://gioUbbiali.github.io/Sustainability-Core-Ontology/SCO.ttl [R=303,L]
+RewriteRule ^$ https://gioUbbiali.github.io/Sustainability-Core-Ontology/SCO.ttl [R=303,L]
```
dgarijo, line 11: here you are technically doing an incorrect redirect: when people ask for rdf/xml or json-ld, you return TTL. 
I suggest removing lines 9 and 10

dgarijo: and the [OR] in line 8

dgarijo:
Please @gioUbbiali see my comment above

gioUbbiali:
We made revisions accordingly. Thanks!

PR #4794, opened 2025-03-14
add AKSW identifier

SimonBin:
for @aksw projects

- also including https://w3id.org/aksw/norse# namespace for Scaseco/jenax

Thanks


SimonBin:
hum, ideally all group members should be allowed to make updates. I've added some key persons to the readme

PR #4799, opened 2025-03-17
Add JSON-LD redirect for `fair/fip/terms` URL

anatoly-scherbakov:
https://w3id.org/fair/fip/terms/ features RDF/XML, TriG, Turtle, N-Triples versions provided via Content Negotiation but seemingly does not provide that same thing for JSON-LD version, even though the JSON-LD version is available.

Creating such a redirect by analogy.


davidlehn:
@tkuhn Please review.

tkuhn:
Yes, that looks good and can be merged.

PR #4818, opened 2025-03-25
Doc-KG Entry for the SALMON Project

srodriguez142857:
Doc-KG: Unstructured Documents to Knowledge Graph Construction, Identification and Validation with Wikidata.


dgarijo:
@srodriguez142857 can you please add yourself as a maintainer?

davidlehn:
I think it's fine, github id is in the /salmon readme.  Might be easier to add simple sub-paths like this to the parent .htaccess, but it's ok.

PR #4821, opened 2025-03-27
Added redirect for PerManOnt

CS22S010:


dgarijo:
@CS22S010 please add your github id in the readme file

CS22S010:
Hi @dgarijo I have added the github id in the readme file

dgarijo:
Thanks. I see the identifier has capital and non capital letters.
Would it be possible to use non capital letters only? Because this causes resolvability issues in some cases. For example, if someone were to create `permanont` redirecting to another URL.

CS22S010:
Thanks for the feedback. I have changed it to lowercase

PR #4831, opened 2025-04-01
added aco

cameronmore:
Added the agent capability ontology path according to the ontology id example.

### DIFF
--- a/aco/.htaccess
+++ b/aco/.htaccess
@@ -1,0 +1,7 @@
+RewriteEngine On
+
+# Redirect the base ontology URI to a specific page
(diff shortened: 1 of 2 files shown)


dgarijo:
@cameronmore please add a maintainer in the README file

cameronmore:
@dgarijo fixed, thank you

cameronmore:
How soon should I expect the merge to content to 'go live'? I have never written an htaccess file before and was curious to know when I should expect it to be working and make a correction if it's not, thank you

davidlehn:
Merges trigger the live system to update.  Maybe a minute or less to update.  (Might take a bit longer in the future, but should be fast enough for most use cases.)

PR #4840, opened 2025-04-02
new path for aco

cameronmore:
I am trying to direct w3id.org/ontology/AgentCapabilityOntology to my documentation page and use that as the namespace for new classes, properties, etc, but let me know if I am doing it wrong and how I can fix it. If it's easier, I would be fine omitting '/ontology/' from the URI.


dgarijo:
@cameronmore I recommend having a look at https://github.com/perma-id/w3id.org/tree/master/example

dgarijo:
@cameronmore why don't you call it `aco` so it's w3id.org/aco? It seems easier, no?

PR #4841, opened 2025-04-02
Add aco

cameronmore:
@dgarijo I agree, this is easier


dgarijo:
Please add a readme file with the maintainer. And I recommend adding content negotiation to an RDF serialization, but that's up to you.

cameronmore:
The aco directory should already contain a readme, this just updates the rules, and I may but I just want to get something minimally working for the moment

PR #4842, opened 2025-04-02
add new identifier and rules

laurentfintoni:
added new identifier (KNOT + sub directories for specific projects) and first set of redirect rules


davidlehn:
Please add the maintainer github user ids to at least the top level README.

laurentfintoni:
GH user ID for maintainer added to top level README.

dgarijo:
@laurentfintoni capital letters in identifiers may lead to problems. Can you please change the capitalization to `knot`?
We have faced this in the past, with mixed capitalization, and we recommend avoiding it all together now.
Also, it's ok for maintainer to be only added in the root directory, but your edit is correct too.

laurentfintoni:
@dgarijo understood, thanks. changes done!

PR #4856, opened 2025-04-08
GerPS Ontologies: fix broken content negotation

BMI24:
I am a bit inexperienced with .htaccess files, so I overlooked that `RewriteCond` only applies to the next `RewriteRule` (which of course makes sense). I fixed that by adding some bloat to our .htaccess files. 

If anyone has a good idea how to simplify the files, I am open to recommendations.


davidlehn:
Yeah the .htaccess setup can get tricky when the rules get more complex.  I suspect this site might be able to take advantage of a simpler meta-config that generates the verbose .htaccess rules.  But doing that at the scale here and with absentee maintainers, it might be a mess.

mod_negotiation is, in theory, more appropriate to handle proper language and accept headers with q params and so on.  The simple regex checks can fail to work for many reasons, but hasn't been an issue in practice it seems.  I fiddled around with that recently but was having trouble making it work and didn't even get to tricks to get it to work with redirects.

There are probably ways to make all this work better at the cost of additional complexity and tooling.  Anyone with those sorts of use cases could always blindly redirect all traffic to their own server that handles the advanced processing.

As a minor simplification, I don't think the "AddType" lines do anything since this service is only doing redirects.  Those are for setting content type on local files.

BMI24:
Thank you, ill think about your pointer!

PR #4860, opened 2025-04-09
Add gfo-core config

ChristophB:
This PR adds redirects to the General Formal Ontology (core version), which contains the core entities of GFO.

### DIFF
--- a/gfo-core/.htaccess
+++ b/gfo-core/.htaccess
@@ -1,0 +1,61 @@
+# Turn off MultiViews
+Options -MultiViews
+
+# Directive to ensure *.rdf files served as appropriate content type,
+# if not present in main apache config
+AddType application/rdf+xml .rdf
+AddType application/rdf+xml .owl
+AddType text/turtle .ttl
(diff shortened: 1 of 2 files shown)


gfo-core/.htaccess
```diff
@@ -0,0 +1,61 @@
...
+# if not present in main apache config
+AddType application/rdf+xml .rdf
+AddType application/rdf+xml .owl
+AddType text/turtle .ttl
+AddType application/n-triples .nt
+AddType application/ld+json .jsonld
+
+RewriteEngine on
+
+RewriteRule ^release/?$ "/"
```
davidlehn, line 14: This seems odd.  I see it's in gfo-light too.  What's it supposed to do?

ChristophB: Thank you for the hint!
The idea was to rewrite `w3id.org/gfo-core/release` (without an actual release tag at the end) to `w3id.org/gfo-core`. Looks like this is not working as intended. I'll provide a fix in a new PR.

PR #4869, opened 2025-04-15
new rewrite rule with location.hash

jhoernsc:
Not sure if that works: redirection with location hash handling:

```RewriteRule ^ontology[/#](.+)$ https://dhi-roma.github.io/g17-ontology/docs/widoco/doc/index.html#$1  [R=303,L,NE]```

Do you have any idea?


dgarijo:
In theory for hashes you don't need to redirect, as it goes directly to the target + anchor. What is it that you'd like to support?

jhoernsc:
Thank you very much,

everything works as expected. We wanted to make sure that such IDs redirect to the corresponding entries in the Widoko documentation:

https://w3id.org/g17/ontology#church

https://w3id.org/g17/ontology/church

PR #4875, opened 2025-04-16
Ontology folder renamed from g17 to grace

jhoernsc:
This reflects more our practice. With g17 we rather mean our Knowledge Graph with data from the 17th century. Our ontology GRACE is suitable for modeling papal graces in general.


dgarijo:
@jhoernsc w3id are supposed to be persistent. Why don't you add a redirection from g17 to grace, just so anyone still goes to the previous link?

jhoernsc:
We had created the g17 ID 4 days ago, nobody really knows it. It was a mistake, sorry. We can also leave the g17 folder and use it for our knowledge graph and I will create a new "grace" w3id if that is the better way.

dgarijo:
If it's not really used, then it should be ok.

jhoernsc:
Thanks!

Another question:

if we use w3ids with curl we don't get the expected turtle output, that we get  if we use the github link directly without redirection.

curl -H "Accept: text/turtle" https://dhi-roma.github.io/grace-ontology/ontology/grace.ttl

gives us the expected text/turtle response.

curl -H "Accept: text/turtle" https://w3id.org/grace/ontology

gives us an only an HTML response:

<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>303 See Other</title>
</head><body>
<h1>See Other</h1>
<p>The answer to your request is located <a href="https://dhi-roma.github.io/grace-ontology/ontology/grace.ttl">here</a>.</p>
<hr>
<address>Apache/2.4.29 (Ubuntu) Server at w3id.org Port 443</address>
</body></html>

Is there a way to get the type that is asked for in the application header of the rewriteRule?

dgarijo:
w3id is a redirect server. You need to add the `-L` flag:
```
curl -sH "accept:text/turtle" -L https://w3id.org/grace/ontology
```

jhoernsc:
But then how could I use tools like WebVowl:

https://service.tib.eu/webvowl/#iri=https://w3id.org/grace/ontology

dgarijo:
I can't get the ontology to load in protege or widoco. I think the content negotiation is not correct. 
Have you validated the file independently?

Bear in mind that the rules are independent. For example:
```
# Rewrite rule to serve <text/turtle>
RewriteCond %{HTTP_ACCEPT} ^.*text/turtle.* 
RewriteRule ^ontology/?$ https://dhi-roma.github.io/grace-ontology/ontology/grace.ttl [R=303,L]
RewriteRule ^ontology/(([0-9]+(?:\.[0-9]+)+))/?(.*)$ https://dhi-roma.github.io/grace-ontology/ontology/$1/grace.ttl#$3 [R=301,L,NE]
```
the rewriteCond applies only to the first rewriteRule.

dgarijo:
Basically, what happens is that it's returning html when these resources ask for rdf/xml. I would support additional types or return a default that is the ttl representation:

```
(base) dgarijo@pop-os:~/Desktop$ curl -sH "accept:application/rdf+xml" -L https://w3id.org/grace/ontology
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>404 Not Found</title>
</head><body>
<h1>Not Found</h1>
<p>The requested URL was not found on this server.</p>
<hr>
<address>Apache/2.4.29 (Ubuntu) Server at w3id.org Port 443</address>
</body></html>
```

jhoernsc:
Thanks a lot, it works now!

PR #4879, opened 2025-04-17
Add flags for rules in /np/ and /kpxl/

tkuhn:


dgarijo:
I am curious, why 307 and not 302?

tkuhn:
I think for these it doesn't matter, but POST request only work properly with 307. So I got into the habit of only using 307, and I haven't experienced a downside yet from this practice :)

PR #4886, opened 2025-04-24
Uploading itandrec

Numa84:
itandrec is an example ontology for educational purposes.
The ontology is about tourism.


davidlehn:
- This service does not host content, only redirects to your content hosted elsewhere.
- There is no .htaccess for the redirects.
- The readme.md is from another project.  Update it with your info.
- The index.html can not be hosted here and is large.  Remove it from the PR and git history (use squash or other techniques).

davidlehn:
Status?

dgarijo:
@Numa84 ping

PR #4891, opened 2025-04-26
Add `evr` namespace for the Event Venue Registry

cthoyt:
This [W3ID](https://w3id.org) provides persistent URLs for entities and artifacts in the [Event Venue Registry (EVR)](https://github.com/event-venue-registry/evr), a database and ontology cataloging and characterizing physical venues at which events (such as conferences) can take place.

This PR adds a configuration under the `evr` namespace for redirecting identifiers in this database and for resolving ontology artifacts generated from this database.

### DIFF
--- a/evr/.htaccess
+++ b/evr/.htaccess
@@ -1,0 +1,16 @@
+Options +FollowSymLinks
+RewriteEngine on
+
+# Redirects for entities
+#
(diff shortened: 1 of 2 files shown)


cthoyt:
thank you @dgarijo :)

PR #4892, opened 2025-04-27
Add redirect GeoSFAIR first example

CipoCar:
Added a new GeoSFAIR/metadata directory with .htaccess and README.md files to handle identifier redirects.


dgarijo:
@CipoCar please void capitalization on your w3id. It causes problems in different systems.Can you please make it `geosfair`?

CipoCar:
> @CipoCar please void capitalization on your w3id. It causes problems in different systems.Can you please make it `geosfair`?
@dgarijo Is the request related to the rule in .htaccess or to the general folder?

PR #4893, opened 2025-04-27
Cultural Artefacts' Contextual Ontology (CACAO)

RubenPeeters:


cacao/.htaccess
```diff
@@ -0,0 +1,16 @@
+<!--
```
davidlehn, line 1: Comments are lines that start with `#`.  This style will cause the server to fail on this directory.
https://httpd.apache.org/docs/2.4/configuring.html#syntax

RubenPeeters: Thank you for pointing this out, I opened a new pull request to resolve this issue

PR #4900, opened 2025-04-29
refine the rewriterule for Datacite endpoint

ACz-UniBi:
The encoded url doesn't seems to work. Refining without encoding - happy of suggestions


dgarijo:
Merging, but I think this will not wok

ACz-UniBi:
brief feedback: the current rule works well :)

PR #4901, opened 2025-04-30
Added folder for redirection links to MALFO ontology

kataph:


kataph:
Many thanks for the prompt answer. 
I implemented the required changes, it should be ok now.

PR #4915, opened 2025-05-05
Addition of 'glue' folder

TheRazorace:
Addition of glue folder which directs to GLUE ontology documentation and ontology files

### DIFF
--- a/glue/.htaccess
+++ b/glue/.htaccess
@@ -8,22 +8,22 @@ RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+
 RewriteCond %{HTTP_ACCEPT} text/html [OR]
 RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
 RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
-RewriteRule ^(.*)\.conneg$ https://dtai-kg.github.io/GLUE-Ontology [NE,R,L]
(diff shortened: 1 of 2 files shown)


dgarijo:
@TheRazorace please add your github user in the README file.
Also, are you sure you want to add the /conneg in every URL?

TheRazorace:
Dear @dgarijo thank you for your comments.
1. I have added the associated Github organization responsible for the project
2. I am not sure about the usage of /conneg in every URL. I adopted it from a previous folder we had. Feel free to edit it or suggest changes. For me, the most important thing is the redirection to the html page.

dgarijo:
- The suggestion is to delete `\.conneg` from your URLs.
- Please add `[TheRazorace](https://github.com/TheRazorace)` in your readme so we know it's you. 
Thanks

TheRazorace:
Thank you for the suggestions! I tried to incorporate both. Could you check now?

glue/.htaccess
```diff
@@ -0,0 +1,29 @@
+Header set Access-Control-Allow-Origin *
+Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
+Options +FollowSymLinks
+RewriteEngine on
+
+## CONTENT NEGOTIATION
+RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+xml) [OR]
```
davidlehn, line 7: You should remove the `[OR]` on this line.  This would be true for all the types you are trying to check for below.

PR #4916, opened 2025-05-05
Create .htaccess for recipe-model

jpcik:
Added .htaccess redirections for the SmartEdge Recipe-model schema


dgarijo:
https://aislab-hevs.github.io/smartedge-ontologies/recipe-model/doc/index-en.html is a 404. Is this correct?

PR #4924, opened 2025-05-09
Add redirect for orbac.owl

blueontologies:
This PR adds a redirect from https://w3id.org/orbac.owl to an ontology hosted on GitHub.


dgarijo:
Please add a readme file indicating who will maintain this w3id.
Also, maybe you want to remove the `.owl` in the w3id?

blueontologies:
Added the requested README file with maintainer info.

I prefer to keep the .owl in the W3ID as it matches the naming used in my research artifacts and ensures consistency with existing references. 

Thank you very much.

PR #4929, opened 2025-05-11
OntoGSN: fixed 404 issues and added redirect for homepage

Tomas-Bueno-Momcilovic:


OntoGSN/.htaccess
```diff
@@ -1,5 +1,21 @@
...
+
+# MIME type for OWL
+AddType application/rdf+xml .owl
 
 RewriteEngine on
-RewriteRule ^ https://fortiss.github.io/ontogsn/ontology [R=302,L]
\ No newline at end of file
+
+# Define a variable for the actual file location
+SetEnvIf Request_URI ^/ontology/?$ ONTOLOGY_URL=https://raw.githubusercontent.com/fortiss/OntoGSN/main/ontogsn.owl
```
davidlehn, line 14: I think these are relative to the root?  So they would need OntoGSN at the start.  Was this checked?  It looks like it might be easier to put the target URLs in the RewriteRules below.

Tomas-Bueno-Momcilovic: Thanks for catching the issue. I modeled according to the polifonia .htaccess file, so I introduced this mistake. Don't know how to test it locally.

PR #4940, opened 2025-05-14
Fix mbdo per-directory rewrite rules to support optional slash and subpathsUpdate htaccess

samar-aqlan:
This PR updates the mbdo/.htaccess file to correct Apache’s per-directory rewrite behavior:

- Enable the rewrite engine.
- Redirect both /mbdo and /mbdo/ to the GH Pages root: https://samar-aqlan.github.io/MBDO-ManagingBridgeDataOntology/
- Proxy any subpath under /mbdo/ (e.g., /mbdo/foo/bar.ttl) to the corresponding resource on GH Pages.

Previously, requests to /mbdo (without trailing slash) fell through to a directory index. These rules ensure a 301 redirect for both slash‐and-no‐slash and all subpaths.

### DIFF
--- a/mbdo/.htaccess
+++ b/mbdo/.htaccess
@@ -1,5 +1,10 @@
-# 开启重写引擎
+# mbdo/.htaccess  —— Make slash optional and proxy subpaths
+
+# 1) Enable rewrite engine
 RewriteEngine On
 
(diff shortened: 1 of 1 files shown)


davidlehn:
Please resolve the conflicts.

davidlehn:
ping

PR #4942, opened 2025-05-15
Update .htaccess

essepuntato:
Added content negotiation for Spoke 4 entities

### DIFF
--- a/changes/.htaccess
+++ b/changes/.htaccess
@@ -1,5 +1,63 @@
+# Allow all (*) origins to access the resource on the server
 Header set Access-Control-Allow-Origin *
+
+# Allow following symbolic links / symlinks / softlinks
 Options +FollowSymLinks
+
+# Turn off MultiViews
+Options -MultiViews
+
+# Ensure files are served as appropriate content type
+AddType application/rdf+xml .rdf
+AddType application/rdf+xml .xml
+AddType text/turtle .ttl
+AddType application/n-triples .n3
+AddType application/ld+json .jsonld
+
+# Required for the redirects
 RewriteEngine on
 
-RewriteRule ^(.*)$ https://www.fondazionechanges.org [R=302,L]
+######################################################
+# Redirects for Spoke 4 - BEGIN
+######################################################
+
+## Redirects - with extensions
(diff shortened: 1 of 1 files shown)


changes/.htaccess
```diff
@@ -1,5 +1,63 @@
...
+RewriteRule ^(4/.+).(json|jsonld)$ https://dharc-org.github.io/spuk/changes/$1.jsonld [R=303,L]
+RewriteRule ^(4/.+).(rdf|xml)$ https://dharc-org.github.io/spuk/changes/$1.xml [R=303,L]
+RewriteRule ^(4/.+).html$ https://dharc-org.github.io/spuk/changes/$1.html [R=303,L]
+RewriteRule ^(4/.+).nt$ https://dharc-org.github.io/spuk/changes/$1.nt [R=303,L]
+RewriteRule ^(4/.+).ttl$ https://dharc-org.github.io/spuk/changes/$1.ttl [R=303,L]
+
+# Serve JSON-LD content from the vocabulary URI if requested
+RewriteCond %{HTTP_ACCEPT} application/json [OR]
+RewriteCond %{HTTP_ACCEPT} application/ld\+json
+RewriteRule ^(4/.+)$ /changes/$1.jsonld [R=303,L]
```
davidlehn, line 34: This will work, but it results in an extra redirect over the network.  It's better for clients and this service to use full URLs here.  Is the traffic rare enough that it's not an issue?

essepuntato: It is a strategy I've already used (which works very well) for other w3id.org sites with a low traffic rate (as the current one, i.e. "changes"). It would be ideal to have the same approach here, as it would simplify management, given that more people will be involved in handling it in the following months. Consider that this set-up is done for content negotiation purposes, but not for applicative scenarios (which will be dealt with queries to a SPARQL endpoint external to this side, rather than accessing the resource IRI).

Considering this, would it be possible to merge this pull request as it is?

davidlehn: Ok.  If the issue is avoiding URL repetition, there are some env var tricks that could be used to only have it in one place.  Will merge now since this can be adjusted later.

PR #4955, opened 2025-05-19
Update .htaccess

andremann:
adding rewrite rules for OpenAPI

### DIFF
--- a/skg-if/.htaccess
+++ b/skg-if/.htaccess
@@ -88,7 +88,11 @@ RewriteRule ^context/skg-if\.json$ https://skg-if.github.io/context/ver/current/
 ### SKG-IF Context [end]
 ## Handling core entities [end]
 
-
 ### Default behaviour [start]
 RewriteRule ^(.*)$ https://skg-if.github.io/$1 [R=302,L]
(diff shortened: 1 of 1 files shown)


skg-if/.htaccess
```diff
@@ -88,7 +88,11 @@ RewriteRule ^context/skg-if\.json$ https://skg-if.github.io/context/ver/current/
...
 
-
 ### Default behaviour [start]
 RewriteRule ^(.*)$ https://skg-if.github.io/$1 [R=302,L]
 ### Default behaviour [end]
+
+### SKG-IF API [start]
+RewriteRule ^api/([0-9]+\.[0-9]+\.[0-9]+)/skg-if-openapi\.yaml$ https://skg-if.github.io/api/openapi/ver/$1/skg-if-openapi\.yaml [R=303,L]
+RewriteRule ^api/skg-if-openapi\.yaml$ https://skg-if.github.io/api/openapi/ver/current/skg-if-openapi\.yaml [R=303,L]
+### SKG-IF API [end]
```
davidlehn, line 98: The catch all rule above will apply before these rules.  They should be moved before the above default rule.

PR #4965, opened 2025-05-20
Add redirect for soil-dsi-bridge ontology

GuaPi-D:
This pull request adds a redirect for https://w3id.org/projects/soil-dsi-bridge
pointing to https://guapi-d.github.io/soil-dsi-bridge/soil-dsi-bridge.ttl

### DIFF
--- a/projects/soil-dsi-bridge/.htaccess
+++ b/projects/soil-dsi-bridge/.htaccess
@@ -1,2 +1,0 @@
-RewriteEngine On
-RewriteRule ^$ https://guapi-d.github.io/soil-dsi-bridge/soil-dsi-bridge.ttl [R=302,L]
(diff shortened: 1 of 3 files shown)


dgarijo:
Please add a readme file indicating who will maintain this w3id

davidlehn:
Why are you using the `/projects/` prefix?  Would you prefer a shorter URL like `https://w3id.org/soil-dsi-bridge`?  I'm not sure the generic `projects` name should be claimed and used as only a parent directory like this.

GuaPi-D:
Thank you for the feedback! I have removed the `/projects/soil-dsi-bridge` directory and added a new `/soil-dsi-bridge` directory as suggested. The `.htaccess` and `README.md` files have been updated accordingly.

PR #4966, opened 2025-05-20
efro

UmairMarry:
This is a resubmission of a previous pull request. I've addressed the feedback by:
 Fixing the .htaccess file syntax
 
 Thanks

### DIFF
--- a/efro/.htaccess
+++ b/efro/.htaccess
@@ -1,5 +1,3 @@
-# Maintainer: umairarshad (https://github.com/umairarshad)
-
 # Optional settings – may be ignored or disallowed on some servers
 Options +FollowSymLinks
 Options -MultiViews
--- a/efro/.htaccess
+++ b/efro/.htaccess
@@ -1,16 +1,29 @@
-# Optional settings – may be ignored or disallowed on some servers
+# Options to follow symbolic links
 Options +FollowSymLinks
+
+# Turn off MultiViews
 Options -MultiViews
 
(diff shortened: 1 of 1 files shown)


efro/.htaccess
```diff
@@ -7,20 +7,23 @@ Options -MultiViews
...
-# Rewrite rule to serve HTML content from the vocabulary URI
-RewriteCond %{HTTP_ACCEPT} !application/rdf+xml.*(text/html|application/xhtml+xml)
-RewriteCond %{HTTP_ACCEPT} text/html [OR]
-RewriteCond %{HTTP_ACCEPT} application/xhtml+xml [OR]
-RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
-RewriteRule ^efro$ https://rgu-computing.github.io/EFRO/EFRO/docs/index-en.html [R=303,L]
-
-# Default response - owl rdf
-RewriteRule ^efro$ https://rgu-computing.github.io/EFRO/EFRO/EFRO.rdf [R=303,NE,L]
+<IfModule mod_rewrite.c>
```
davidlehn, line 12: Some comments:
- The IfModule isn't needed on this site.  mod_rewrite will always be on.
- The RewriteBase is only used for local files, and this service only uses redirects.
- This is mapping URLs like https://w3id.org/efro/efro.  Rules are relative to the current directory.  Did you mean to have the duplication?  I'm guessing so since the target also does that.

PR #4994, opened 2025-05-26
Update .htaccess of dfgfo

SArndt-TIB:
@doigl could you review?


doigl (approved):
Everything seems to work as expected. I just would vote for removal of the lines 20-22, because they are never used. But they do also not do any harm.

doigl (approved):

SArndt-TIB:
@doigl I deleted the lines. Thanks for review.
Sorry, perma-id, for not marking this Pull request as a draft!
We are ready for hand-over now. Thanks a lot!

PR #5007, opened 2025-06-03
updated ontology name

ritamargherita:
Sorry I have been requesting to update several times, just have been unsure about the best naming (hopefully this is last...)


dgarijo:
Please don't add capital letters in the w3id, as it has led to errors in the past.

PR #5015, opened 2025-06-05
Fix default RewriteRule for `/cff/`

sdruskat:
Thanks for merging #5014 so quickly!

Unfortunately, I failed to commit a bug fix in the default `RewriteRule`, which this PR now fixes.
Tested the complete `.htaccess` file in an online tester as well this time.

Sorry for the extra work, and thanks for merging!


sdruskat:
Thanks @dgarijo 👋!

PR #5025, opened 2025-06-06
Built out and tested NLM Classification redirects.

Superraptor:
Added a redirect mechanism for the National Library of Medicine (NLM) Classification. It uses an odd construction that is difficult to reference. For example, a code like "QS 642" has the equivalent link "https://classification.nlm.nih.gov/schedules/QS#QS 642", with the "QS" portion duplicated.

### DIFF
--- a/NLM/.htaccess
+++ b/NLM/.htaccess
@@ -1,0 +1,21 @@
+# Example
+#
+# https://w3id.org/NLM/QS 642 redirects to https://classification.nlm.nih.gov/schedules/QS#QS 642 (https://classification.nlm.nih.gov/schedules/QS#QS%20642)
+#
+# ## Contact
+# This space is administered by:
+#
+# Clair Kronk
+# Clair.Kronk@mountsinai.org
+# GitHub username: Superraptor
+
+# Examples tested using:
+# https://htaccess.madewithlove.com/
+# (6 June 2025)
+
+RewriteEngine On
+RewriteBase /
+
+# Redirect URLs.
+RewriteRule ^([A-Z]{1,2})\ ([\d\.A-Z]+)$ https://classification.nlm.nih.gov/schedules/$1#$1%20$2 [R=301,NE,L]
+RewriteRule ^([A-Z]{1,2})%20([\d\.A-Z]+)$ https://classification.nlm.nih.gov/schedules/$1#$1%20$2 [R=301,NE,L]
--- a/NLM/.htaccess
+++ b/NLM/.htaccess
@@ -1,6 +1,6 @@
 # Example
 #
-# https://w3id.org/NLM/QS 642 redirects to https://classification.nlm.nih.gov/schedules/QS#QS 642 (https://classification.nlm.nih.gov/schedules/QS#QS%20642)
+# https://w3id.org/NLM/QS%20642 redirects to https://classification.nlm.nih.gov/schedules/QS#QS%20642
 #
 # ## Contact
 # This space is administered by:
@@ -18,4 +18,3 @@ RewriteBase /
 
 # Redirect URLs.
 RewriteRule ^([A-Z]{1,2})\ ([\d\.A-Z]+)$ https://classification.nlm.nih.gov/schedules/$1#$1%20$2 [R=301,NE,L]
-RewriteRule ^([A-Z]{1,2})%20([\d\.A-Z]+)$ https://classification.nlm.nih.gov/schedules/$1#$1%20$2 [R=301,NE,L]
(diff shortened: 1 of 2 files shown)


NLM/.htaccess
```diff
@@ -0,0 +1,21 @@
+# Example
+#
+# https://w3id.org/NLM/QS 642 redirects to https://classification.nlm.nih.gov/schedules/QS#QS 642 (https://classification.nlm.nih.gov/schedules/QS#QS%20642)
```
davidlehn: Might be better to write that w3id URL with an encoded space?  A literal space in URLs is invalid.  Browsers and other tools may escape it for you, but servers probably wouldn't accept it.  (Though you may need to write more raw HTTP to even get that low level.)

NLM/.htaccess
```diff
@@ -0,0 +1,21 @@
...
+# Examples tested using:
+# https://htaccess.madewithlove.com/
+# (6 June 2025)
+
+RewriteEngine On
+RewriteBase /
+
+# Redirect URLs.
+RewriteRule ^([A-Z]{1,2})\ ([\d\.A-Z]+)$ https://classification.nlm.nih.gov/schedules/$1#$1%20$2 [R=301,NE,L]
+RewriteRule ^([A-Z]{1,2})%20([\d\.A-Z]+)$ https://classification.nlm.nih.gov/schedules/$1#$1%20$2 [R=301,NE,L]
```
davidlehn: The pattern is checking against the decoded URL, so the "%20" variation won't be used normally.  I assume you're not going to send a doubly encoded " " -> "%20" -> %2520"?  Could likely remove the line.

NLM/.htaccess
Superraptor: Added the encoded space!

NLM/.htaccess
Superraptor: Remove the doubly encoded line!

Superraptor:
@davidlehn despite testing going fine at https://htaccess.madewithlove.com/, I'm getting some weird behavior in the implementation. In theory "https://w3id.org/NLM/QS%20642" should redirect to "https://classification.nlm.nih.gov/schedules/QS#QS%20642". Instead, it's redirecting to "https://classification.nlm.nih.gov/schedules/QS#QS0642", inserting an extra zero. I'm genuinely not sure how to fix this behavior, and any advice is much, much appreciated!

davidlehn:
Escaping is hard.  I think the easiest approach is to escape the target `%` with `\%`.  Though there may be other solutions by putting quotes around parts here, and/or doing other escaping, and/or using the `B` flag somehow.  This worked locally:
```diff
-RewriteRule ^([A-Z]{1,2})\ ([\d\.A-Z]+)$ https://classification.nlm.nih.gov/schedules/$1#$1%20$2 [R=301,NE,L]
+RewriteRule ^([A-Z]{1,2})\ ([\d\.A-Z]+)$ https://classification.nlm.nih.gov/schedules/$1#$1\%20$2 [R=301,NE,L]
```

Superraptor:
@davidlehn It's strange. The `\%` solution did not work when attempting tests via https://htaccess.madewithlove.com/, but I went ahead and updated the PR because this might be one of those special cases that only really functions in deployment scenarios. Fingers crossed!

davidlehn:
Need a new PR since this was already merged.

That commit removing `NE` results in this:
```
$ curl -sI http://w3id.localhost/NLM/QS%20642 | grep -i ^location
Location: https://classification.nlm.nih.gov/schedules/QS%23QS0642
```
The `\%` suggestion resulted in:
```
$ curl -sI http://w3id.localhost/NLM/QS%20642 | grep -i ^location
Location: https://classification.nlm.nih.gov/schedules/QS#QS%20642
```
There's probably multiple escape issues with the config file, the URLs themselves, the rewrite engine, that htaccess site input, etc.  This is pushing the limits of testing other than with apache itself.

Could also match full pattern and the start as subpattern, quote to avoid the pattern space escape, adjust rewrites, and use magic B and BNP params.  This lets it escape the space for you.  These are all becoming quite difficult to understand!  I'm not sure what would work best for other expected inputs.
```
RewriteRule "^(([A-Z]{1,2}) [\d\.A-Z]+)$" https://classification.nlm.nih.gov/schedules/$2#$1 [R=301,NE,B,BNP,L]
```

Superraptor:
Added a new PR here: https://github.com/perma-id/w3id.org/pull/5032. I think the back-escape should work based on your testing and a parallel suggestion per StackOverflow here: https://stackoverflow.com/questions/79661185/url-rewrite-with-an-unencoded-hash-character-and-an-encoded-space-20/79661204#79661204. Thank you so so much!

PR #5028, opened 2025-06-09
Merge DQA Project

OumaimaAMAL:


dgarijo:
@OumaimaAMAL can you please:
- add your github id to the readme?
- change the repo to be `dqa` instead of `DQA`? Capitalization caused issues in the past with some repos

davidlehn:
> * change the repo to be `dqa` instead of `DQA`? Capitalization caused issues in the past with some repos

@dgarijo We need a FAQ or similar for this.  I haven't been enforcing or suggesting it.  I don't recall the exact reasoning for this.  I think it's because some tools are broken?  One real issue here is making sure we don't have multiple dirs that only differ in case.  But in general, specs and the rest of the web are fine with any case, seems that should be the situation here too.

OumaimaAMAL (commented):
Update README.md by adding my GitHub id

DQA/README.md
```diff
@@ -8,6 +8,6 @@ It introduces structured evaluation processes, traceability of applied methods,
 Ontology documentation redirection: https://w3id.org/DQA -> https://www.irit.fr/recherches/MELODI/ontologies/DQA/
 
 ## Maintainers
-* Oumaima AMAL (oumaima.amal@irit.fr) 
+* Oumaima AMAL (oumaima.amal@irit.fr) (GitHub ID : 117732363)
```
TallTed:
```suggestion
* Oumaima AMAL (oumaima.amal@irit.fr) (GitHub ID: [OumaimaAMAL](https://github.com/OumaimaAMAL))
```

PR #5032, opened 2025-06-10
Modifying /NLM/ rewrite code based on testing by @davidlehn in PR #5025

Superraptor:
Thanks so much to @davidlehn for suggesting an escape character, which I modified in this new `.htaccess` file. This was also the suggestion made in a parallel StackOverflow request (https://stackoverflow.com/questions/79661185/url-rewrite-with-an-unencoded-hash-character-and-an-encoded-space-20/79661204#79661204). Fingers crossed it works!

### DIFF
--- a/NLM/.htaccess
+++ b/NLM/.htaccess
@@ -17,4 +17,4 @@ RewriteEngine On
 RewriteBase /
 
 # Redirect URLs.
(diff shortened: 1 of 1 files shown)


Superraptor:
@davidlehn confirming it works! On this page: https://lgbtdb.wikibase.cloud/wiki/Item:Q301, there is a link to https://w3id.org/NLM/WM%20274, which resolves to correctly to https://classification.nlm.nih.gov/schedules/WM#WM%20274 !

PR #4800, opened 2025-03-18
Update for final VC 2.0 URLs.

davidlehn:
> [!IMPORTANT]
> Wait to merge until VC side is ready.

- See https://github.com/w3c/verifiable-credentials/issues/61.
- https://w3id.org/security/data-integrity/v2
  - https://www.w3.org/2025/credentials/vcdi/context/v2.jsonld
- https://w3id.org/security# (with content negotiation)
  - https://www.w3.org/2025/credentials/vcdi/vocab/v2/vocabulary.html
  - https://www.w3.org/2025/credentials/vcdi/vocab/v2/vocabulary.jsonld
  - https://www.w3.org/2025/credentials/vcdi/vocab/v2/vocabulary.ttl
  - https://www.w3.org/2025/credentials/vcdi/vocab/v2/vocabulary.svg
- https://w3id.org/security/multikey/v1
  - https://www.w3.org/2025/credentials/vcdi/multikey/context/v1.jsonld
- https://w3id.org/security/jwk/v1
  - https://www.w3.org/2025/credentials/vcdi/jwk/context/v1.jsonld

@iherman please review.

PR #5036, opened 2025-06-13
feat: add namespace redirection for ODRL Formal Semantics Ontology definition

YassirSellami:
Add namespace `odrl-fs` to redirect to the ontology definition of the ODRL Formal Semantics. 

Relates to this Pull Request in the ODRL repository: https://github.com/w3c/odrl/pull/130

### DIFF
--- a/odrl-fs/.htaccess
+++ b/odrl-fs/.htaccess
@@ -1,0 +1,29 @@
+Header set Access-Control-Allow-Origin *
+
+# Directive to ensure *.rdf files served as appropriate content type,
+# if not present in main apache config
+AddType application/rdf+xml .rdf
+AddType application/rdf+xml .owl
+AddType text/turtle .ttl
+AddType text/n3 .ttl
+AddType application/n-triples .nt
+AddType application/ld+json .json
+
+Options +FollowSymLinks
+
+RewriteEngine on
+
+# Redirect to Shacl shapes
+RewriteCond %{HTTP_ACCEPT} text/turtle
+RewriteRule ^$ https://w3c.github.io/odrl/formal-semantics/ontology/shacl.ttl [R=303,L]
+
+# Redirect to JSON-LD context
+RewriteCond %{HTTP_ACCEPT} application/ld\+json
+RewriteRule ^$ https://w3c.github.io/odrl/formal-semantics/ontology/json-ld_context.json [R=303,L]
+
+# Redirect to OWL ontology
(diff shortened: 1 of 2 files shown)


davidlehn:
May I ask why you are using w3id for this instead of w3c servers?  This looks like an ODRL group project?

Based on experience in the VC, DI, and DID groups, there is pushback for using w3id for id URLs (and JSON-LD context URLs).  The reasonable thinking is it would be preferred for W3C, rather than a 3rd party, to have control over ids and redirects that w3c specs use.  There are nuances and historical reasons to break from that, but for newly minted ids and redirects, I think it makes sense.

I'm not sure the details of the w3c hosting, but it can be setup to do a similar thing to what you have here.  Maybe even better if the mod_negotiation and proxy parts work in your use case.  For example, here's some discussion related to VC and DI redirects recently setup:
https://github.com/w3c/verifiable-credentials/issues/61
https://github.com/perma-id/w3id.org/pull/4800

YassirSellami:
> May I ask why you are using w3id for this instead of w3c servers? This looks like an ODRL group project?
> 
> Based on experience in the VC, DI, and DID groups, there is pushback for using w3id for id URLs (and JSON-LD context URLs). The reasonable thinking is it would be preferred for W3C, rather than a 3rd party, to have control over ids and redirects that w3c specs use. There are nuances and historical reasons to break from that, but for newly minted ids and redirects, I think it makes sense.
> 
> I'm not sure the details of the w3c hosting, but it can be setup to do a similar thing to what you have here. Maybe even better if the mod_negotiation and proxy parts work in your use case. For example, here's some discussion related to VC and DI redirects recently setup: [w3c/verifiable-credentials#61](https://github.com/w3c/verifiable-credentials/issues/61) #4800

Hello @davidlehn, that's a very valid question. And the answer would be that the current ODRL working drafts unlike VC/DID are not currently under a W3C Working Group. It's rather a W3C Community Group which is a completely different governance as far as I am aware.

I am not aware of the inner working of W3C hosting some there might some nuance I am missing regarding this.

Maybe you have a clearer response @vroddon ?

YassirSellami:
> Ok, we can merge. Did you want to wait until the related PR is merged so the targets exist?

I think it should be good, as most likely it would be content that might change and not the file names.

davidlehn:
> I think it should be good, as most likely it would be content that might change and not the file names.

You say that now.  Tomorrow they will change. :-)   We do try to wait for targets to exist to avoid churn, but I'll merge this now.

Issue #4896, opened 2025-04-28
Enhance documentation: naming-policy


ACz-UniBi:
Based on the discussion at https://github.com/perma-id/w3id.org/pull/4894, I am wondering whether the [naming policy](https://w3id.org/#naming-policy) should be refined to discourage the use of uppercase letters in the project ID and the identifier itself.

dgarijo:
The issue we faced, if I remember correctly, was in the `un` repo: https://github.com/perma-id/w3id.org/commits/master/un.
I think there was a lowercase and uppercase problem which led to problems for new people doing PRs when their system conflated both of them. But I cannot pin the exact rename issue in the history of the repo.

I think we should not allow repos with the same name but different case sensitivity. E.g., if someone does `exAmple` and we already have `example`. A simple solution is to mandate non capital letters like @TallTed proposes.

Another is to have a simple action that will not allow new folders if a folder with that name already exists (but has different capitalization). This one is perhaps more flexible.

For the time being, I have been suggesting using non capital letters in new PRs

davidlehn:
- We're all in agreement on not allowing dirs with conflicting case.  And eventually we'll enforce that with checks.
- Are there other arguments in favor of enforcing lower case?  I think that should be up to the authors.
- There was some issue with tools that normalized any paths to lower case (which is an absurd thing to do) and it was a problem that w3id is not setup to be case insensitive.  But I think since that's only happened once during the lifetime of this service, and it's against standards, it's not our problem.

PR #5053, opened 2025-06-19
Add MemOnto

SolineFelice:


dgarijo:
@SolineFelice can you please add the w3id using non-capital letters? Like memonto.
This lead to issues in the past. Thanks in advance?

davidlehn:
@dgarijo Can you please list those issues in https://github.com/perma-id/w3id.org/issues/4896 while we figure out details of a policy.  It's not clear to me when or if case ever needs to be enforced.

SolineFelice:
@dgarijo Yes of course, thank you for noticing.

dgarijo:
@davidlehn I added some explanations to the issue based on what I can remember.

PR #5059, opened 2025-06-20
simplied regex, added rules to fix redirect not working

CodingCerberus:


ancmed/.htaccess
```diff
@@ -1,8 +1,8 @@
 RewriteEngine on
+RewriteBase /
```
davidlehn, line 2: This is only used for local paths.  Not needed for redirects.

CodingCerberus: thanks David, just struggling why w3id.org/ancmed/311 doesn't redirect to https://ancmed.ulb.be/details.php?row=311

when I test it on https://htaccess.madewithlove.com/ it works fine with my rules, and the previous ones I had, can you help me understand what I've done wrong?

ancmed/.htaccess
```diff
@@ -1,8 +1,8 @@
...
 
 # Urls ending in numbers redirect to the relevant php page for the entry
-RewriteRule ^(.*)\/([0-9]+)$ https://ancmed.ulb.be/details.php?row=$2 [R=302,L]
-RewriteRule ^(.*)\/([0-9]+)/$ https://ancmed.ulb.be/details.php?row=$2 [R=302,L]
+RewriteRule ^(.*/)?([0-9]+)/?$ https://ancmed.ulb.be/details.php?row=$2 [R=302,L]
```
davidlehn: Did you only want numbers?  Or any match plus trailing numbers with optional trailing `/`?  If only numbers, you can simplify:
```suggestion
RewriteRule ^([0-9]+)/?$ https://ancmed.ulb.be/details.php?row=$2 [R=302,L]
```

```
$ curl -sI http://w3id.localhost/ancmed/311 | grep -i location
Location: https://ancmed.ulb.be/details.php?row=311
```

CodingCerberus: For now, we'd just like any number after "w3id.org/ancmed/" to redirect to the corresponding row page, with an optional trailing / .

ancmed/.htaccess
```diff
@@ -1,8 +1,8 @@
 RewriteEngine on
+RewriteBase /
 
 # Urls ending in numbers redirect to the relevant php page for the entry
-RewriteRule ^(.*)\/([0-9]+)$ https://ancmed.ulb.be/details.php?row=$2 [R=302,L]
-RewriteRule ^(.*)\/([0-9]+)/$ https://ancmed.ulb.be/details.php?row=$2 [R=302,L]
+RewriteRule ^([0-9]+)/?$ https://ancmed.ulb.be/details.php?row=$2 [R=302,L]
```
davidlehn, line 5: Need to change that to `$1` after the removal of the first pattern capture.

davidlehn: Fixed in https://github.com/perma-id/w3id.org/pull/5067.

PR #5092, opened 2025-06-30
Moving URI from test to productions

riccardoAlbertoni:
Please 
@bfabio @mfortini @giorgialodi  
include this new change in the negotiations pertaining to the "Portal of Sources for the History of the Italian Republic" project.


bfabio:
lgtm

PR #5107, opened 2025-07-04
added squirrel resources from fuzzy-sl and ogham LOD

florianthiery:
new sub-folder "squirrels"


florianthiery:
I have done two commits to fix it would it work now @dgarijo ?

PR #5109, opened 2025-07-04
fix rewrite rule

florianthiery:
it should redirect to https://w3id.org/archlink/terms/<ANYTHING> and https://w3id.org/archlink/samian/<ANYTHING> using the sub-folder structure


florianthiery:
Does that work as we axpect, it is the same problem like in #5107  as we used the sub-folder structure. thx @florianthiery and @LasseMempel

PR #5117, opened 2025-07-07
Update XFSC redirect to point to Eclipse GitHub

bearn01d:
Awaiting approval from @clange.

### DIFF
--- a/xfsc/.htaccess
+++ b/xfsc/.htaccess
@@ -1,4 +1,4 @@
 Header set Access-Control-Allow-Origin *
 
 RewriteEngine On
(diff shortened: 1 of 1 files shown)


clange (approved):
Fine, thank you! I confirm that the old redirect is obsolete and the new one is correct.

clange:
> @clange The README.md in this dir lists a gitlab username. At some point can you add github clange to the readme so we can more easily check approvals are from the right account. Thanks.

Thanks @davidlehn for pointing out – I'm now addressing this in #5125 .

PR #5127, opened 2025-07-09
coso fixing rewrite rules in .htaccess

thahmann:


dgarijo:
Please have @kschweikert approve this PR. If you add yourself to the maintainer list with your GitHub id, I will be able to verify whether you are part of it in the future :)
Btw, `RewriteRule ^$ https://github.com/SAWGraph/contaminoso ` will not return the content in rdf/xml as you intend. You have to create a github page, or return the raw content.

thahmann:
@dgarijo: I added myself as an admin to the fork. Hope that makes verification easier now. 

Thanks for the suggestion on the .htaccess file. We did want one redirect (if the url is exactly w3id.org/coso with nothing afterwards) to the repo (rather than a specific file) to allow pointing to additional documentation and example data hosted in the repo.  But you're right that makes sense for html/text request not rdf/xml requests. I hope I fixed it now.

PR #5131, opened 2025-07-10
Update .htaccess (APTO)

Filipi-Soares:
I updated the script to support content negotiation using AgroPortal.


jonquet:
Will need to be corrected after https://github.com/agroportal/project-management/issues/800 is fixed

PR #5144, opened 2025-07-15
create redirect for mds

vandttran:
Requesting a new namespace, called mds.


vandttran:
maintainer's github accounts have been added

PR #5147, opened 2025-07-16
Creation of W3ID for COLREG Ontology

Sabbus:
I would like to have a W3 unique identifier for an ontology developed to model the Convention on the International Regulations for Preventing Collisions at Sea. The ontology is going to be published in the ontology showcase of the [FOIS 2025](https://www.dmi.unict.it/fois2025/) conference.


dgarijo:
@Sabbus can you please add your GitHub id next to your name in the readme file?

PR #5151, opened 2025-07-17
Add new w3id for dh.aks.ac.kr 

dongshins:
This pull request adds a new persistent identifier for https://w3id.org/dh.aks.ac.kr/.

# Redirect Information

The identifier redirects permanently (HTTP 301) to the address of the current DH host at the Academy of Korean Studies at:
→ https://dh.aks.ac.kr/

The `.htaccess` file uses Apache rewrite rules to handle:
- Base path redirection:
  → https://w3id.org/dh.aks.ac.kr/ → https://dh.aks.ac.kr/
- Subpath redirection:
  → https://w3id.org/dh.aks.ac.kr/something → https://dh.aks.ac.kr/something

# Maintainer Contact

This w3id is maintained by:

Dongshin SEO  
📧 oriental.neo@gmail.com  
🐙 GitHub: @dongshins

All files are placed under the `dh.aks.ac.kr/` directory, following w3id.org conventions.

### DIFF
--- a/dh.aks.ac.kr/.htaccess
+++ b/dh.aks.ac.kr/.htaccess
@@ -1,0 +1,14 @@
+# dh.aks.ac.kr
+#
+# https://w3id.org/dh.aks.ac.kr/ redirects to https://dh.aks.ac.kr/
+#
+# ## Contact
+# This space is administered by:
+#
+# Dong Shin SEO
+# oriental.neo@gmail.com
+# GitHub username: dongshins
+
+RewriteEngine on
+RewriteRule ^$ https://dh.aks.ac.kr/ [R=301,L]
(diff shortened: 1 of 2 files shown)


dongshins:
@davidlehn 
Thank you for your insight! In our case, the hostname dh.aks.ac.kr represents not just a server, but the identity of the ontology and related projects developed under this domain.
Given the limited resources to maintain the current server and the uncertainty about future hosting environments, we chose this w3id as a stable and meaningful reference layer.
Even if the physical server changes, we can simply update the redirect target while keeping the symbolic identity.
We understand the general practice of using project names, but in our scenario, the hostname itself is the most representative identifier.

PR #5161, opened 2025-07-24
added tklont for traditional knowledge label ontology

biktorrr:
Ontology that formalizes Traditional Knowledge Labels (TKLs) for which we would like to use w3ids


biktorrr:
done

PR #5166, opened 2025-07-28
Add redirect for /explainability/sense/result

allNormal:


Kat-rin-sc:
the added redirect is fine.

PR #5173, opened 2025-07-29
Added  (rdf/json) content negotiationto dfc taxonomies redirections

RaggedStaff:
Content negotiation enabled to allow serving of either rdf or json-ld content as appropriate, without diverging URI's for resources.


dgarijo:
Closing due to inactivity. Please reopen when the comment before is addressed.

PR #5175, opened 2025-07-31
Addition of URIs resolving to additional vocabulary schemes

angusnixon:
Adding URIs resolving to the Data Reduction Software and Analysis Scale vocabulary landing pages (note these are currently in RVA DEMO)


angusnixon:
no problems just added this

PR #5209, opened 2025-08-08
Add redirect for jelly-cli install script

Ostrzyciel:
Issue: https://github.com/Jelly-RDF/cli/issues/172

This adds a redirect for the [jelly-cli](https://github.com/Jelly-RDF/cli) installation script. I tested this on a local Apache instance, it works.

You can review the install script this points to here: https://github.com/Jelly-RDF/cli/blob/main/install.sh

### DIFF
--- a/jelly/.htaccess
+++ b/jelly/.htaccess
@@ -1,5 +1,10 @@
 RewriteEngine on
 
+AddType application/x-sh .sh
+
+# Install script for jelly-cli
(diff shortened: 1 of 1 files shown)


jelly/.htaccess
```diff
@@ -1,5 +1,10 @@
 RewriteEngine on
 
+AddType application/x-sh .sh
```
davidlehn, line 3: This doesn't do anything for redirects.  Only used when serving local files.

PR #5215, opened 2025-08-11
Enhance hmontology namespace to redirect individual ontology terms

shamilasudalshana:
## Summary
This PR updates the `.htaccess` for the `hmontology` namespace to ensure that individual ontology term IRIs (e.g., `https://w3id.org/hmontology/hasDominantGeologicalClass`) redirect correctly to their corresponding term pages in the current hosting location.

## Changes
- Added rewrite rules to handle subpaths and redirect them either to:
  - The matching term page (`/<term>/`) on the backend site, or
  - A section (`#<term>`) of the main ontology HTML page, depending on hosting structure.
- Preserved existing content negotiation and 303 redirects for the ontology root.

## Context
Previously, only the base namespace (`https://w3id.org/hmontology/`) redirected to the ontology’s main HTML page or RDF serializations.  
Requests to term IRIs did not resolve as intended, impacting linked data usability.

Maintainer: Shamila Herath (shamilasri40@gmail.com)

### DIFF
--- a/hmontology/.htaccess
+++ b/hmontology/.htaccess
@@ -1,16 +1,24 @@
 Options +FollowSymLinks
 RewriteEngine On
 
-# Redirect to JSON-LD if requested
+# ---------- Ontology root (content negotiation) ----------
+# JSON(-LD)
 RewriteCond %{HTTP_ACCEPT} application/ld\+json [OR]
 RewriteCond %{HTTP_ACCEPT} application/json
 RewriteRule ^$ https://hydro-knowledge-graph-f48785.pages.rwth-aachen.de/latest.jsonld [R=303,L]
 
-# Redirect to Turtle if requested
+# Turtle / N-Triples / N-Quads
 RewriteCond %{HTTP_ACCEPT} text/turtle [OR]
 RewriteCond %{HTTP_ACCEPT} application/n-triples [OR]
 RewriteCond %{HTTP_ACCEPT} application/n-quads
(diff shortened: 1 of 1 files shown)


hmontology/.htaccess
```diff
@@ -1,16 +1,24 @@
...
+# ---------- Ontology root (content negotiation) ----------
+# JSON(-LD)
 RewriteCond %{HTTP_ACCEPT} application/ld\+json [OR]
 RewriteCond %{HTTP_ACCEPT} application/json
 RewriteRule ^$ https://hydro-knowledge-graph-f48785.pages.rwth-aachen.de/latest.jsonld [R=303,L]
 
-# Redirect to Turtle if requested
+# Turtle / N-Triples / N-Quads
 RewriteCond %{HTTP_ACCEPT} text/turtle [OR]
 RewriteCond %{HTTP_ACCEPT} application/n-triples [OR]
```
dgarijo: Please note that n-triples != turtle. You would be returning the wrong content for this request

dgarijo:
Please see my comment @shamilasudalshana

hmontology/.htaccess
shamilasudalshana: Thanks for pointing that out. I’ve updated the .htaccess so that application/n-triples and application/n-quads are no longer matched with text/turtle.
Currently, we only publish JSON-LD and Turtle, so I’ve restricted the rules to those formats and kept HTML as the default fallback.
I’ll add rules for RDF/XML or other serializations when we start publishing them.

PR #5219, opened 2025-08-12
Add CANDIL and Mouseworld ontologies

idomingu:


candil/interface/.htaccess
```diff
@@ -0,0 +1,48 @@
...
+# if not present in main apache config
+AddType application/rdf+xml .rdf
+AddType application/rdf+xml .owl
+AddType text/turtle .ttl
+AddType application/n-triples .n3
+AddType application/ld+json .jsonld
+# Rewrite engine setup
+RewriteEngine On
+#Change the path to the folder here
+RewriteBase /
```
davidlehn: `RewriteBase` and `AddType` are only used when serving files.  Not used for this redirect-only service.  Can leave those here if you want, but they won't be used.  Same in other .htaccesses.  What would you prefer?

idomingu: I copied them from another project a long time ago. I will remove them based on your recommendation.

PR #5233, opened 2025-08-14
adding specification endpoint to inspec

lokal-profil:
adds support for [w3id.org/inspec/specification](http://w3id.org/inspec/specification)


lokal-profil:
https://diggsweden.github.io/interoperable-specifications/docs/ is now ready so moving this PR out of draft state

PR #5242, opened 2025-08-19
fix: remove # in redirects as that does not work

woutslabbinck:
fragments are only handled in browser, thus this was translated to its unicode variant `%23`, leading to 404s


dgarijo:
You may want to use github pages to serve the turtle files. Otherwise you serve the github HTML page of your resource

PR #5253, opened 2025-08-22
feat(factoryx): add initial factoryx namespace

arnoweiss:
## Background

Factory-X is a dataspace aimed at interoperable data exchange between organizations in the automation industry. Some aspects of the chosen technology stack require the maintenance of IDs redirecting to json-ld contexts and json schemas. This namespace will be used for precisely that.

## Future Work

- Add more contacts to distribute governance
- Move redirect targets away from (1) github raw and (2) the factoryx-edc repo to something not ratelimited.

## Notes

All preconditions met. PR is ready for review.

### DIFF
--- a/factoryx/.htaccess
+++ b/factoryx/.htaccess
@@ -1,0 +1,21 @@
+#
+# Factory-X Namespace Forwarding Rules
+# tested with https://htaccess.madewithlove.com/
+#
+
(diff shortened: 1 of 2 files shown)


arnoweiss:
is there anything that we can help the review process with?

PR #5255, opened 2025-08-23
Add testo redirects

micheldumontier:
This PR adds `testo/.htaccess` with redirects for the `testo` namespace.

### DIFF
--- a/testo/.htaccess
+++ b/testo/.htaccess
@@ -1,0 +1,48 @@
+Header set Access-Control-Allow-Origin *
+Options -MultiViews
+
+RewriteEngine on
+RewriteRule ^ - [E=ONTO:testo]
+RewriteRule ^ - [E=ONTO_IRI:https://w3id.org/%{ENV:ONTO}/]
+#RewriteRule ^ - [E=WEBSITE:index.html]
+RewriteRule ^ - [E=WEBSITE:https://micheldumontier.github.com/ontostart/]
+RewriteRule ^ - [E=DOCS:https://ontostart.github.io/docs/]
+
+RewriteRule ^ - [E=VERSIONS:versions]
+RewriteRule ^ - [E=LATEST:%{ENV:VERSIONS}/latest]
+#RewriteRule ^ - [E=FILE_IRI:http://localhost:8080/versions/]
+RewriteRule ^ - [E=FILE_IRI:https://raw.githubusercontent.com/micheldumontier/ontostart/refs/heads/main/%{ENV:VERSIONS}/]
+
+RewriteBase /
+
+# HTML documentation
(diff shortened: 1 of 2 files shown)


testo/.htaccess
```diff
@@ -0,0 +1,48 @@
...
+RewriteRule ^docs/(.*)$ %{ENV:DOCS}/$1 [R=303,L]
+RewriteRule ^ontology$ %{ENV:LATEST}/%{ENV:ONTO}.ttl [R=303,L]
+RewriteRule ^versions/(([0-9]+\.[0-9]+\.[0-9]+)|latest)$ %{ENV:VERSIONS}/$1/ [R=303,L,NE]
+
+# Redirects for RDF/XML files
+RewriteCond %{HTTP_ACCEPT} application/rdf\+xml [OR]
+RewriteCond %{HTTP_ACCEPT} application/xml
+RewriteCond %{HTTP_ACCEPT} application/owl\+xml
+RewriteRule ^$ %{ENV:LATEST}/%{ENV:ONTO}.owl [R=303,L]
+RewriteRule ^versions/(([0-9]+\.[0-9]+\.[0-9]+)|latest)$ %{ENV:VERSIONS}/$1/%{ENV:ONTO}.owl [R=303,L,NE]
```
davidlehn, line 30: Has this all been tested?  I don't think it will work as you expect here.  RewriteCond lines only apply to the next RewriteRule.  The versions match above would have caught this already.  I think the solution is cond repetition, but if you know of a shortcut, please share.  In any case, please test and confirm these rules do what you intend.

micheldumontier: i tested this on a local apache server

davidlehn:
```
$ curl -sI -H "Accept: application/xml" http://w3id.localhost/tmp1/versions/0.0.0 | grep Location
Location: http://w3id.localhost/versions/0.0.0/
```
- How did you test it?
- There are no RewritCond's on the first `versions` rule, so that will always be the one that is used regardless of accept headers.
- ENV:VERSIONS is not setup with a host, so it's returning a path relative to w3id root.

testo/README.md
```diff
@@ -0,0 +1,7 @@
+# TESTO Ontology
+[https://w3id.org/ontostart/](https://w3id.org/ontostart/) -> [https://micheldumontier.github.io/ontostart/index.html](https://micheldumontier.github.io/ontostart/index.html)
```
davidlehn, line 2: Looks like those paths should be updated to `testo`.

dgarijo:
ping @micheldumontier

micheldumontier:
i created a new pull request that should address all these issues.

PR #5268, opened 2025-08-28
Minor updates to README and new private domain

pmcb55:
Minor updates to README (showing codebase is private for now), and using new private domain.

### DIFF
--- a/lit/README.md
+++ b/lit/README.md
@@ -8,7 +8,7 @@ and more intuitive for developers.
 
 ### Authors/Maintainers: 
 
-- Pat McBennett (pat.mcbennett@gmail.com)
+- Pat McBennett (Email: pat.mcbennett@gmail.com, GitHub: https://github.com/pmcb55)
 
 > [!NOTE]
 > The LIT is currently private, but the intention is to make it freely available (probably under
--- a/lit/README.md
+++ b/lit/README.md
@@ -1,8 +1,16 @@
-# LIT: Linked data Integration Toolkit
-A collection of developer libraries (in Java and Javascript) that make working with Linked Data (i.e. RDF) easier and more intuitive for development teams.
+# Welcome to the LIT (the Linked Data Integration Toolkit)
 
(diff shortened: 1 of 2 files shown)


lit/README.md
```diff
@@ -1,8 +1,16 @@
...
+Finally - a truly Enterprise-grade, Production-ready developer toolkit for Linked Data!
 
-The LIT project is available online at https://github.com/pmcb55/LIT
+The LIT is a collection of Enterprise-grade, Production-ready, fully tested developer libraries
+(written in TypeScript) specifically designed to make working with Linked Data (i.e., RDF) easier
+and more intuitive for developers.
+
+### Authors/Maintainers: 
+
+- Pat McBennett (pat.mcbennett@gmail.com)
```
davidlehn: Please add your github username here too.

lit/README.md
```diff
@@ -1,8 +1,16 @@
...
+
+> [!NOTE]
+> The LIT is currently private, but the intention is to make it freely available (probably under
+MIT license, and probably on GitHub) once polished and considered feature complete enough for
+Enterprise development.
```
davidlehn, line 16: This service tries to point more at public resources but it's fine.

I would suggest making things public to start if you intend to at some point anyway.  Any type of project is difficult to get usable at all, much less to a state that is "polished and considered feature complete enough for Enterprise development."  Release early and often.

pmcb55: Thanks David, yeah I totally agree with you in principal, but so far what I've been playing around within this project (i.e., a Linked Data Integration Toolkit, the 'LIT') has gone through a lot of iteration, and so it seems making it public before it has 'settled down' might just cause unecessary confusion and noise.
(But I don't believe it's too far away from being settled now... (famous last words, of course!))

lit/README.md
pmcb55: Done: 6abf2c48baa48d8ffa3d86a613a434310b292406

PR #5271, opened 2025-08-29
added application ontologies for logistics and heat treatment

ThHanke:


joergwa:
I approve it.  Thank you!!

PR #5282, opened 2025-09-05
Improve term resolution with aliases and upper ontology imports

Angatar:


Angatar:
Hi, thanks for pointing this out. I had initially tested the redirections locally at root without the evorao/ subfolder, so you’re right,  there was an issue. I’ve now fixed it by adding `evorao/` to the alias redirects, and it works as expected.

For example:
`https://w3id.org/evorao/DOI` → `https://w3id.org/evorao/Doi` → `https://www.ebi.ac.uk/ols4/ontologies/evorao/classes/https%253A%252F%252Fw3id.org%252Fevorao%252FDoi`

The final OLS URL doesn’t return a 200 yet, but that’s only because it needs a few days to propagate on OLS.

PR #5287, opened 2025-09-05
accommodate optional ending slash

micheldumontier:
also changed path for ontospy to be generic


dgarijo:
@davidlehn can you please share this info?

micheldumontier:
no, we're good. thanks

PR #5302, opened 2025-09-12
Add w3id.org/megabanana

sravan953:


megabanana/.htaccess
```diff
@@ -0,0 +1,13 @@
+# Example
+#
+# https://w3id.org/examples/simple/ redirects to https://example.com/
```
davidlehn, line 3: Looks like something from example template.

sravan953: Yes, I left the comment in place.

PR #5317, opened 2025-09-16
Add ontostart redirects

micheldumontier:
This PR adds `ontostart/.htaccess` with redirects for the `ontostart` namespace.


micheldumontier:
err... fixed obvious problem.

PR #5327, opened 2025-09-18
Update redirect URL in .htaccess

sravan953:
Changed the RewriteRule target from the us-west1 to the us-central1 Google Cloud Run region for megabanana. This updates the redirect destination for all incoming requests.

### DIFF
--- a/megabanana/.htaccess
+++ b/megabanana/.htaccess
@@ -1,21 +1,12 @@
 # Example
-
 #
-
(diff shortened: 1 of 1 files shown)


megabanana/.htaccess
```diff
@@ -1,13 +1,22 @@
 # Example
+
```
davidlehn: Are the newlines on purpose?

sravan953: Nope, did not see this. Fixed!

PR #5365, opened 2025-10-06
Update .htaccess for FUEL ontology

agbeltran:
Adding content negotiation for different formats


agbeltran:
thanks @dgarijo, I had another pass to the .htaccess file

PR #5367, opened 2025-10-06
LDES: conneg on vocabulary and adding server primer

pietercolpaert:
This prepares LDES for the 1.0 release in November: it adds a separate doc for the vocabulary using conneg, and adds a permalink for the server primer

### DIFF
--- a/ldes/.htaccess
+++ b/ldes/.htaccess
@@ -1,9 +1,19 @@
+AddType text/turtle .ttl
+AddType application/ld+json .jsonld
+
 Header set Access-Control-Allow-Origin *
 Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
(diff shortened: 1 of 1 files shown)


ldes/.htaccess
```diff
@@ -1,9 +1,19 @@
+AddType text/turtle .ttl
+AddType application/ld+json .jsonld
+
```
davidlehn, line 3: These are only used for local files.  They don't do anything for redirects.

pietercolpaert: I was already wondering indeed: was following the example at https://github.com/perma-id/w3id.org/tree/master/examples#example-2-supporting-multiple-media-types-mime-types

davidlehn: Hmm.  Not sure what that was there.  It does no harm to have `AddType`, but since it is to determine media type of a file, and this service only does redirects, I'm not sure why anyone uses those other than cargo culting.  And apparently, due to reading our docs!

PR #5375, opened 2025-10-08
Handle IDs changed in the Clavis Aethiopica

smaugustine:
This is to handle specific redirects for doublets which were removed from the Clavis Aethiopica (see https://github.com/BetaMasaheft/Documentation/issues/3046).

It also deals with the issue raised in https://github.com/BetaMasaheft/Documentation/issues/2653.

### DIFF
--- a/malke/.htaccess
+++ b/malke/.htaccess
@@ -2,4 +2,11 @@ Header set Access-Control-Allow-Origin *
 Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
 Options +FollowSymLinks
 RewriteEngine on
+# Moved IDs
+RewriteRule 0285 6392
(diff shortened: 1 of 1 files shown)


smaugustine:
@dgarijo To be honest, just with https://htaccess.madewithlove.com (which seems successful), but I will be happy for any advice because I only rarely work with `.htaccess` files and it's been some time.

The goal is to rewrite the specific ID numbers first before applying the general rule.

smaugustine:
E.g.:

`https://w3id.org/malke/0285/` =>
`https://w3id.org/malke/6392/` =>
`https://malkeagubae.com/works/6392/`

davidlehn:
@dgarijo This is to remap ids here before doing the redirect.
@smaugustine Can you do this remapping on your server?  Seems like it would be better to track and manage id mappings on your side rather than here.  Especially if those mappings change often.

PR #5381, opened 2025-10-13
Add redirect for Religious Education Vocabulary (reliedu) – SkoHub Vocabs Integration

joerglohrer:
AMB-konformes, kontrolliertes Vokabular für religionspädagogische Bildungsressourcen.  
Automatische Visualisierung und Content Negotiation via SkoHub Vocabs.


reliedu/.htaccess
```diff
@@ -0,0 +1,133 @@
...
+# Documentation: https://github.com/rpi-virtuell/reliedu-vocab
+
+# Turn off MultiViews
+Options -MultiViews
+
+# Directive to ensure *.rdf files served as appropriate content type
+AddType application/rdf+xml .rdf
+AddType text/turtle .ttl
+AddType application/ld+json .json
+AddType application/ld+json .jsonld
```
davidlehn: These are used for local files, and not needed in a redirect service like this.

reliedu/.htaccess
```diff
@@ -0,0 +1,133 @@
...
+
+# Rewrite engine setup
+RewriteEngine On
+RewriteBase /reliedu
```
davidlehn: This won't be used for redirects.

reliedu/.htaccess
```diff
@@ -0,0 +1,133 @@
...
+RewriteRule ^location/kirche/?$ https://rpi-virtuell.github.io/reliedu-vocab/location/kirche.html [R=303,L]
+RewriteRule ^location/gemeinde/?$ https://rpi-virtuell.github.io/reliedu-vocab/location/gemeinde.html [R=303,L]
+RewriteRule ^location/kloster/?$ https://rpi-virtuell.github.io/reliedu-vocab/location/kloster.html [R=303,L]
+RewriteRule ^location/friedhof/?$ https://rpi-virtuell.github.io/reliedu-vocab/location/friedhof.html [R=303,L]
+RewriteRule ^location/museum/?$ https://rpi-virtuell.github.io/reliedu-vocab/location/museum.html [R=303,L]
+RewriteRule ^location/natur/?$ https://rpi-virtuell.github.io/reliedu-vocab/location/natur.html [R=303,L]
+RewriteRule ^location/digital/?$ https://rpi-virtuell.github.io/reliedu-vocab/location/digital.html [R=303,L]
+
+# Catch-all for any other concept URIs (fallback to SkoHub search)
+RewriteRule ^(.+)/?$ https://rpi-virtuell.github.io/reliedu-vocab/?search=$1 [R=303,L]
```
davidlehn: Rules are processed in order.  I think this will catch everything before the rules below.  Maybe it should be lower?

reliedu/README.md
davidlehn, line 1: - This readme is too verbose.  Please host your documentation on your own site.  It's better to keep this very simple and refer to your docs elsewhere that are easier for you to maintain.
- Please add your github username to the maintainer line.

reliedu/.htaccess
joerglohrer: Agreed! I've removed the `RewriteBase` directive as it's not needed for redirects. The simplified version now just uses a straightforward rewrite rule without unnecessary directives.

reliedu/.htaccess
joerglohrer: You're correct - the catch-all rule would have intercepted everything. I've completely restructured the `.htaccess`  Instead of explicit rules for each concept, I now use a single catch-all redirect to https://rpi-virtuell.github.io/reliedu-vocab/$1. SkoHub Vocabs on GitHub Pages handles all the content negotiation automatically, so we don't need individual rewrite rules for each concept. This is much cleaner and follows best practices.

reliedu/.htaccess
joerglohrer: You're absolutely right! I've removed all AddType directives. They're indeed only needed for serving local files, not for a redirect service. The `.htaccess`is now simplified to just the essential redirect rule, following the other templates.

reliedu/README.md
joerglohrer: @davidlehn Absolutely right! I've completely restructured the README following the openeduhub template. It's now reduced from 268 lines to just 13 lines with:

- Brief one-line description
- Links to external documentation
- Links to vocabulary (https://rpi-virtuell.github.io/reliedu-vocab)
- Contacts section with my GitHub username (@joerglohrer)
All verbose documentation, examples, usage guides, and detailed descriptions are now hosted in the external repository where they're much easier to maintain and update. The w3id.org README is now minimal and clean, just as it should be.

dgarijo:
We are seeing an increase in AI generated PRs. @davidlehn I am inclined to open an action to do some quick checks like the maintainer being present in the readme. For the rest, maybe we should involve the owners a bit more. I will not be playing ping pong with AI bots if owners have not bothered to check their PRs work as they should.

PR #5383, opened 2025-10-15
Request a New Identifier "https://w3id.org/chuo-dl/"

tsunekawa:


chuo-dl/README.md
davidlehn, line 1: Rename to `README.md`.

tsunekawa: done.

PR #5414, opened 2025-10-22
Add pull request template.

davidlehn:
@dgarijo Something like this has been needed forever.  It's a small step towards adding automated checks.  Maybe it will help with our constant need to ask for maintainer github ids.  I'm going to merge immediately and see how it goes, but please suggest changes.


dgarijo:
Awesome!
I would add a recommendation on new ids: [ ] (recommended) my w3id has non capital letters

dgarijo:
That is why I suggested it as a recommendation, not an enforcement. But I am fine either way.

davidlehn:
Future vision is less of this sort of checklist text and tests would warn or fail when issues are found.  For new files it could check case and show "/!\ Using upper case path name characters.", and link to some docs to talk about it.  Mostly this template text was to avoid the immediate problems that are happening daily!

PR #5444, opened 2025-11-03
dpv: remove trailing slash (attempt 2)

coolharsh55:
The previous PR #5442 did not work. It redirected w3id.org/dpv/examples/ to https://w3id.org/var/www/w3id.org/dpv/examples instead of w3id.org/dpv/examples The fix (hopefully) is to simply rewrite without the trailing slash in this form:
```
# Remove trailing slash if exists
RewriteRule ^(.*)/$ /$1 [R=302,L]
```

### DIFF
--- a/dpv/.htaccess
+++ b/dpv/.htaccess
@@ -24,7 +24,7 @@ SetEnvIf Request_URI ^.*$ DEV=https://dev.dpvcg.org
 SetEnvIf Request_URI ^.*$ LATEST=2.2
 
 # Remove trailing slash if exists
-RewriteRule ^(.*)/$ /$1 [R=302,L]
+RewriteRule ^(.*)/$ /dpv/$1 [R=302,L]
 
 ##### VOCAB: DPV #####
 


dpv/.htaccess
```diff
@@ -19,14 +19,13 @@ Header set Access-Control-Allow-Origin *
...
-# Remove trailing slash if exists
-RewriteCond %{REQUEST_URI} /$
-RewriteRule ^(.*)/$ $1 [L,R=301]
-
 SetEnvIf Request_URI ^.*$ BASE=https://w3c.github.io/dpv
 SetEnvIf Request_URI ^.*$ DEV=https://dev.dpvcg.org
 SetEnvIf Request_URI ^.*$ LATEST=2.2
 
+# Remove trailing slash if exists
+RewriteRule ^(.*)/$ /$1 [R=302,L]
```
davidlehn: These things are tricky.  I think you might want this:
```suggestion
RewriteRule ^(.*)/$ /dpv/$1 [R=302,L]
```
But you probably need to do some testing to check.  Without your prefix in there, it would redirect to your sub-path relative to the root.

Note that the redirect R is going to force an extra redirect on the client.  That adds more requests and latency for everyone.  You could remove the R param and have it do an internal redirect.  (Please test if you do.)  There are tradeoffs to how you handle this.  Arguably, it's best to only support URLs you intend and let everything else fail.  People will update their URLs that way rather than continue to use the wrong ones.

Another approach is also to end most of your rules with an optional slash like `^...../?$`.  A lot of rules to update for that, and again, having one correct URL might be better.

coolharsh55: Thanks David. I think `/dpv/$1` does make sense. I think the trailing slash is common enough for people so we don't want to ignore it, but also letting the browser know this isn't a correct url seems sensible?

davidlehn: Sure, that works fine.  Using 302 vs 301 for such things is might help depending on the tools used and the use case.

PR #5603, opened 2025-12-26
Added CO-META redirections to handle individuals and versions

icasrod:
## Brief Description
We have added versioning to the original repository. Hence, we have updated the htaccess rules to allow for versioned IRIs

## General Checklist
- [ x] Changes have been tested.
- [ x] The number of commits is minimal.
- [ x] Commits only include redirects and basic information. Serving content and full documentation is not supported on this service.


icasrod:
Thank you very much

PR #5677, opened 2026-01-23
Changes file ending to .ttl to correctly reflect turtle syntax

Miguel2617:
This update adjusts the redirection rules for one ontology to reflect a format change from .owl to .ttl., since the ontology has always been written in Turtle syntax. There are no changes to the ontology content itself — only the file extension was corrected to accurately reflect the format being used. The redirect rules only switch from .owl to .ttl starting at the corresponding new version.

### DIFF
--- a/hsu-aut/IEEE1872-2/.htaccess
+++ b/hsu-aut/IEEE1872-2/.htaccess
@@ -1,16 +1,22 @@
 RewriteEngine on
 
-# Redirect ontology version IRI to the given version
+# Redirect ontology version IRI to the given version (until v1.0.0 to .owl)
 RewriteCond %{HTTP_ACCEPT} application/rdf\+xml [OR]
 RewriteCond %{HTTP_ACCEPT} text/turtle [OR]
 RewriteCond %{HTTP_ACCEPT} application/x-turtle
-RewriteRule ^(\d\.\d\.\d)/?$ https://raw.githubusercontent.com/hsu-aut/IndustrialStandard-ODP-IEEE1872-2/v$1/AuR_IEEE1872-2.owl [R=303,NC,L]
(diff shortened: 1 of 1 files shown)


hsu-aut/IEEE1872-2/.htaccess
```diff
@@ -1,16 +1,22 @@
...
-# Redirect ontology version IRI to the given version
+# Redirect ontology version IRI to the given version (until v1.0.0 to .owl)
 RewriteCond %{HTTP_ACCEPT} application/rdf\+xml [OR]
 RewriteCond %{HTTP_ACCEPT} text/turtle [OR]
 RewriteCond %{HTTP_ACCEPT} application/x-turtle
-RewriteRule ^(\d\.\d\.\d)/?$ https://raw.githubusercontent.com/hsu-aut/IndustrialStandard-ODP-IEEE1872-2/v$1/AuR_IEEE1872-2.owl [R=303,NC,L]
+RewriteRule ^(1\.0\.0)/?$ https://raw.githubusercontent.com/hsu-aut/IndustrialStandard-ODP-IEEE1872-2/v$1/AuR_IEEE1872-2.owl [R=303,NC,L]
+
+# Redirect ontology version IRI to the given version (from v1.1.0 to .ttl)
+RewriteCond %{HTTP_ACCEPT} application/rdf\+xml [OR]
```
davidlehn, line 10: Is the intent to redirect rdf+xml and turtle to .ttl?  Similar below.  And similar above, sending turtle types to .owl.

davidlehn: I do see the PR comment.  Double checking this is what you want.

Miguel2617: The intention behind this is that the ontology, which was always written in Turtle syntax, is now also named as .ttl, which was previously named as .owl. However, this is intended to direct version the previous version 1.0.0 to the .owl file,  and the newer versions beginning with 1.1.0 to the .ttl file.

PR #5754, opened 2026-02-18
Add real-estate to w3id.org/italia

marco-trocino-demanio:
Add real-estate to w3id.org/italia redirecting to the Agenzia del Demanio semantic assets.

cc: @mfortini @Clou-dia - Could you please review this?


Clou-dia (commented):
Looks good to me — approving. Thanks!

PR #6001, opened 2026-04-30
HPCC Ontology

paitools:
# HPCC Ontology

This identifier is for the HPCC Ontology, an OWL ontology for modeling
HPC thermal management telemetry.

The ontology extends the W3C SSN/SOSA standards to capture HPC cooling concepts,
including heat pumps, heat exchangers, pumps, coolers, and re-coolers. 
Designed for the datacenter thermal management domain, the ontology supports anomaly detection,
real-time condition monitoring, and advanced system analytics.

The ontology is developed as part of the HPCCool, a framework for ontology-based
data access (OBDA), HPC cooling‑system management, and real-time analytics.

Pavle Ivanovic
Email: ivanovip@hsu-hh.de  
GitHub: https://github.com/paitools

### DIFF
--- a/hpcc/.htaccess
+++ b/hpcc/.htaccess
@@ -1,0 +1,31 @@
+# https://w3id.org/hpcc/ redirects via HTTP 303 using content negotiation:
+# - HTML requests are redirected to:
+#   https://paitools.github.io/HPCCOntology/documentation/index-en.html
+# - RDF/OWL requests are redirected to:
+#   https://paitools.github.io/HPCCOntology/HPCC.owl
+
+# ## Contact
+# This space is administered by:
+#
+# Pavle Ivanovic
+# ivanovip@hsu-hh.de
+# GitHub username: paitools
+
+RewriteEngine On
+
+# --- Human-facing documentation (HTML) ---
+RewriteCond %{HTTP_ACCEPT} text/html [OR]
+RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml
+RewriteRule ^$ https://paitools.github.io/HPCCOntology/documentation/index-en.html [R=303,L]
+
+# --- Machine-facing ontology (RDF/OWL) ---
+RewriteCond %{HTTP_ACCEPT} application/rdf\+xml [OR]
+RewriteCond %{HTTP_ACCEPT} text/turtle [OR]
+RewriteCond %{HTTP_ACCEPT} application/ld\+json [OR]
+RewriteCond %{HTTP_ACCEPT} application/n-triples [OR]
+RewriteCond %{HTTP_ACCEPT} application/owl\+xml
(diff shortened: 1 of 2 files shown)


hpcc/.htaccess
```diff
@@ -0,0 +1,31 @@
...
+RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml
+RewriteRule ^$ https://paitools.github.io/HPCCOntology/docs/index-en.html [R=303,L]
+
+# --- Machine-facing ontology (RDF/OWL) ---
+RewriteCond %{HTTP_ACCEPT} application/rdf\+xml [OR]
+RewriteCond %{HTTP_ACCEPT} text/turtle [OR]
+RewriteCond %{HTTP_ACCEPT} application/ld\+json [OR]
+RewriteCond %{HTTP_ACCEPT} application/n-triples [OR]
+RewriteCond %{HTTP_ACCEPT} application/owl\+xml
+RewriteRule ^$ https://paitools.github.io/HPCCOntology/HPCC.owl [R=303,L]
```
davidlehn, line 27: Do you want to return owl for all the types above?

paitools: Yes. I would like all RDF types resolved to the canonical OWL file. 
Thanks.

davidlehn: Ok.  That will confuse any client that asks for one type and gets another.

TallTed: > Ok. That will confuse any client that asks for one type and gets another.

Clients *usually* ask for multiple types, with quality ratings to indicate their preferences, via the `Accept:` header. Servers have their own quality ratings for the types they will deliver. Upon a request from a client, the server blends the two quality rating lists, and typically delivers the type which this blend rates with the highest quality.

The server always has control of what type is delivered in the end. The server declares the type it's delivering, in the `Content-type:` header. Remember that the server is delivering a _representation_ of the resource requested by the client, not necessarily a copy of the bits of that resource as stored on the server.

davidlehn: Quality param processing *should* happen, but I think every single file on this site is using the RewriteCond string grep method instead.  It's problematic but it's unclear what the solution should be.  It's non-trivial.  But since it appears to be fine in practice for now, I've let it be.

My point here was if a client sends `Accept: application/ld+json` and has no idea how to process OWL and gets back `Content-Type: application/owl+xml`, that's going to cause trouble.

TallTed: > My point here was if a client sends `Accept: application/ld+json` and has no idea how to process OWL and gets back `Content-Type: application/owl+xml`, that's going to cause trouble.

It already happens all the time. Not every time, but plenty. How a client handles whatever they get is up to the client. General web architecture calls for a graceful failure, which commonly means displaying the raw content or downloading the document (typically with some user interaction). It's rarely "trouble."

PR #6052, opened 2026-05-12
/ldac/ update profile and terms redirects

rosanna-smith:
Updated redirects for /profile and /terms under /ldac/

@ptsefton for approval


ptsefton:
Approved from LDaCA end

PR #6089, opened 2026-05-18
Add Tony Awards persistent identifiers

tony07122004:
## Brief Description

This pull request adds W3ID redirects for the Tony Awards Semantic Resource ontology and resource namespace.

Project repository:
https://github.com/tony07122004/group10-semantic-requirements

Ontology URI:
https://w3id.org/tony-awards/ontology

Ontology namespace:
https://w3id.org/tony-awards/ontology#

Resource namespace:
https://w3id.org/tony-awards/resource#

## General Checklist

- [x] The number of commits is minimal.
- [x] Commits only include redirects and basic information.

## New ID Directory Checklist


tony-awards/.htaccess
```diff
@@ -0,0 +1,24 @@
+Options +FollowSymLinks
+RewriteEngine On
+RewriteBase /tony-awards/
```
davidlehn: this is not needed for redirects

tony07122004:
Done, I removed the unnecessary Options directive from the .htaccess file.

dgarijo:
It's not the options directive, but the rewrite base what you should remove

davidlehn:
As a side question, if this is for a group "final project", why are you using this service?  Surely there is something better suited to support a temporary use?  We'd prefer this service is used for long lived resources.  The concern is abandonment of configs that stay around forever.

dgarijo:
This should go on the `doisr` namespace, which I will handle.
@tony07122004 please make the change

tony07122004:
Done. I moved the Tony Awards redirects to the doisr namespace and removed the old tony-awards directory.

PR #6103, opened 2026-05-20
Update /conjectures/ redirects for DEC

fvitali:
This PR updates the existing /conjectures/ namespace, administered by Fabio Vitali, to redirect the DEC entry point to the live DEC viewer.

Changes:

- Replaces the obsolete Altervista catch-all redirect.
- Adds redirects for:
  - https://w3id.org/conjectures/dec
  - https://w3id.org/conjectures/dec/viewer
- Updates the namespace README and contact metadata.

The target is the current public DEC viewer deployment:

http://204.216.209.229:3000/

### DIFF
--- a/conjectures/.htaccess
+++ b/conjectures/.htaccess
@@ -1,4 +1,5 @@
 RewriteEngine On
 
-RewriteRule   "^(.*)$" https://conjectures.altervista.org//$1 [R=303,L,QSA]
-
+# DEC live viewer.
+RewriteRule   "^dec/?$" http://204.216.209.229:3000/ [R=302,L,QSA]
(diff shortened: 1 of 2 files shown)


davidlehn:
- Browsers will probably complain when going from https to http.
- DNS names you can update are usually nicer than IPs.

Are you sure you want it setup this way?

fvitali:
You are totally right. I have now replaced the bare HTTP IP target with a stable HTTPS DNS name:

https://dec.fabiovitali.it/

Both redirects now point there:

https://w3id.org/conjectures/dec
https://w3id.org/conjectures/dec/viewer

Could you please proceed with the review when convenient?

PR #6201, opened 2026-06-15
Add /lde/ redirect for the LDElements (LDE) vocabularies

ddeboer:
Adds a new permanent identifier `/lde/` for the [LDElements (LDE)](https://github.com/ldelements/lde) vocabularies – composable building blocks for linked-data pipelines.

- `lde/.htaccess` – redirects `https://w3id.org/lde/` and its sub-paths (302, path-preserving) to <https://ldelements.github.io/lde/>, where the vocabulary documents are published.
- `lde/README.md` – identifier description, the sub-namespaces in use (`provenance#`, `metric#`, `failure#`, `distribution-validity-failure#`), and maintainer contact.

The redirect is intentionally temporary (302) while the vocabularies are being published; it will be switched to a permanent code once stable. Contact info is in both the `.htaccess` comment and the `README.md`.

### DIFF
--- a/lde/.htaccess
+++ b/lde/.htaccess
@@ -1,0 +1,15 @@
+# # /lde/
+#
+# Permanent identifiers for the LDElements (LDE) vocabularies.
+#
+# https://w3id.org/lde/ and its sub-paths redirect to
+# https://ldelements.github.io/lde/
+#
+# ## Contact
+# This space is administered by:
+#
+# David de Boer
+# GitHub username: ddeboer
+
+RewriteEngine on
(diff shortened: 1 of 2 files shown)


ddeboer:
Withdrawing this for now. On reflection, the LDElements vocabulary should stay a behind-the-scenes implementation detail rather than become publicly resolvable: the LDE TypeScript packages return plain objects, and the consuming applications own the RDF vocabulary that appears in output (standard W3C vocabularies plus their own namespace). With no LDE-coined identifiers surfacing in published output, there is nothing to resolve at `w3id.org/lde` yet. We may reopen a similar request later if that changes. Thanks!

PR #6224, opened 2026-06-18
Route the CACAO-RS ODRL profile URI

RubenPeeters:
Adds a specific redirect for `https://w3id.org/cacao/profile/rights` before the existing CACAO catch-all rule. The target profile is introduced by REEVALUATE/CACAO-RS#1.

### DIFF
--- a/cacao/README.md
+++ b/cacao/README.md
@@ -4,6 +4,9 @@
 ## Uses
 This [W3ID](https://w3id.org) provides a persistent URI namespace for the Cultural Artifacts Contextual Ontology (CACAO) resources.
 
+The CACAO-RS ODRL profile is available at
+<https://w3id.org/cacao/profile/rights>.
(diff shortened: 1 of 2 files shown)


cacao/README.md
```diff
@@ -12,4 +15,4 @@ _PhD Researcher_,
 Sint-Katelijne-Waver, Belgium,
 <ruben.peeters@kuleuven.be>,
 GitHub: [RubenPeeters](https://github.com/RubenPeeters),
```
TallTed, line 17: It appears that you want these lines to wrap, following the commas. Note that when the Markdown gets rendered as HTML, these lines *do not* wrap. You need to end each with two spaces (invisible formatting, which I do not recommend) or HTML line breaks, i.e., `<br>` (visible formatting, which I do recommend).

PR #6346, opened 2026-07-10
Dpp repo redirect

RichZele:
Adds a path-preserving redirect so https://w3id.org/dpp/repo and its sub-paths (/repo/shapes, /repo/use-cases, /repo/mappings, …) resolve to the DPP-DPPO-CEON-Ontology-Alignment GitHub repository.


dgarijo:
You are hosting files, which is not supported by w3id. Only redirects are appropriate. Please see the guidelines

PR #6361, opened 2026-07-14
Added SHACL redirects

k-vaughn:
## Brief Description

Updated htaccess to support
- redirects for SHACL files
- RDF redirects for classes and properties such that the appropriate ttl file is returned

## General Checklist

### DIFF
--- a/itsdata/.htaccess
+++ b/itsdata/.htaccess
@@ -10,8 +10,6 @@
 RewriteEngine On
 Options -MultiViews
 
-RewriteEngine On
-
 # Set FORMAT based on Accept header for RDF formats
 RewriteCond %{HTTP_ACCEPT} application/rdf\+xml [OR]
 RewriteCond %{HTTP_ACCEPT} text/turtle [OR]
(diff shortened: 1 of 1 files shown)


davidlehn:
I'm guessing this is fine, I can't tell.  Looks robot generated.  Are you sure you want this sort of rewrite and regex complexity?  I suspect more validation via regex is happening than needed.  Easier to have simpler matching and let invalid urls not work.  I have no idea if those `expr` constructs work.  We've limited some of the more advanced apache features to avoid doing extra work on the server, like what it looks like is happening here.  But if you want to merge it, we can.

k-vaughn:
Accepted. I will restructure to use a php script to simplify the logic in the .htaccess file

davidlehn:
That wasn't what I intended to suggest.  This service doesn't support php, so that won't work.  What is this trying to do?  All the regexes are difficult to read. :-)  The concern is maybe too much validation is happening here.  Obviously you could redirect all traffic to your own server and do any processing or run php there.  Most use cases here have (somewhat) simple patterns and can redirect to a static server elsewhere.  Is this more complex than that?

PR #6399, opened 2026-07-21
aio: resolve base/full release products (owl, obo, json)

turbomam:
Adds redirect rules to the existing `aio/` namespace so the AIO release products resolve, the same way `aio.owl` already does.

Today only `aio.owl` resolves. The registered ontology product `aio-full.owl` (and `aio-base.owl`, plus the `.obo` and `.json` serializations) fall through to the catch-all and do not resolve to the ontology. This PR adds one rule per product, each identical in form to the existing `aio.owl` rule, pointing at the raw file on `main`. All redirect targets return 200.

No existing rule is changed; the diff is purely additive.


turbomam:
Good catch. Escaped the literal dots in the product rewrite patterns so they only match the exact filenames (commit c7eca98).

PR #6394, opened 2026-07-21
OSO: content negotiation via Virtuoso instead of GitHub raw

Artemis-IA:
Virtuoso (oso-virtuoso v0.8.29, hosted at virtuoso.ifremer.fr) now reliably serves all RDF serializations for the OSO ontology (EMSO ERIC), replacing GitHub raw redirects.

- Versioned serializations (ttl/owl/nt/n3/trig/jsonld) -> `https://virtuoso.ifremer.fr/oso-versions/$1/OSO.<ext>` (Virtuoso DAV, verified 200 OK)
- Main ontology serializations -> `https://virtuoso.ifremer.fr/oso/rdf/OSO[.<ext>]` (Virtuoso native content negotiation + SPARQL CONSTRUCT, verified 303 -> 200 with correct Content-Type)
- HTML documentation still redirects to GitHub Pages (static hosting, unaffected)
- `rdfjson` format and unversioned `void`/`dcat` metadata left pointing to GitHub raw (not confirmed supported by Virtuoso yet)

All target URLs verified via curl against the isival validation environment before this PR.

PR #6407, opened 2026-07-23
OSO: skip intermediate redirect hop for main ontology IRI content negotiation

Artemis-IA:
Follow-up to #6394 (merged).

While auditing the FAIR score for OSO on EarthPortal, found that the main (unversioned) ontology IRI required 3 redirect hops to resolve to content:
1. Apache DirectorySlash 301 (add trailing slash)
2. w3id.org 303 -> https://virtuoso.ifremer.fr/oso/rdf/OSO
3. Virtuoso internal 303 -> https://virtuoso.ifremer.fr/oso/sparql?query=CONSTRUCT...

This PR removes hop #2 by pointing the .htaccess rules directly at the SPARQL CONSTRUCT URL, reducing to 2 hops total. Confirmed via curl that `https://virtuoso.ifremer.fr/oso/sparql?query=CONSTRUCT...&format=<mime>` returns 200 directly for all 6 formats (ttl, rdf+xml, jsonld, n-triples, n3, x-trig).

### DIFF
--- a/earthsemantics/OSO/.htaccess
+++ b/earthsemantics/OSO/.htaccess
@@ -55,19 +55,19 @@ RewriteRule ^$ https://emso-eric.github.io/oso-ontology/ [R=303,L]
 # Turtle
 RewriteCond %{HTTP_ACCEPT} text/turtle [OR]
 RewriteCond %{HTTP_ACCEPT} application/x-turtle
-RewriteRule ^$ https://virtuoso.ifremer.fr/oso/rdf/OSO [R=303,L]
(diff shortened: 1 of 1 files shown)


Artemis-IA:
Closing in favor of a corrected fix. Root-caused the O'\''FAIRe FAIR assessment tool (used by EarthPortal) source code: it never follows redirects (HttpURLConnection.setFollowRedirects(false)) and only accepts HTTP 200 or 302 as "resolvable" — 301 and 303 are always rejected, regardless of hop count. This PR reduced 3 hops to 2 but kept 303/301 status codes, so it would not have fixed F1Q4/A1Q1. Replacement PR incoming with the correct fix (302 status codes + DirectorySlash Off).

PR #6408, opened 2026-07-23
OSO: fix URI resolvability & content negotiation for O'FAIRe/EarthPortal FAIR assessment

Artemis-IA:
## Summary

Consolidates all OSO `.htaccess` fixes into a **single change** so the ontology IRIs pass O'FAIRe's `ResolvableURLTest`. O'FAIRe does **not** follow redirects and only accepts **HTTP 200/302** with a `Content-Type` matching the `Accept` header — `301`/`303` are rejected.

### `earthsemantics/.htaccess` (parent)
- Add `DirectorySlash Off` — stops Apache issuing a `301` when the trailing slash is missing.
- Exclude the `/earthsemantics/OSO` subtree from the catch-all EarthPortal redirect so the OSO `.htaccess` handles its own content negotiation.

### `earthsemantics/OSO/.htaccess`
- Add `DirectorySlash Off`.
- Serve content from the Virtuoso DAV versioned URLs (`https://virtuoso.ifremer.fr/oso-versions/1.1.0/OSO.*`) instead of the `/oso/rdf/OSO` intermediate hop — all targets verified to return `200` with the correct `Content-Type`.
- Use `R=302` (not `303`) for the main and versioned ontology IRI content negotiation.
- Override the `Content-Type` on `302` responses via `SetEnvIf` + `Header always set Content-Type` for the formats O'FAIRe tests (`application/json`, `application/xml`, `text/plain`, `application/rdf+xml`). `text/html` already carries a correct `Content-Type`. The `RewriteRule T=` flag does not apply to redirect responses, hence this approach.
- Remove the `rdfjson` rules: `OSO.rdfjson` does not exist in the `oso-ontology` repository (returned `404`) and `rdf+json` is not part of O'FAIRe's content-negotiation test set.

## Verification

Tested locally with a Docker Apache mirror and against the live redirect targets with `curl` (no redirect following, `HEAD`, per-format `Accept`):

| Test | Result |
|------|--------|
| Main IRI (`/OSO`, `/OSO/`), no `Accept` | `302` |
| Versioned IRI (`/OSO/1.1.0`, `/OSO/1.1.0/`), no `Accept` | `302` |
| Content negotiation, 5 O'FAIRe formats | `302`/`200` with matching `Content-Type` |
| Virtuoso DAV targets (`ttl/owl/jsonld/nt/n3/trig`) | `200` + correct `Content-Type` |
| GitHub Pages doc, `void.ttl`, `dcat.ttl` | `200` |
| Parent redirect for non-OSO paths | `302` to EarthPortal |

> Note: the `earthportal.eu/check_resolvability` live check will only reflect these fixes once this PR is merged and deployed to w3id.org.

### DIFF
--- a/earthsemantics/.htaccess
+++ b/earthsemantics/.htaccess
@@ -1,5 +1,9 @@
 Options +FollowSymLinks
 
 RewriteEngine on
+DirectorySlash Off
+
 # Redirect the base ontology URI to its EarthPortal page and each of its class/concept to its view
-RewriteRule ^(.+)$ https://earthportal.eu/ontologies/$1 [R=302,L] 
+# Exclude OSO subdirectory which has its own .htaccess for content negotiation
+RewriteCond %{REQUEST_URI} !^/earthsemantics/OSO [NC]
+RewriteRule ^(.+)$ https://earthportal.eu/ontologies/$1 [R=302,L]
--- a/earthsemantics/OSO/.htaccess
+++ b/earthsemantics/OSO/.htaccess
@@ -1,28 +1,54 @@
 Options +FollowSymLinks
 RewriteEngine On
+DirectorySlash Off
 
+# ====================================================================
+# Content-Type override for O'FAIRe content negotiation (A1Q3)
+# ====================================================================
+# O'FAIRe's ResolvableURLTest disables redirect following and only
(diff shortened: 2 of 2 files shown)


dgarijo:
@spiel-ifremer approve pls

spiel-ifremer (approved):

PR #6411, opened 2026-07-23
deliberation: fix /mappings redirect and default fallback for /ontology

Stocastico96:
Two fixes for the `deliberation` entry (https://w3id.org/deliberation), maintained by me:

1. **`/deliberation/mappings` returned 404 for every client.** The ontology's mappings module is referenced via `owl:imports <https://w3id.org/deliberation/mappings>`, but no rule existed for the path, so it fell through to the catch-all and hit a non-existent file. A rule now serves `mappings.owl`.
2. **`/deliberation/ontology` returned 404 for clients without a recognized `Accept` header** (e.g. plain `curl`, some RDF libraries). A default fallback now serves the RDF serialization, keeping the existing HTML/RDF content negotiation unchanged.

Tested against the GitHub Pages target: both `ontologies/mappings.owl` and `ontologies/deliberation.owl` resolve with HTTP 200.


davidlehn:
Fix conflicts.

Stocastico96:
Conflicts fixed — the branch has been rebuilt on top of current master (my fork's master had diverged from an old contribution). The PR now contains only the intended change to `deliberation/.htaccess`. Apologies for the noise from the force-push auto-closing the PR.

PR #6415, opened 2026-07-23
polylab: adding namespace

SirkoS:
## Brief Description

This adds a namespace for the [PolyLab ontology](https://dlr-dw.gitlab.io/poly-ontologies/polylab-ontology/).

> The PolyLab ontology is used for semantic description of laboratory processes. It connects planned and executed real world activities of researchers with the data involved in scientific experiments to improve the existing workflows. 

Note that the forwardings for version IRIs have been prepared although they are not (yet) supported by the hosted ontology.
This is planned for the future and should then not require an additional pull request here.

## General Checklist

## New ID Directory Checklist

### DIFF
--- a/polylab/.htaccess
+++ b/polylab/.htaccess
@@ -1,0 +1,125 @@
+# Turn off MultiViews
+Options -MultiViews
+
+# Directive to ensure *.rdf files served as appropriate content type,
+# if not present in main apache config
+AddType application/rdf+xml .rdf
+AddType application/rdf+xml .owl
+AddType text/turtle .ttl
+AddType application/n-triples .n3
+AddType application/ld+json .json
+AddType application/ld+json .jsonld
+
+RewriteEngine on
+
+
+############################## Rules for Previous Versions ##############################
+
+# Rewrite rule to serve HTML with local name
+RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+xml)
+RewriteCond %{HTTP_ACCEPT} \*/\* [OR]
+RewriteCond %{HTTP_ACCEPT} text/html [OR]
+RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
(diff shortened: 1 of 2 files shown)


polylab/.htaccess
```diff
@@ -0,0 +1,125 @@
...
+
+############################## Rules for Previous Versions ##############################
+
+# Rewrite rule to serve HTML with local name
+RewriteCond %{HTTP_ACCEPT} !application/rdf\+xml.*(text/html|application/xhtml\+xml)
+RewriteCond %{HTTP_ACCEPT} \*/\* [OR]
+RewriteCond %{HTTP_ACCEPT} text/html [OR]
+RewriteCond %{HTTP_ACCEPT} application/xhtml\+xml [OR]
+RewriteCond %{HTTP_USER_AGENT} ^Mozilla/.*
+RewriteCond %{REQUEST_URI} /(\d+)\.(\d+)\.(\d+)/(.+)$$
```
davidlehn: - Is the `$$` intentional?
- Why are all these using this RewriteCond matching instead of using a RewriteRule pattern and $N replacements?  Mostly curious, since I forgot about this style.  I assume it's tested and works.

SirkoS: > Is the $$ intentional?

The `$$` was a mistake that apparently did not affect the function

> Why are all these using this RewriteCond matching instead of using a RewriteRule pattern and $N replacements? Mostly curious, since I forgot about this style. I assume it's tested and works.

I admit that that my knowledge of htaccess is quite limited. 
I usually just build upon my last try and try to adapt to the next use case.
Here, I based on my previous namespace for [oto](https://github.com/perma-id/w3id.org/blob/master/oto/.htaccess).
If you have suggestion how to reduce the overall number of rules or improve the htaccess, those are more than welcome.

Regarding "tested", yes, by now I have a script that shows me whether all IRIs work as expected.
This was "all green" for this namespace - including the typo before.

PR #6442, opened 2026-07-27
New relese of CARE-SM Models (v2) needs independent namespace

markwilkinson:
## Brief Description
New release of CARE-SM Models (v2) needs independent namespace

## General Checklist

- [ ] Changes have been tested.
- [X] The number of commits is minimal. Squash if needed.
- [X] Commits only include redirects and basic information. Serving content and full documentation is not supported on this service.

## New ID Directory Checklist

- [X] Maintainer details are in `.htaccess` or `README.md`.
- [X] GitHub username ids are listed in the maintainer details.


davidlehn:
I've got to ask: Is there a plan for the next version?  Would be nice to avoid having N top level version dirs.

dgarijo:
@markwilkinson why not doing CARE-SM and then handling versions directly underneath?
(see comment from @davidlehn )

dgarijo:
Ping @markwilkinson

markwilkinson:
They are pointing at different servers... and we need both to exist long-term... and there is no "intermediate" server that could distinguish between one and the other - they're just pointing at vocabs that are independent from one another.

There isn't going to be a plethora of versions... there have been two versions over several years, so I am not sure where the concern is coming from.

markwilkinson:
If I had called it something completely different, I think this concern would never have been raised ;-)

dgarijo:
The concern is that we group all domain namespaces under the same folder (the repository is already big enough).
Therefore, you can have w3id.org/CARE-SM and handle any version there, like w3id.org/CARE-SM/v1 or similar.

@davidlehn since there are only 2 versions is this case acceptable? I don't want to start an anti pattern here.

markwilkinson:
AHHHHH!!!   I understand!!   Thank you!

Yes, that is completely reasonable.  I will close this PR and modify the existing record.  

Sorry for my misunderstanding, and thanks for your patience!

PR #6456, opened 2026-07-29
OSO: fix bare ontology IRI returning a directory listing (mod_alias instead of mod_rewrite)

Artemis-IA:
## Problem

`https://w3id.org/earthsemantics/OSO` (the bare ontology IRI, no trailing slash) returns **HTTP 200 with an Apache directory listing** instead of negotiating content:

```console
$ curl -sI -H "Accept: text/turtle" https://w3id.org/earthsemantics/OSO
HTTP/1.1 200 OK
Content-Type: text/turtle          # <-- forced by the Header directives

$ curl -s -H "Accept: text/turtle" https://w3id.org/earthsemantics/OSO | head -1
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
```

Because the HTML listing is relabelled as `application/xml` for a browser's `Accept`, browsers fail with:

> error on line 1 at column 55: Space required after the Public Identifier

`https://w3id.org/earthsemantics/OSO/` (with trailing slash) has always worked correctly.

## Root cause

With `DirectorySlash Off`, the request resolves to the `OSO/` directory, but **mod_rewrite cannot process it at all**. In per-directory context mod_rewrite strips the per-directory prefix `.../OSO/` from the filename; for this request the filename is `.../OSO` (no trailing slash), the strip fails, mod_rewrite declines, and **every `RewriteRule` in `OSO/.htaccess` is skipped**. The request falls through to `mod_autoindex`, while the `Header always set Content-Type` directives — which are *not* mod_rewrite and so still apply — relabel the listing.

I confirmed this on a local Apache 2.4 with the repo's `.htaccess` files mounted as-is and `LogLevel rewrite:trace4`. Replacing the whole file with a single catch-all:

```apache
RewriteRule ^(.*)$ http://diag.invalid/uri=$1 [R=302,L]
```

still returns 200 for `/earthsemantics/OSO` — the rule never fires — while `/earthsemantics/OSO/` and `/earthsemantics/OSO/ttl` both redirect as expected. Only `Header`/`SetEnvIf` survive.

This also explains the two earlier attempts:

- **#6429** added `RewriteCond %{REQUEST_FILENAME} -d` + `RewriteCond %{REQUEST_URI} !/$`. It could never have worked: the rules are skipped entirely, and in this context `%{REQUEST_URI}` is already `/earthsemantics/OSO/` **with** a trailing slash, so `!/$` never matches. The trace shows `RewriteCond: input='/earthsemantics/OSO/' pattern='!/$' =>` failing.
- **#6429** also added `Options -Indexes`, which only converted the listing into a `403 Forbidden`. **#6449** removed it, which restored the listing.

## Fix

Use `mod_alias`, which has no per-directory prefix limitation and runs in the fixup phase, before the autoindex handler:

```apache
RedirectMatch 302 ^/earthsemantics/OSO$ /earthsemantics/OSO/
```

`302` (not `301`) keeps O'FAIRe's `A1Q1`/`A1Q3` satisfied — it does not follow redirects and accepts only 200/302 — and the existing `Content-Type` overrides still apply to this redirect, so the negotiated type is preserved on the first hop.

`mod_alias` is already relied upon by 29 other `.htaccess` files in this repository, so it is guaranteed to be enabled.

This PR also trims the comment blocks in this file (180 → 126 lines), keeping only the non-obvious rationale.

## Verification

Apache 2.4 in Docker, `AllowOverride All`, both `earthsemantics/.htaccess` and `earthsemantics/OSO/.htaccess` mounted byte-identical to this branch.

Main IRI, before → after:

| `Accept` | `/OSO` before | `/OSO` after | `/OSO/` (unchanged) |
|---|---|---|---|
| `text/html` | 200 listing | 302 → `/OSO/` | 302 → `emso-eric.github.io/oso-ontology/` |
| `text/turtle` | 200 listing | 302 → `/OSO/`, `text/turtle` | 302 → `OSO.ttl` |
| `application/rdf+xml` | 200 listing | 302 → `/OSO/`, `application/rdf+xml` | 302 → `OSO.owl` |
| `application/xml` | 200 listing | 302 → `/OSO/`, `application/xml` | 302 → `OSO.owl` |
| `application/ld+json` | 200 listing | 302 → `/OSO/` | 302 → `OSO.jsonld` |
| `application/json` | 200 listing | 302 → `/OSO/`, `application/json` | 302 → `OSO.json` |
| `application/n-triples` | 200 listing | 302 → `/OSO/` | 302 → `OSO.nt` |
| `text/n3` | 200 listing | 302 → `/OSO/`, `text/n3` | 302 → `OSO.n3` |
| `application/trig` | 200 listing | 302 → `/OSO/` | 302 → `OSO.trig` |
| `text/plain` | 200 listing | 302 → `/OSO/`, `text/plain` | 302 → `OSO.ttl` |
| browser `Accept` string | 200 listing | 302 → `/OSO/`, `application/xml` | 302 → `emso-eric.github.io/oso-ontology/` |
| no `Accept` header | 200 listing | 302 → `/OSO/` | 302 → `emso-eric.github.io/oso-ontology/` |

No regressions elsewhere:

- All 22 explicit paths still return 302/303: `OSO.{ttl,owl,rdf,jsonld,json,nt,n3,trig}`, the bare `{ttl,owl,rdf,jsonld,json,nt,n3,trig}` aliases, `sparql`, `ui`, `void`, `dcat`.
- Versioned IRIs unchanged: `/OSO/1.1.0` negotiates per `Accept`, and with no `Accept` returns 302 → `OSO.ttl` (O'FAIRe `F1Q4`).
- `/OSO/<unknown>` returns a clean 404.
- The parent namespace still redirects: `/earthsemantics/Other` → 302 → `earthportal.eu/ontologies/Other`.
- No directory listing and no 403 is reachable under `/earthsemantics/OSO`.

### DIFF
--- a/earthsemantics/OSO/.htaccess
+++ b/earthsemantics/OSO/.htaccess
@@ -2,18 +2,9 @@ Options +FollowSymLinks
 RewriteEngine On
 DirectorySlash Off
 
-# ====================================================================
-# Content-Type override for O'FAIRe content negotiation (A1Q3)
-# ====================================================================
-# O'FAIRe's ResolvableURLTest disables redirect following and only
-# accepts HTTP 200/302 with Content-Type matching the Accept header.
-# The T= flag does NOT work on redirect responses. Instead, we use
-# SetEnvIf to detect the Accept header and Header always set to
-# override the Content-Type on the 302 response.
-# O'FAIRe tests these 5 formats: application/json, application/xml,
-# text/html, text/plain, application/rdf+xml.
-# text/html already has correct CT on 302 (text/html; charset=iso-8859-1),
-# so no override is needed for it.
+# O'FAIRe does not follow redirects and requires the Content-Type of the
+# 302 to match the Accept header. The T= flag has no effect on redirects,
+# so the Content-Type is forced via SetEnvIf + Header instead.
 SetEnvIf Accept "application/json" CT_JSON
 SetEnvIf Accept "application/xml" CT_XML
 SetEnvIf Accept "text/plain" CT_PLAIN
@@ -28,18 +19,13 @@ Header always set Content-Type "application/rdf+xml" env=CT_RDFXML
 Header always set Content-Type "text/turtle" env=CT_TURTLE
 Header always set Content-Type "text/n3" env=CT_N3
 
-# ====================================================================
-# Redirect directory root without trailing slash to trailing slash
-# ====================================================================
+# Trailing slash for the bare IRI. Must be mod_alias: for /earthsemantics/OSO
+# mod_rewrite cannot strip the per-directory prefix ".../OSO/" from the
+# filename ".../OSO", declines, and every RewriteRule below is skipped --
+# mod_autoindex then serves a directory listing as 200.
+RedirectMatch 302 ^/earthsemantics/OSO$ /earthsemantics/OSO/
 
-RewriteCond %{REQUEST_FILENAME} -d
-RewriteCond %{REQUEST_URI} !/$
-RewriteRule ^.*$ %{REQUEST_URI}/ [R=302,L]
-
-# ====================================================================
-# Versioned ontology IRI content negotiation
-# R=302 (not 303) for O'FAIRe F1Q4 compatibility (only accepts 200/302)
-# ====================================================================
+# Versioned IRIs -- 302 (not 303) for O'FAIRe F1Q4, which only accepts 200/302
 
 RewriteCond %{HTTP_ACCEPT} text/turtle [OR]
 RewriteCond %{HTTP_ACCEPT} application/x-turtle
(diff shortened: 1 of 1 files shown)


davidlehn:
Seems you're having some trouble with your droids?  Commit churn here and parent dir.  The robot output above, in the commit comments, and in previous PRs is too much to read.  Is it now doing what you want?  It seems maybe this got more complex than it needed to be.  Feel free to ask how to set things up.

Artemis-IA:
Thanks for the feedback, and sorry about the noise. The verbose commit messages and bot output were from an AI coding assistant — I'll keep those shorter going forward.

The three PRs each addressed a different layer of the same symptom (bare IRI returning a directory listing):

#6429 attempted to fix it with a mod_rewrite rule, but that rule could never fire — in per-directory context mod_rewrite declines to strip the per-directory prefix from a bare directory filename, so every RewriteRule in the file is skipped.
#6449 removed -Indexes to stop the 403, but that just exposed the underlying autoindex listing again.
#6456 is the actual fix: mod_alias RedirectMatch bypasses the mod_rewrite per-directory limitation entirely.
So yes, it's doing what I want now — verified locally on Apache 2.4 with all content types and paths. Happy to squash the commits before merge if you prefer a cleaner history.

PR #6468, opened 2026-07-30
Add object-provenance vocabulary redirect

RichardHerold:
## Brief Description

Claims the top-level identifier `object-provenance` for a vocabulary of terms describing the recording, batching, and archival of provenance claims about physical objects (mineral specimens, gemstones, and other collectibles).

- **Identifier:** `https://w3id.org/object-provenance/v1#` — versioned and hash-based, so terms are fragments and the rules serve the whole vocabulary.
- **Redirect target:** `https://object-provenance.github.io/vocab/v1/`, content-negotiated for JSON-LD. **Live now:** [landing page](https://object-provenance.github.io/vocab/v1/), [context.jsonld](https://object-provenance.github.io/vocab/v1/context.jsonld), [KEY-SET.md](https://object-provenance.github.io/vocab/v1/KEY-SET.md) all return 200.
- **302, not 301,** deliberately: a 301 is cached permanently and cannot be recalled, which would defeat the repointing ability this service provides.
- **Why permanence is a hard requirement:** these IRIs sit inside canonicalized JSON documents (JCS, RFC 8785) that are hashed and anchored to a public blockchain — a record can never be reissued, so its vocabulary must never depend on one organization's domain. The identifier is needed *before* the first anchored record, which is why the request comes now.
- **Scope:** only terms with no published equivalent are meant to be minted; the README names the five v1 terms that currently duplicate Dublin Core / PROV-O and the plan for them, rather than claiming a discipline the context file doesn't yet meet.
- Distinct from the existing `w3id.org/provenance` (a CIDOC-CRM database of artworks): this is a vocabulary for recording and archival mechanics, not a dataset.

## General Checklist
- [x] The number of commits is minimal. Squash if needed. *(single commit)*
- [x] Commits only include redirects and basic information. Serving content and full documentation is not supported on this service. *(two files: `.htaccess` + `README.md`; all content is served from the redirect target)*

## New ID Directory Checklist
- [x] Maintainer details are in `.htaccess` or `README.md`. *(both)*

## Testing

Reproduced the repo's Apache recipe locally (httpd 2.4, `AllowOverride All`, `mod_rewrite` + `mod_headers`, full repository tree with this PR's ref checked out) and exercised the rules:

| Request against local checkout | Observed |
|---|---|
| `/object-provenance/v1` · `Accept: application/ld+json` | `302 → https://object-provenance.github.io/vocab/v1/context.jsonld` |
| `/object-provenance/v1` · `Accept: application/json` | `302 → …/vocab/v1/context.jsonld` |
| `/object-provenance/v1` · browser `Accept` | `302 → …/vocab/v1/` |
| `/object-provenance/v1/` (trailing slash) | same as above |
| `/object-provenance/` | `302 → …/vocab/` |

Control entries (`/provenance/`, `/examples/ontology`) behaved per their own files under the same harness, so the harness is faithful. All redirect targets return 200 in production.

## Contact

RelicQuery — hello@relicquery.org · GitHub: @RichardHerold

Happy to adjust the identifier or the rules if anything conflicts with the naming policy or something reserved.

### DIFF
--- a/object-provenance/.htaccess
+++ b/object-provenance/.htaccess
@@ -1,0 +1,32 @@
+# # /object-provenance/
+#
+# A vocabulary of terms for recording, batching, and archiving provenance
+# claims about physical objects. See README.md in this directory.
+#
+# https://w3id.org/object-provenance/v1# redirects to
+# https://object-provenance.github.io/vocab/v1/
+# with content negotiation for JSON-LD.
+#
+# The namespace is hash-based, so the fragment never reaches the server and
+# these rules serve the whole vocabulary.
+#
+# 302 rather than 301: a 301 is cached permanently by browsers and
+# intermediaries and cannot be recalled, which would destroy the repointing
+# ability this service exists to provide.
+#
+# ## Contact
+# This space is administered by:
+#
+# RelicQuery — hello@relicquery.org
+# GitHub username: RichardHerold
+
+Options +FollowSymLinks
+RewriteEngine on
+
+RewriteCond %{HTTP_ACCEPT} application/ld\+json [OR]
+RewriteCond %{HTTP_ACCEPT} application/json
+RewriteRule ^v1/?$ https://object-provenance.github.io/vocab/v1/context.jsonld [R=302,L]
+
+RewriteRule ^v1/?$ https://object-provenance.github.io/vocab/v1/ [R=302,L]
+
+RewriteRule ^/?$ https://object-provenance.github.io/vocab/ [R=302,L]
--- a/object-provenance/README.md
+++ b/object-provenance/README.md
@@ -1,0 +1,94 @@
+# object-provenance
+
+Permanent identifiers for an object provenance vocabulary: terms describing
+the recording, batching, and archival of provenance claims about physical
+objects (mineral specimens, gemstones, and other collectibles).
+
+## Namespace
+
+    https://w3id.org/object-provenance/v1#
+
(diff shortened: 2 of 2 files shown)


RichardHerold:
Amended the branch (still one commit) to correct something in my own README before a reviewer has to find it.

The Scope section claimed the namespace "covers only terms with no published equivalent." That is the rule this vocabulary is held to, but `v1` does not currently meet it — `op:title`, `op:description`, `op:recordIdentifier` and `op:publishedAt` duplicate Dublin Core, and `op:provenanceEvents` / `op:ownerDisplayName` duplicate PROV-O attribution. Anyone opening the linked `context.jsonld` would have seen the claim contradicted by the first few terms.

The README now names those five explicitly, and says why the mapping is scheduled rather than already done: PROV-O attribution changes the shape of the document, not just its key strings, so it is a restructuring rather than a rename. New records adopt the mapped terms under a later version; records already anchored under `v1` are never remapped, because rewriting them would change their hashes.

Requesting the identifier at this stage rather than after the vocabulary is tidier is deliberate — it has to exist before the first anchored record, not after. Happy to hold the PR until the mapping lands if you would rather review a cleaner term set.

davidlehn:
That's a lot of docs in the README here.  It would be better to host all those on your own site.  Easier to maintain for you with fewer PRs here.

dgarijo:
Ping @RichardHerold (simplify the readme)

dgarijo:
Closing due to inactivity. Please reopen if appropriate

PR #6378, opened 2026-07-17
Update SSTIM catalog and ecosystem routes

ttm:
## What changed

All changes are scoped to the existing `sstim/` namespace directory.

- add content-negotiated routes for the `/sstim/ecosystem` ontology module
  (Turtle default, JSON-LD, RDF/XML; HTML requests go to the project landing
  page, matching every other module route);
- add exact Turtle routes for the stable BSC catalog identities — the BSC
  framework, the BSC Lab and BioSynCare implementations, and the Patch Studio
  software component (intentionally distinct from the `/sstim/patch-studio`
  ontology module); and
- add four compact live namespace rules for `/sstim/specialist/{id}`,
  `/sstim/organization/{id}`, and
  `/sstim/ecosystem-record/{relationship,activity,role}/{id}`: RDF requests
  resolve to the mutable live-only projection, HTML requests to the landing
  page, and the reserved `synthetic-*` fixture slug prefix is rejected by the
  rules themselves.

## Why

SSTIM 0.7.0 released the qualified ecosystem relationship contract, and the
project has deployed the corresponding static catalog resources and the
separately hosted mutable live-only projection. Exact rules avoid conflating
the Patch Studio software component with its ontology module.

The registry configuration stays generic on purpose: dataset membership
belongs to the live RDF projection, so adding, correcting, or retracting an
ecosystem record never requires a w3id.org rule change. No person,
organization, or fixture is named in `.htaccess`; synthetic contract fixtures
use the reserved `synthetic-*` slug and remain reachable only through their
direct test artifact. The privacy and removal boundary is documented in
`sstim/README.md`.

The namespace continues to serve the SSTIM OWL/SKOS ontology, published under
CC BY 4.0.

## Impact

RDF clients can dereference each catalog subject to its owning Turtle resource
and any live ecosystem identifier to the mutable projection. Browser requests
continue to reach the project landing page. Existing SSTIM routes, content
negotiation, and the frozen version snapshots (including 0.7.0) are unchanged.

## Verification

- Every static redirect target returns HTTP 200 with the expected Turtle,
  JSON-LD, or RDF/XML media type and wildcard CORS (re-verified 2026-07-17).
- The mutable `current.ttl` endpoint returns HTTP 200, `text/turtle`, wildcard
  CORS, and `Cache-Control: no-store`; its deployed SHA-256 is
  `1e8a96eb4ecf0fa55516dbb862086027d0d47f80aa05189aaff3cfda477a15a5`.
[...]

PR #6480, opened 2026-08-02
sstim: add modular profile, module, and manifest routes

ttm:
## Brief Description

Updates the existing `sstim/` directory for the SSTIM ontology's modular
reorganisation. All targets are `https://labiosyncare.github.io/ontology/`, as
before; this PR only adds and retargets redirects.

**New routes**

- `/sstim/profile/{kernel,core,core-plus,full}` — the four W3C PROF profile
  entry points that name SSTIM's conformance closures.
- `/sstim/{stimulus,common,technique,configuration,session,evidence,neuromodulation,core-shapes,neuromodulation-evidence,evidence-exposure,technique-exposure}`
  — modules extracted from the former monolithic core file.
- `/sstim/kernel` — the exact two-class kernel distribution.
- `/sstim/module/exposure` — the exact Exposure module.
- `/sstim/manifest` and `/sstim/manifest-schema/1` — the machine-readable bill
  of materials and its JSON Schema.

**Retargeted routes**

- `/sstim` and `/sstim/exposure` now serve generated namespace catalogues rather
  than a single file. Both are hash namespaces whose terms are now spread across
  several modules, so a catalogue is what makes `/sstim#Preset` and
  `/sstim/exposure#StimulusChannel` still dereference to their definitions.
  `/sstim/kernel` and `/sstim/module/exposure` are the exact single-module
  documents for anyone who wants those instead.

**Replaced wildcard**

The snapshot rules were two regex wildcards matching any `X.Y.Z` and any
`*.ttl` under it. They are replaced by explicit per-file rules generated from
the frozen snapshots that actually exist, so a plausible-looking but nonexistent
version or filename now 404s instead of redirecting to a missing target. This
also fixes a latent problem: the wildcard sent every `/sstim/X.Y.Z` to
`sstim-core.ttl`, which was correct while that file was the whole ontology and
would have become wrong at the next release, where it is a two-class kernel.
Future versions get a `/manifest` route only once a frozen `manifest.json`
sibling exists beside them.

`Vary: Accept` is set, since these routes negotiate Turtle, JSON-LD, RDF/XML and
HTML from the same URL.

Rules are generated and verified in the source repository
(`scripts/sstim-w3id-snapshot-routes.mjs`, `scripts/check-w3id-route-targets.mjs`),
which check that the committed route table matches the frozen snapshots on disk
and that every `/ontology/` target is an artifact the project actually
publishes.

## General Checklist

## New ID Directory Checklist

Not applicable — `sstim/` already exists.

Maintainer: Renato Fabbri, [@ttm](https://github.com/ttm), as recorded in
`sstim/README.md` and previously in #6378, #6391 and #6393.

- [x] Please squash commits for me. I understand this will likely require resyncing my local repository before making further PRs.

### DIFF
--- a/sstim/.htaccess
+++ b/sstim/.htaccess
@@ -13,17 +13,131 @@ AddType application/rdf+xml   .rdf .owl
 AddType application/ld+json   .jsonld
 AddType application/n-triples .nt
 
-Header set Access-Control-Allow-Origin *
+Header always set Access-Control-Allow-Origin *
+Header always set Vary "Accept"
 
 RewriteEngine On
 
 # -------------------------------------------------------------------------
-# Versioned immutable snapshots. These routes back owl:versionIRI values such
-# as /sstim/0.1.0 and direct frozen-file URLs such as
-# /sstim/0.1.0/sstim-core.ttl. (Turtle only — snapshots are not multi-format.)
+# Versioned immutable snapshots. This generated region is fail-closed: only
+# files that actually exist in a frozen repository snapshot acquire routes.
+# Future modular snapshots add exact /manifest and /manifest.schema.json routes
+# only after those immutable sibling files exist. Regenerate with
+# `node scripts/sstim-w3id-snapshot-routes.mjs --write` after cutting a release.
 # -------------------------------------------------------------------------
-RewriteRule ^([0-9]+\.[0-9]+\.[0-9]+)/([A-Za-z0-9._-]+\.ttl)$ https://labiosyncare.github.io/ontology/$1/$2 [R=302,L]
-RewriteRule ^([0-9]+\.[0-9]+\.[0-9]+)/?$ https://labiosyncare.github.io/ontology/$1/sstim-core.ttl [R=302,L]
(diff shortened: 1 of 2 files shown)


ttm:
Updated and squashed to a single commit.

One fix since opening: `/sstim/vocab` with `Accept: text/html` was resolving to the OWL reference index, but the SKOS vocabulary has its own generated page at `/ontology/docs/vocab/`. Only the HTML representation changed; Turtle, JSON-LD and RDF/XML for that route are unchanged.

Also set `Access-Control-Allow-Origin` with `always` so it is present on the 303 responses themselves, which a cross-origin `fetch` following the redirect needs.

On testing: every redirect target was confirmed to return 200 before opening. Since w3id.org performs the negotiation itself, the rules are additionally checked in the source repository against a model of the relevant mod_rewrite subset — Turtle defaulting, absent and wildcard `Accept`, browser headers, and `q=0` both making a type unacceptable and not suppressing the others. Happy to adjust anything that does not fit house style.

PR #6487, opened 2026-08-04
Add gUFOevents identifiers to the nemo space

jpalmeida:
## Brief Description

To add a newly published ontology (gUFOevents).

gUFOevents (https://github.com/nemo-ufes/gufoevents) is an extension of gUFO with an account of events. Its ontology IRI, version IRI, documentation, SHACL shapes and eight example namespaces are all under https://w3id.org/nemo/, and until now every one of them fell through to the group's website.

The ontology IRI is content negotiated: a browser is sent to the documentation and everything else to the Turtle. The test is on text/html and comes first, because browsers ask for more than they will use -- Firefox sends application/xml;q=0.9 -- while RDF tooling does not ask for text/html at all.

The version IRI resolves to a tag rather than to the default branch, so that it keeps denoting version 1.0.0.

Tested with a local Apache checkout: all 21 identifiers redirect as intended under no Accept header, a browser's, Protege's and text/turtle, and the existing gufoshapes and catch-all rules are unaffected.

Contact: João Paulo A. Almeida (GitHub: jpalmeida)

## General Checklist
- [ x ] Changes have been tested.
- [ x ] The number of commits is minimal. Squash if needed.
- [ x ] Commits only include redirects and basic information. Serving content and full documentation is not supported on this service.

### DIFF
--- a/nemo/.htaccess
+++ b/nemo/.htaccess
@@ -6,13 +6,72 @@
 # https://w3id.org/nemo/gufoshapes redirects to
 # https://nemo-ufes.github.io/gufoshapes/gufoshapes.ttl
 #
+# gUFOevents <https://github.com/nemo-ufes/gufoevents>:
+#
+# https://w3id.org/nemo/gufoevents            the ontology (Turtle), or its
+#                                             documentation for a web browser
+# https://w3id.org/nemo/gufoevents/1.0.0      the version 1.0.0 ontology
+# https://w3id.org/nemo/doc/gufoevents        the documentation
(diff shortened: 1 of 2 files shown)


davidlehn:
`Claude-Session: https://claude.ai/code/session_01E6A5LyWBVxUvfCNUnsAb8x`.
What is the point of that?  I assume only you have access to it.

jpalmeida:
> `Claude-Session: https://claude.ai/code/session_01E6A5LyWBVxUvfCNUnsAb8x`. What is the point of that? I assume only you have access to it.

thanks, removed it now

PR #6488, opened 2026-08-04
Add /semprini/ for the Semprini metamodel ontology

JuhaKor:
This adds `/semprini/`, the namespace of the Semprini metamodel ontology.

Semprini is an openly licensed (Apache-2.0 / CC BY 4.0) compiler that turns the business
models and taxonomies an organization already maintains into a governed RDF knowledge
graph. Each adopting organization runs its own instance with its own base IRI, but every
instance binds to one shared metamodel — the `sem:` vocabulary published under this
namespace — so a SPARQL query, SHACL shape or agent written against `sem:` works against
any instance. That shared metamodel is why the namespace belongs on w3id.org rather than
on a company domain: resolution must not depend on one organization keeping a domain
registered.

Metamodel terms are hash IRIs, for example `https://w3id.org/semprini/ontology#Entity`,
so requests arrive for the ontology document itself.

**Redirect targets (all live now):**

| Request | Resolves to |
|---|---|
| `/semprini/ontology` with an RDF `Accept` | https://juhakor.github.io/semprini/ontology/sem.ttl |
| `/semprini/ontology` from a browser | https://juhakor.github.io/semprini/ontology/ |
| `/semprini/ontology/0.1.0` | the same pair under `/ontology/0.1.0/` |
| `/semprini/` | https://juhakor.github.io/semprini/ |

The ontology is published in Turtle only, so every RDF media type resolves to the one
document, which is served as `text/turtle`. Redirects are `302` rather than `301`, since
the hosting location may change while the identifiers may not.

The rules were tested locally against 45 path and `Accept` combinations before
submission; every rule targets a URL that currently returns 200.

**Contact:** Juha Korpela (@JuhaKor), juha.korpela@datakor.fi, on behalf of Datakor
Consulting Oy, Finland. Contact details are in both `.htaccess` and `README.md`.

Source repository: https://github.com/JuhaKor/semprini

### DIFF
--- a/semprini/.htaccess
+++ b/semprini/.htaccess
@@ -1,26 +1,10 @@
 # # /semprini/
 #
-# The namespace of the Semprini metamodel ontology.
-#
-# Semprini is an openly licensed compiler that turns the business models and
-# taxonomies an organization already maintains into a governed RDF knowledge
-# graph. Every deployment binds to this one shared metamodel, so the namespace
-# must resolve independently of any single organization's domain.
-#
-# Metamodel terms are hash IRIs under https://w3id.org/semprini/ontology#
-# (for example https://w3id.org/semprini/ontology#Entity), so requests arrive
-# for the ontology document itself:
-#
-#   /semprini/                  -> project home
-#   /semprini/ontology          -> the current ontology: Turtle for RDF clients,
-#                                  documentation for browsers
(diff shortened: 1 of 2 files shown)


davidlehn:
The docs in this are verbose. Have you considered using a link to your site instead?

JuhaKor:
Fair point, had some WIP stuff on the site so the docs had duplicate information. Trimmed the docs now.

PR #6501, opened 2026-08-05
Retire the 2025-2026 redirects of the /FAIR-course-UT/ namespace

luizbonino:
The `/FAIR-course-UT/` namespace serves the FAIR Principles and FAIRification Process course of the University of Twente. Each edition of the course adds redirects for the student project repositories, and those repositories are not maintained after the course ends, so the redirects stop being useful.

This pull request removes the redirects of the 2025-2026 edition. The namespace itself is kept, because it is reserved for the coming editions, and the file now explains that convention so the intent is clear to a later reader.

It also adds the maintainer contact block that the contribution guidelines ask for, which this `.htaccess` was missing. The address matches the one already given in the namespace `Readme.md`.

My apologies for the noise these entries caused while the edition was running.


dgarijo:
Isn't it easier if you redirect to your server in a generic manner and host there the assignments? So we avoid the noise...

PR #6554, opened 2026-08-14
Preserve FAIRagro namespace and enable lowercase IRIs

cscheuner:
This PR renames the FAIRagro directory to fairagro so that the existing published lowercase IRIs resolve correctly.

The original FAIRagro path is preserved with a redirect to the canonical lowercase namespace.

This ensures that both existing https://w3id.org/FAIRagro/... links and the published https://w3id.org/fairagro/... IRIs remain usable.


davidlehn:
This is problematic.  We have a rule to not have multiple filenames that only differ by case.  Too many people using case insensitive file systems were causing havoc with broken commits and PRs.  Can you adapt to using one form or the other?  (There is a way to support both, but we'd really prefer to not add more special cases.)

dgarijo:
Ping @cscheuner

cscheuner:
Thanks for pointing this out. I have adapted the PR to use fairagro/ as the single directory name and removed the case-only duplicate. This also matches the lowercase form of the already published IRIs (https://w3id.org/fairagro/...).

PR #6565, opened 2026-08-17
added ro-crate-masp profile to ro directory

rosanna-smith:
Added ro-crate-masp/profile to ro directory.
ro-crate-masp is starting to see some adoption so we need a profile URI.
https://w3id.org/ro/ro-crate-masp/profile
At the moment this is housed in a masp tools repository, when the profile is more mature it will be moved to the research object github.
@ptsefton for approval.


dgarijo:
ping @ptsefton

ptsefton:
This is now being implemeted, eg in novacrate by @craquet so we need a URI for the profile.

ptsefton:
Admins, please merge this.

PR #6581, opened 2026-08-20
Croissant Policy Profile: add /croissant-policy namespace

doytsujin:
## What this adds

`/croissant-policy` — a persistent namespace for the **Croissant Policy Profile**, an additive policy profile for [Croissant](https://mlcommons.org/working-groups/data/croissant/) dataset descriptors.

Croissant 1.1 lets a dataset carry machine-readable data use conditions; it does not specify how any of them is evaluated. This profile supplies that half: a dataset declares the operations it admits and the conditions under which it admits them, from a closed set of five operators, so a gate in front of the data can permit or refuse a request from the descriptor alone and leave a record of what was checked.

This is an independent profile built on Croissant. It is not an MLCommons product and carries no endorsement from the Croissant project.

## Redirects

All `303 See Other` to `https://doytsujin.github.io/ok-croissant-policy-profile/ns/`.

| Request | Resolves to | Media type |
|---|---|---|
| `/croissant-policy/` | version index | `text/html` |
| `/croissant-policy/0.1.0` | profile document | `text/html` |
| `/croissant-policy/0.1.0` with `Accept: application/ld+json` | `context.jsonld` | `application/ld+json` |
| `/croissant-policy/0.1.0/profile.jsonld` | PROF description of the profile's resources | `application/ld+json` |
| `/croissant-policy/0.1.0/shapes.ttl` | SHACL shapes for the conformance clauses | `text/turtle` |
| `/croissant-policy/0.1.0/shapes-odrl.ttl` | the same clauses for the ODRL carrier | `text/turtle` |
| `/croissant-policy/0.1.0/<term>` | the profile document defining that term | `text/html` |

303 rather than 302 because the namespace identifies a profile rather than a document.

## Testing

Every target above returns `200` with the media type listed, today — the redirect is the only missing piece.

The rules themselves are checked by [`tests/test_w3id_redirects.py`](https://github.com/doytsujin/ok-croissant-policy-profile/blob/main/tests/test_w3id_redirects.py) in the source repository, which parses this `.htaccess` and applies it to every path in the table: rule ordering and the `[L]` flag, the `SetEnvIf` base variable, and the one `Accept` condition. It also asserts the things ordering makes easy to get wrong — that the named artifacts are not shadowed by the generic term rule, and that the catch-all is last.

## Versioning

A version is never changed in place, and a new version gets a new path rather than a new host. Extending the profile's operator set is a version bump rather than a profile extension, because a conforming evaluator must refuse an operator it does not implement — so a term added silently to an existing version would turn a permit into a refusal for older consumers.

## Contact

Alexander Chernov — GitHub [@doytsujin](https://github.com/doytsujin), ORCID [0009-0007-3198-2712](https://orcid.org/0009-0007-3198-2712).

Source, specification and issues: <https://github.com/doytsujin/ok-croissant-policy-profile>
Archived at [doi:10.5281/zenodo.22018156](https://doi.org/10.5281/zenodo.22018156).

### DIFF
--- a/croissant-policy/.htaccess
+++ b/croissant-policy/.htaccess
@@ -1,0 +1,60 @@
+# Name of the project: Croissant Policy Profile
+#
+# Description: An additive policy layer for Croissant dataset descriptors.
+#   Croissant describes what a dataset is; it does not say what may be done with
+#   it. This profile lets a dataset declare the operations it admits and the
+#   conditions under which it admits them, so that a gate in front of the data
+#   can permit or refuse a request from the descriptor alone and leave a record
+#   of what it checked.
+#
+#   This is an independent profile built on Croissant. It is not an MLCommons
+#   product and carries no endorsement from the Croissant project.
+#
+# Specification, reference implementation and conformance validator:
+#   https://github.com/doytsujin/ok-croissant-policy-profile
+#
+# Contacts:
+# - Alexander Chernov, GitHub username: doytsujin
+#   ORCID: https://orcid.org/0009-0007-3198-2712
+#   Preferred route: issues on the repository above
+
+Header set Access-Control-Allow-Origin *
+Options +FollowSymLinks -MultiViews
+AddType application/ld+json .jsonld
+AddType text/turtle .ttl
+
+RewriteEngine on
+
+SetEnvIf Request_URI ^.*$ PROFILE_URL=https://doytsujin.github.io/ok-croissant-policy-profile/ns
+
+# A specific version, asked for as JSON-LD: the context document.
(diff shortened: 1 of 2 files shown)


davidlehn:
I merged https://github.com/perma-id/w3id.org/pull/6576 before seeing this. There are now conflicts here.  Please resolve or do another PR to make changes.

doytsujin:
Closing as a duplicate — this namespace was already added and merged in #6576 (2026-08-21), and https://w3id.org/croissant-policy/0.1.0 now resolves correctly under both a plain request and JSON-LD content negotiation. Apologies for the extra review load. Thanks to the maintainers.

PR #6616, opened 2026-08-30
pkm: add content negotiation for the PKM vocabulary

dpw67:
## What

Updates `pkm/.htaccess` to serve RDF to clients that ask for it, and refreshes `pkm/README.md` to describe what now resolves.

The `/pkm` namespace was registered as a plain 302 to GitHub Pages while there was nothing behind it. There is now a published SKOS vocabulary — 223 concepts and 18 collections, each with its own Turtle file — so the redirect needs to distinguish an RDF client from a browser.

## Rules added

| Request | Response |
| --- | --- |
| `/pkm/vocab/Recipe` + `Accept: text/turtle` | 303 → `vocab/terms/Recipe.ttl` |
| `/pkm/vocab` + `Accept: text/turtle` | 303 → `vocab/pkm-vocab.ttl` |
| `/pkm/agents`, `/pkm/resources` | 303 → their `index.ttl` |
| `/pkm/ontology`, `/pkm/taxonomy` | 303 → their Turtle |
| anything else | 302 → GitHub Pages, **unchanged** |

`303 See Other` rather than 302 on the term rule is the httpRange-14 convention: a concept is not a document, so the redirect says "here is a document *about* it."

The term pattern `^vocab/([A-Za-z][A-Za-z0-9_-]*)$` has no dot in the character class, so `/vocab/Recipe.ttl` and the version IRI `/vocab/0.1.1` fall through to the catch-all rather than matching. The catch-all stays last.

## Checks

- Redirect targets verified live: <https://dpw67.github.io/pkm/vocab/terms/Recipe.ttl>, <https://dpw67.github.io/pkm/vocab/pkm-vocab.ttl>, `agents/index.ttl`, `resources/index.ttl`.
- Existing behaviour for every path not listed above is byte-identical.
- Single commit, contact info in both files.

Namespace source: <https://github.com/dpw67/pkm>

### DIFF
--- a/pkm/.htaccess
+++ b/pkm/.htaccess
@@ -1,5 +1,45 @@
-# w3id.org/pkm — KSA Personal Knowledge Management ontology
-# Contact: Doug Warren <doug78645@gmail.com> (GitHub: @dpw67)
+# https://w3id.org/pkm — PKM (Personal Knowledge Management) namespace
+#
+# A SKOS vocabulary and related Linked Data for the Knowledge System
+# Architecture (KSA) project. Served from GitHub Pages at dpw67.github.io/pkm.
+#
+# ## Contact
+# This space is administered by:
+#
+# Doug Warren
+# GitHub username: dpw67
+# Email: doug@warrenweb.net
+
+Options +FollowSymLinks
 RewriteEngine on
-RewriteRule ^$ https://dpw67.github.io/pkm/ [R=302,L]
-RewriteRule ^(.+)$ https://dpw67.github.io/pkm/$1 [R=302,L]
+
+# Individual vocabulary terms: /vocab/Recipe -> vocab/terms/Recipe.ttl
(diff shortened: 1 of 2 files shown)


dpw67:
Re-verified all four redirect targets just now, after a GitHub Pages rebuild — every one returns `200` with `Content-Type: text/turtle`:

| Target | Status | Bytes |
| --- | --- | ---: |
| `vocab/terms/Recipe.ttl` | 200 | 1,655 |
| `vocab/pkm-vocab.ttl` | 200 | 265,116 |
| `agents/index.ttl` | 200 | 669 |
| `resources/index.ttl` | 200 | 1,667 |

The HTML fallback (`^(.*)$ → https://dpw67.github.io/pkm/$1`) is the behaviour live today, so merging this only adds the Turtle branch — no existing URI changes.

PR #6618, opened 2026-08-31
update zcap/.htaccess to redirect to w3c-ccg.github.io not digitalbazaar.github.io

gobengo:
## Brief Description

This PR updates `w3id.org/zcap/` to redirect to URLs at `w3c-ccg.github.io` instead of `digitalbazaar.github.io`, reflecting that zcap is now a work item in W3C Credentials Community Group, and the JSON-LD context will be published by the [W3C CCG zcap-spec repo](https://github.com/w3c-ccg/zcap-spec/), not digitalbazaar's zcap-context repo.

My motivation for this PR is to resolve this issue in the zcap-spec https://github.com/w3c-ccg/zcap-spec/issues/89

## General Checklist

  - I asked an LLM if there are any syntax changes. It said there are none in my changes

## New ID Directory Checklist

### DIFF
--- a/zcap/.htaccess
+++ b/zcap/.htaccess
@@ -1,16 +1,26 @@
-# Authorization Capabilities (zCap) Context
+# Authorization Capabilities
 #
 # homepage:
-#   https://digitalbazaar.github.io/zcap-context/
+#   https://w3c-ccg.github.io/zcap-spec/
 # maintainers:
 #   @msporny
 #   @dlongley
 #   @dmitrizagidulin
 #   @davidlehn
-#
+#   @gobengo
+
 Header set Access-Control-Allow-Origin *
 Header set Access-Control-Allow-Headers DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified$
 Options +FollowSymLinks
 RewriteEngine on
-RewriteRule ^$ https://digitalbazaar.github.io/zcap-context [R=302,L]
-RewriteRule ^v1$ https://digitalbazaar.github.io/zcap-context/contexts/zcap-v1.jsonld [R=302,L]
+
+# /
+# Redirect to zcap spec
+RewriteRule ^$ https://w3c-ccg.github.io/zcap-spec/ [R=302,L]
+
+# /v1
+# Redirect to version 1 of zcap
+# application/json:
+#   the JSON-LD context for version 1, maintained in
+#   <https://github.com/w3c-ccg/zcap-spec/blob/main/context/zcap-v1.jsonld>
+RewriteRule ^v1$ https://w3c-ccg.github.io/zcap-spec/v0/contexts/zcap-v1.jsonld [R=302,L]
--- a/zcap/.htaccess
+++ b/zcap/.htaccess
@@ -23,4 +23,4 @@ RewriteRule ^$ https://w3c-ccg.github.io/zcap-spec/ [R=302,L]
 # application/json:
 #   the JSON-LD context for version 1, maintained in
 #   <https://github.com/w3c-ccg/zcap-spec/blob/main/context/zcap-v1.jsonld>
-RewriteRule ^v1$ https://w3c-ccg.github.io/zcap-spec/v0/contexts/zcap-v1.jsonld [R=302,L]
+RewriteRule ^v1$ https://w3c-ccg.github.io/zcap-spec/latest/contexts/zcap-v1.jsonld [R=302,L]


gobengo:
@dmitrizagidulin @msporny wdyt?

dmitrizagidulin (approved):
👍 I like it.

zcap/.htaccess
```diff
@@ -1,16 +1,26 @@
...
 # homepage:
-#   https://digitalbazaar.github.io/zcap-context/
+#   https://w3c-ccg.github.io/zcap-spec/
 # maintainers:
 #   @msporny
 #   @dlongley
 #   @dmitrizagidulin
 #   @davidlehn
-#
+#   bengo.is a.k.a. https://github.com/gobengo/
```
msporny: Use your github handle at least so folks know you're a maintainer and can easily click to verify you.

msporny (approved):

zcap/.htaccess
davidlehn: Could match the style:
```suggestion
#   @gobengo
```

zcap/.htaccess
```diff
@@ -1,16 +1,26 @@
...
+# /
+# Redirect to zcap spec
+RewriteRule ^$ https://w3c-ccg.github.io/zcap-spec/ [R=302,L]
+
+# /v1
+# Redirect to version 1 of zcap
+# application/json:
+#   the JSON-LD context for version 1, maintained in
+#   <https://github.com/w3c-ccg/zcap-spec/blob/main/context/zcap-v1.jsonld>
+RewriteRule ^v1$ https://w3c-ccg.github.io/zcap-spec/context/zcap-v1.jsonld [R=302,L]
```
davidlehn: A wildcard style has worked will for other similar specs/contexts.  Makes it easier to drop in contexts in the spec without needing updates here.  The following would automatically support v1, v2rc1, v2rc2, v2, v3, etc when using the appropriate pattern for spec files:
```suggestion
RewriteRule ^v(.*)$ https://w3c-ccg.github.io/zcap-spec/context/zcap-v$1.jsonld [R=302,L]
```
And if I may ask, why was that changed from `contexts/` to `context/`?  I suspect the v1 context file has to remain static forever at this point.  So there will be more contexts.

gobengo: > why was that changed from contexts/ to context/?

I changed it here to match what would resolve in the new repo, which is different than the `zcap-context` repo from before and has a different structure.
In zcap-context, the directory was `contexts/`.
In zcap-spec, the directory is `context/`.

zcap/.htaccess
gobengo: My GitHub handle is in the URL there.

>  and can easily click to verify you.

In my IDE, it's easier to click on URLs than these kinds of `@username` with ambiguous host.

Where is there any kind of clickification?

(I'm going to accept the suggestion anyway for consistency)

zcap/.htaccess
gobengo: I don't want URLs other than `v1` to do anything other than 404 (at `w3id.org` for now, no redirect).
My intention here in this PR is only to change the redirect target from the old one to the new one, not to add new features like a general purpose rewrite for future versions.

msporny: What @davidlehn is trying to say politely is that this:

> In zcap-spec, the directory is context/.
> don't want URLs other than v1 to do anything other than 404

... will lead to problems in the future based on our experience managing these sorts of repositories and contexts over the years. 

There will be multiple contexts in that directory, and thus the directory should be named "contexts". 

You do not want to have to keep coming back to this file to change it (the redirect will 404, which is fine, one more step but easier to manage). Having to come back to this file for every context release introduces a brittleness where you have to update this file whenever you update the repo and that sort of back-and-forth has led to failures in production (because someone forgot to update the w3id.org file, or someone updated the file, but had a syntax error that wasn't caught and it broke production for someone out there that wasn't doing the right thing and was caching files). All that to say, what @davidlehn is suggesting is best practice.

gobengo: tl;dr: I hear both of you advocating for `context/` -> `contexts/` in zcap-spec and, more generally, a plural naming convention for (some) parts of the URL structure of `https://w3c-ccg.github.io/zcap-spec/` that this PR redirects to. I don't prefer that, but in the interest of finding consensus, I have done that (which required changes to zcap-spec repo). But please let's just punt the new regex in the RewriteRule out of this PR because all it does is add present risk and future constraints when we add a new context file, and it adds near-zero value until then compared to leaving our options open until we are actually shipping a new context file URL, which is likely to be in the distant future when context will have changed and we are likely to make a different decision on the best RewriteRule.

---

I don't want it to become a pattern that PR review for a small, simple change to a redirect target with two other editors approving becomes a brainstorm for entirely new, untested, functionality like this more complicated (and buggy) RewriteRule regex.

@davidlehn I don't think your suggestion is appropriate to add because it's not germane, it has at least one bug, it adds risk to solve a purported problem we don't have, it provisions many new URLs paths on zcap-spec that add no value today but do add cost/constraint to future versions of how we publish, and it would be sufficient and better to design for multiple versions if/when we actually have them.
* Your code suggestion uses a regex that would rewrite `v1/foo/bar` to `v1/foo/bar.jsonld`, and lead to a 404 at a URL served by the zcap-spec. That would make it look like an error on the zcap-spec, when really it is an error in the regex you're suggesting, and it's an error we can avoid by keeping the discussion of this PR focused on the scope I intended.
* yagni. A better time to design a general purpose URL RewriteRule for more than one version of the context would be when we actually have more than one context, and we aren't unnecessary speculating. Then we will have the actual situation you are prematurely optimizing for. There's no reason to design it until then. Doing so only costs time/money and increases the risk of short term defects due to an unnecessary code change. I did consider doing what you're suggesting, but the reason I left it out of scope is that I really wanted to avoid unnecessarily constraining the solution space for future us when we do actually have that situation (unlikely in the next year). I don't want to speculatively provision new URL patterns in zcap-spec now for `v2` versions that don't actually exist, because that way if/when anyone ever does publish `v2`, I don't want them to have to consider backward compatibility with the (unnecessarily provisioned) `https://w3c-ccg.github.io/zcap-spec/context/zcap-v2.jsonld` (not to mention infinity other zcap-spec URLs, given the broad wildcard) that this would redirect to. Personally I think an eventual URL structure like `https://w3c-ccg.github.io/zcap-spec/v2.0.0/context.jsonld` makes a lot more sense, now that we have semver in the zcap-spec gh-pages URLs. I really don't want to have to come to consensus on *any* future/new version URLs right now, but this suggestion is forcing the issue, and I only bring it up to illustrate that we don't have consensus on the URL structure for v2 et al, and this PR is very useful even without it.

---

@msporny
> this... will lead to problems in the future based on our experience managing these sorts of repositories and contexts over the years.

I appreciate you and @davidlehn trying to help us avoid problems. At the same time, I expect reviewers to clearly elaborate what the problems are, and to do so in a way that doesn't block other work that isn't the root cause of those problems. Merging this PR would not lead to any problems that aren't already in the status quo. If the problem is in the status quo, the best place to describe it is in an issue, not blocking a PR that makes other improvements that do not exacerbate the purported problem.
Also, I don't think you've clearly articulated a problem:

> Having to come back to this file for every context release introduces a brittleness where you have to update this file whenever you update the repo

@msporny I think you're exaggerating. Nothing about my PR would lead to a situation where "you have to update this file whenever you update the [zcap-spec] repo".
* notably, my PR doesn't change this regex. It keeps it how it's been for years, which has not led to that situation
* it would only be required when we bump a major version of the context URL beyond v1. That's extremely unlikely to happen in the next 12 months. Thus, it is extremely unlikely that this regex change will save any more time this year than it has already cost to discuss the suggestion
* I'm confident *when we actually have more than one context file* (not now), we will be able to use the latest up-to-date information and patterns to address your concern, e.g. using either the URL format @davidlehn suggested or something more like the alternative I mentioned above that has valid semver in the URL paths. Until then, doing nothing has the same benefits as @davidlehn's suggestion, without adding redirects to a bunch of new URLs on the zcap-spec that will 404 for the medium-term future.

---

>  I suspect the v1 context file has to remain static forever at this point.

@davidlehn That is different than what I was suspecting. I suspected the v1 context file can change, but only in ways that are backward-compatible (e.g. https://github.com/w3c-ccg/zcap-spec/pull/96). We'll definitely have multiple context files if/when we want to make a backward-incompatible change to the context file that would justify a major version bump, but not necessarily any time sooner. Only commenting here to name we are operating from different premises on this wrt the speculation of how soon this regex change will deliver any value. Because it will deliver no value today, there are other issues that will, I would like to defer work on it until we do have a second context file. Please allow me to decline this suggestion. We can accomplish the goal without it.

Here is where I rename `context/` -> `contexts/` in zcap-spec: https://github.com/w3c-ccg/zcap-spec/pull/99

gobengo: Thanks both — I've made the `context/` → `contexts/` change here, along with the corresponding rename in zcap-spec (w3c-ccg/zcap-spec#99).

I've also updated the zcap-spec deployment workflow. In addition to the exact-version paths it already published:

* https://w3c-ccg.github.io/zcap-spec/v0.3.0/contexts/zcap-v1.jsonld

it now also publishes to semver-range paths for `v{major}`, `v{major}.{minor}`, and `latest`:

* https://w3c-ccg.github.io/zcap-spec/v0/contexts/zcap-v1.jsonld
* https://w3c-ccg.github.io/zcap-spec/v0.3/contexts/zcap-v1.jsonld
* https://w3c-ccg.github.io/zcap-spec/latest/contexts/zcap-v1.jsonld

> Having to come back to this file for every context release introduces a brittleness where you have to update this file whenever you update the repo and that sort of back-and-forth has led to failures in production

@msporny This PR doesn't create that requirement. Once deployed, `https://w3id.org/zcap/v1` redirects to a `latest` URL, and the zcap-spec GitHub Actions workflow republishes that URL on every release. Updating zcap-spec will not require touching this .htaccess. A change here would only be needed if we published a second context file under a new major version — which I don't expect this year.

On the generalized RewriteRule, I'd like to leave it out of this PR:

* Every URL it newly matches resolves to a 404 — there is no v2 context to serve. The only difference is who serves the error and how many round trips it takes. Today `https://w3id.org/zcap/v2` gets a plain "Not Found" from w3id in one request, with the typed URL still in the address bar, which reads correctly as "that doesn't exist." After the change, the client makes a second request and gets a GitHub 404 advising them to "make sure that the filename case matches the URL as well as any file permissions" — implying a client or server fault that isn't there.
* As written, the regex matches any path under `v1/`, so `v1/foo/bar` → `v1/foo/bar.jsonld`. Narrowing it would help, but wouldn't change the point above.
* It provisions redirect targets for files that don't exist and may never exist. If someone publishes a v2 context later, I'd rather they not have to weigh backward compatibility with `.../contexts/zcap-v2.jsonld`. I'd personally prefer something closer to `https://w3c-ccg.github.io/zcap-spec/v2/context.jsonld` now that the gh-pages URLs carry semver — but that's a question we can settle when there's an actual v2 context to settle it for.
* @davidlehn on "I suspect the v1 context file has to remain static forever at this point. So there will be more contexts." — I don't think that follows. The v1 context can still change in backward-compatible ways; w3c-ccg/zcap-spec#96 is one such change. A second context file is only forced by a backward-incompatible change, which we don't have on the horizon.
  * And if it turns out we can't change the referent of `https://w3id.org/zcap/v1` at all, I'd rather register a new w3id alias than freeze that one. That way the context URL version can track the major version of zcap-spec, e.g. `https://w3id.org/authorization-capabilities/v0/context.jsonld` for the current v0.3.0 release. But I'd prefer using `zcap` for as long as we can avoid backward-incompatible changes to https://w3id.org/zcap/v1

Based on yalls feedback, I spent a few hours in both repos, improving how we deploy things to give us more redirect targets. But I still don't think the RewriteRule change is worth it, primarily because of how it adds no value, but does pave a path for users to end up on GitHub's more confusing 404 page that reads like an error instead of w3id.org's 404 page that is clear that we just haven't published anything beyond `v1` to the host that serves our context URL (`w3id.org`). We may not release a v2 context URL for quite some time, and we can revisit this then with more real-world context informing the design. Until then we won't be redirecting users to GitHub error pages on URLs that have never resolved, and possible never will.

I'm happy to design a general rule when there's a real second context file to design against.
Can we merge this as-is and open an separate issue for the multi-version RewriteRule scheme?

davidlehn: With my w3id hat on, to further elaborate on wildcards:
- The wildcard idea is to avoid churn here, avoid delays while waiting for w3id maintainers to do merges, and make it easy to use new paths by dropping a file in at the target.
- Which server 404s for an unknown path seems unimportant.  How often does that come up?  Garbage in, garbage out.  I've always suggested more advanced path validation should happen on the target rather than this service.  Keep it simple here.
- Using wildcards does (usually) imply keeping a known path structure on the target for every version, which might not always work.  But in practice it seems lots of people have success with this method.

With my digitalbazaar/json-ld/cbor-ld hats on:
- There are concerns about this particular context changing.  I believe it needs to remain static.
- There are now comments on the related spec repo PRs.
- Backwards and forwards compatibility can be tricky with JSON-LD contexts.  Things like using `@protected` and making changes that might conflict with how someone else was using that context the considered static.  Also possible CBOR-LD related issues due to changes.  That's why other projects make static versions.
- We've been trying to promote using cached versions of known published static contexts, such as the one here.  It avoids network issues, avoids contexts changing out from under you, is far more performant, etc. 
- We are using this context via the NPM package created in the repo this redirect went to.  Changes would cause that to no longer match the live context.  Others may have same expectations.
- Guidelines, processes, specs, and tooling for these sorts of issues are a work in progress.
- Naming is always hard.  Strictly making `https://w3id.org/zcap/vN` match spec version 1.0.0 has pros/cons.  It "matches" at first glance I suppose.  But it doesn't really matter either.  I always thought of it as the version of a *context*, independent from a major spec version.  Also if spec version N+1 comes out, but the context didn't change, then there would be no need to mint a new context id.  And integers seem easier to deal with than v1.2.3.  Lots more that could be discussed either way.

Considering all that, what's the path forward here?  I've got multiple hats on with this PR so happy to continue over in spec repo issues too.

msporny: I'm fine to merge the PR as-is, but know that this is the wrong approach and has anti-patterns embedded in it that will turn out badly. For example:

> The v1 context can still change in backward-compatible ways

Incorrect, that is a known anti-pattern, and I've tried to elaborate more on why here:

https://github.com/w3c-ccg/zcap-spec/issues/95#issuecomment-5510477325

> A change here would only be needed if we published a second context file under a new major version — which I don't expect this year.

Also incorrect, you just triggered the need for a new context here:

https://github.com/w3c-ccg/zcap-spec/pull/96

> publishes to semver-range paths for v{major}, v{major}.{minor}, and latest

This is a bad idea, it will invite people to write software for semvers, which you definitely do not want to promote. It leads to fragmented ecosystems in the long-run.

Things that you do not have consensus on as an Editor of the associated specification:

* That semver management of a specification is the right thing to do for the main spec or related context management (specifications are not software, semver makes less sense and is sometimes an anti-pattern wrt. standards).
* That context files can change once published (known anti-pattern and goes against establish standards publication best practices).

What you are proposing is going to lead to fragmentation in the ecosystem and create churn in w3id and raise the maintenance load for the w3id.org maintainers. 

I am not supportive of the direction, but am not blocking this PR (I guarantee you that we will be reviewing another PR for this redirect in a few weeks time).

BigBlueHat: @msporny if the current `https://w3id.org/zcap/v1` needs to stay immutable, this would break that, correct? Since it maps to the `latest/` folder?

Obviously, making the redirected-to (or rewritten-to) result immutable is outside the scope of this repo...but maybe a hash of the contents or something should surface here as a comment?

Changing these "permanent vanity" URLs in this case can--as you've described--be cataclysmic... So looking for more ways to denote when/where that's actually the reality (since not all uses of w3id.org face this requirement).

I'll make a related issue on w3c-ccg/zcap-spec also.

BigBlueHat: Explained more in https://github.com/w3c-ccg/zcap-spec/issues/101

msporny: > @msporny if the current https://w3id.org/zcap/v1 needs to stay immutable, this would break that, correct? Since it maps to the latest/ folder?

Yes, good catch. I was presuming that @gobengo wasn't planning on modifying that context, but it seems that he also intends to do that... and yes, that would be really, really, bad.

gobengo: No, I do not intend to change that.

davidlehn:
@gobengo Please fix the merge conflict.  Since this was created, the ids were all moved to `ids/`.  As discussed on the related call, we'll merge this once that's fixed.

dgarijo:
Ping @gobengo

PR #6628, opened 2026-09-02
shapetrees: serve the rendered specification and primer, not just the vocabulary

ericprud:
The `shapetrees` entry currently sends every path to the raw `shapetrees.ttl`. That is right for the namespace URI itself, but it means the two documents the namespace is supposed to name are unreachable: `https://w3id.org/shapetrees/TR/specification/` and `https://w3id.org/shapetrees/TR/primer/` both hand back the Turtle file.

Those documents now exist again. They were published at `shapetrees.org` until that domain expired, and have been rebuilt on GitHub Pages, one Pages site per source repository:

| Request | Redirects to |
| --- | --- |
| `/shapetrees` with an RDF `Accept` header | `https://shapetrees.github.io/specification/shapetrees.ttl` |
| `/shapetrees` from a browser | `https://shapetrees.github.io/` |
| `/shapetrees/TR/specification/` | `https://shapetrees.github.io/specification/` |
| `/shapetrees/TR/primer/` | `https://shapetrees.github.io/primer/` |
| anything else under `/shapetrees/` | the same path on `https://shapetrees.github.io/` |

Two changes beyond the new `/TR/` paths are worth flagging:

- The vocabulary moves from `raw.githubusercontent.com` to GitHub Pages, so it is served as `text/turtle` rather than `text/plain`.
- It is served only to clients that ask for RDF. A browser following `https://w3id.org/shapetrees` now lands on the Shape Trees home page instead of downloading a Turtle file. RDF clients are unaffected.

Redirects are `302` so the Pages hosts can be moved later without poisoning caches.

I am one of the listed maintainers of this entry (@ericprud). The rules were tested under Apache 2.4 with `mod_rewrite`, checking each row above plus a browser `Accept` header, which correctly falls through to the home page.

### DIFF
--- a/shapetrees/.htaccess
+++ b/shapetrees/.htaccess
@@ -1,8 +1,41 @@
 # shapetrees
 #
-# https://w3id.org/shapetrees redirects to
-# https://raw.githubusercontent.com/shapetrees/specification/refs/heads/main/shapetrees.ttl
+# Shape Trees, a W3C Solid Community Group work effort.
+# Maintained at https://github.com/shapetrees/specification
+#
+# The documents were formerly published at shapetrees.org; that domain expired
+# and is now squatted. They live on GitHub Pages instead, one Pages site per
+# source repository, so this serves the rendered documents as well as the
+# vocabulary.
 
+Options +FollowSymLinks
 RewriteEngine on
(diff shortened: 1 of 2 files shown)


davidlehn:
Please rebase on master.  The files moved to be under `ids/`.

dgarijo:
This PR has conflicts, needs to be rebased and moved. It has shown no activity in 3 weeks, so I am closing it. Please reopen if appropriate

PR #6676, opened 2026-09-10
Add gto: persistent identifiers for GoTriple resources

AlessandroBertozzi:
## Brief Description

New top-level identifier `gto` for the resources published by
[GoTriple](https://www.gotriple.eu), the discovery platform for the social
sciences and humanities built by the TRIPLE project.

    https://w3id.org/gto/{type}/{reference}

`{type}` is the entity type spelled out (`document`, `dataset`, `media-object`,
`semantic-artefact`, `project`, `profile`); `{reference}` is the ARK name of the
resource, without its NAAN, so the w3id IRI and the ARK of the same resource
share their last segment by construction.

Two redirects only:

| request | code | target |
|---|---|---|
| `/gto/` | 302 | `https://kg.gotriple.eu/` |
| `/gto/{type}/{reference}` | 302 | `https://kg.gotriple.eu/ark:64989/{reference}` |

Resources resolve to their complete ARK (NAAN 64989) on the GoTriple knowledge
graph. Content negotiation and the 404 of an unknown name are that server's
responsibility, not this layer's. The `{type}` segment is not carried over: ARK
names are unique inside the NAAN. Nothing about the entity types is encoded in
the `.htaccess`, so this space will not need another pull request here when a
type is added. The target host and the NAAN are deployment facts rather than
naming decisions; the `w3id.org/gto/…` IRIs themselves never move, which is why
they are minted here.

Out of scope on purpose: the TRIPLE ontology is a different namespace
(`https://gotriple.eu/ontology/triple/`), documented at
https://github.com/atrium-research/triple-ontology, and is not served from this
space.

Maintainer: Alessandro Bertozzi, GitHub username `AlessandroBertozzi`.

## General Checklist

## New ID Directory Checklist

### DIFF
--- a/ids/gto/.htaccess
+++ b/ids/gto/.htaccess
@@ -1,0 +1,44 @@
+# # /gto/
+#
+# GoTriple — persistent identifiers for the resources published by the GoTriple
+# discovery platform (https://www.gotriple.eu), the discovery service for the
+# social sciences and humanities built by the TRIPLE project.
+#
+#     https://w3id.org/gto/{type}/{reference}
+#
+# {type} is the entity type spelled out (document, dataset, media-object,
+# semantic-artefact, project, profile). {reference} is the ARK name of the
+# resource without its NAAN, so the w3id IRI and the ARK of the same resource
+# share their last segment by construction.
+#
+# Scope: resources only. The TRIPLE ontology is a separate namespace on purpose
+# (https://gotriple.eu/ontology/triple) and is not served from here.
+# Naming rules:
+# https://github.com/atrium-research/triple-ontology/blob/main/URI-CONVENTIONS.md
+#
+# ## Contact
+# This space is administered by:
+#
+# Alessandro Bertozzi
+# GitHub username: AlessandroBertozzi
+
+# "always": every response here is a redirect, and a CORS check applies to each
+# hop of the chain — the default (2xx only) would never fire.
+Header always set Access-Control-Allow-Origin "*"
+Options +FollowSymLinks
+RewriteEngine on
+
+# The prefix root is not a resource: send a human to the knowledge graph itself.
+RewriteRule ^$ https://kg.gotriple.eu/ [R=302,L]
+
+# Every resource is forwarded to its complete ARK on the GoTriple knowledge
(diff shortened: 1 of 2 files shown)


davidlehn:
I loaded up that kb site and see it's listing millions of various types of things at the top.  Is this going to be a frontend that will lead to all those?  Asking because many LLM scrapers are already hammering this service.  Not quite sure how handle that and what advice to give for these wildcard type of ids here.  Other than that, this looks fine.

AlessandroBertozzi:
Hi @davidlehn, thanks for looking into this, and sorry for the slow reply.

A bit of context first. GoTriple is a discovery platform for social sciences and
humanities research, built by the TRIPLE project. It aggregates records from many
providers (publications, datasets, researcher profiles, projects, around 30 million
in total) and what you opened is the knowledge graph we are building on top of that
corpus. The `/gto/` space is there to give those records stable identifiers that don't
depend on where the graph is hosted.

So no, it won't be a frontend. `/gto/` only resolves individual resources: you ask for
one identifier and you get that record back, as RDF if you're a machine or as the
human-readable page if you're a browser. There is no index or listing behind it, and
nothing in this space links into the corpus, so it can't be used to enumerate those
millions of records. People browse GoTriple on its own URL and never go through w3id.
The traffic you'd see here comes from clients that already hold an identifier (found
in an RDF dump, a SPARQL result, someone else's graph) and want to dereference it,
which is a very different order of magnitude from the corpus itself.

As for where we are: the resolution routes on `kg.gotriple.eu` are still being
developed, which is why an identifier currently lands on the app shell if you try
one. We wanted to register the prefix now because the identifiers will start
appearing in the data well before the resolver is final; if the target needs
adjusting later, that's a small follow-up PR. The setup is essentially the same as
`/oc/`, which already serves a bigger corpus this way
(`w3id.org/oc/meta/br/0601` → `opencitations.net/meta/br/0601`).

Happy to answer anything else.

davidlehn:
Interesting you mention `/oc/` since that's one of the LLM crawler targets.  It apparently a frontend URL to a large amount of resources that Anthropic, OpenAI, etc want to slurp up.

PR #6684, opened 2026-09-11
cdif: add unversioned aliases and document the versioning policy

smrgeoinfo:
All changes are confined to `ids/cdif/`. No other namespace is touched, and no existing rule changes behaviour.

### 1. Unversioned aliases (`.htaccess`, +64)

Adds 20 rules so an unversioned CDIF URI redirects to the current version:

```
https://w3id.org/cdif/core/          ->  https://w3id.org/cdif/core/1.1/
https://w3id.org/cdif/core/schema    ->  https://w3id.org/cdif/core/1.1/schema
```

Covered: `core`, `discovery`, `data_description`, `data_structure`, `manifest`, `provenance`, `conceptscheme`, `codelist` (all at 1.1), plus `xasCore` and `xasOptional` (1.0).

These are **302**, deliberately, while the existing versioned conformance URIs stay **303**: an alias target is expected to move at each minor release, a versioned URI's target must not. The patterns do not match a version segment, so they cannot shadow the versioned rules regardless of ordering.

Every alias target has existing versioned rules in the same file, and all ten versioned URIs resolve today.

### 2. Versioning policy (`README.md`, +61)

Documents something the rules cannot express on their own: a version's rules point at GitHub Pages, which serves each release repo's `main` branch, and `main` is always the newest release. That is correct only while the version named in the URI *is* the newest. When a new minor version ships, the outgoing version's rules must be repointed to its release tag, or `core/1.1/schema` quietly begins serving 1.2 content — with no error anywhere, since a 1.2 schema is a perfectly valid schema.

Thus the tag rules use `raw.githubusercontent.com`, which serves `text/plain`. That was measured against CDIF's consumers (the JSON Schema and SHACL loaders name the format rather than sniffing, and CDIF records carry an inline `@context`). Although it looks like an inconsistency next to the Pages rules it should not be "fixed" back to Pages — doing so would silently start serving the wrong version.

### 3. Maintenance guide (`CLAUDE.md`, new, 80 lines)

The step-by-step checklist for cutting a new version, verification commands, and the traps — including the ordering constraint between sub-path and bare conformance-URI rules, and `profile-provenance`, which is unreleased and so does not follow the pattern.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ids/cdif/.htaccess
+++ b/ids/cdif/.htaccess
@@ -46,8 +46,18 @@ SetEnvIf Request_URI ^.*$ ADA_BASE=https://usgin.github.io/geochemBuildingBlocks
 SetEnvIf Request_URI ^.*$ DDE_BASE=https://usgin.github.io/ddeBuildingBlocks
 SetEnvIf Request_URI ^.*$ ECRR_BASE=https://usgin.github.io/ecrrBuildingBlocks
 
-# Release-repo bases (each release repo is its own GitHub Pages site,
-# serving from the reviewRevision202606 branch as of 2026-05-30)
+# Release-repo bases (each release repo is its own GitHub Pages site).
+#
+# Pages serves each repo's `main` branch, and main is always the CURRENT RELEASE
+# -- work happens on `updates` and reaches main by pull request at release time.
+# So these bases always resolve to the newest release, with GitHub Pages'
+# correct content types (application/json for schemas).
+#
+# That is why the versioned rules below can point at these bases only while the
+# version they name is current. When the next minor version ships, the outgoing
+# version's rules must be repointed at its release tag on raw.githubusercontent
+# -- see README.md, "Cutting a new version". Miss that step and, say,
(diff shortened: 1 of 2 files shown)


davidlehn:
- Remove CLAUDE.md from the PR and history.  Only .htaccess and readmes are supported here.
- The readme was already too verbose of content to be hosted here.  This is making it longer with verbose LLM output.

I'd suggest hosting that CLAUDE.md and extended README.md contents on your site, similar to testing tools you refer to.  Then add some references in the README.md to them.  Agents are pretty good at following links like that.

smrgeoinfo:
Done — CLAUDE.md is removed from the PR and from the branch history (amended, not reverted), and the README addition is cut from 61 lines to a short note plus a link. The policy, release checklist and the weekly check now live at CDIF/validation docs/w3id-redirect-policy.md, next to the tool that runs them. The PR is two files now: the alias rules and that note.

smrgeoinfo:
thanks!

PR #6704, opened 2026-09-16
campop: enable place and context rules

docuracy:
The namespace said "Entity-level rules will be added in a follow-up pull request when the corresponding static API is deployed". It is deployed, so they are added.

  /campop/place/{HCS}.{Place}   content-negotiated, 303
  /campop/context               the JSON-LD context, 302

Both resolve to https://docuracy.github.io/CAMPOP-Places/.

GitHub Pages is static and cannot read an Accept header, so the negotiation has to happen here and Pages only ever serves a file. application/ld+json, application/geo+json and application/json get the data; everything else, including text/html and */*, gets the page. An explicit .json suffix beats any Accept header, for curl and for clients that send only */*.

303 See Other rather than 302 for places: a place is not a document, so its identifier must not 200. The context is a 302 because it genuinely IS a document and there is no non-information resource to redirect away from.

Place identifiers are {HCS}.{Place} -- a three-letter Historic Counties Standard county code and the modern place name, e.g. AGL.Beaumaris. The county code leads so a column of identifiers sorts into county order, and the whole string is authoritative.

Source and criterion are dropped from the namespace layout: they do not belong in an agnostic dataset of places.

locolligo/CAMPOP-Places is untouched and keeps resolving, so existing links to the 2022 deposit continue to work.

## General Checklist

## New ID Directory Checklist

- [ ] GitHub username ids are listed in the changed maintainer details.
- [ ] The GitHub account submitting this PR is listed as a maintainer of the directories and files changed in this PR or one or more of those maintainers are tagged to approve these changes.


docuracy:
Withdrawn to correct the description and README wording (and the checklist section, which is an update to an existing directory, not a new one). The redirect rules are unchanged. A replacement PR follows from docuracy:campop-place-rules-v2.

